PR #21, opened 2021-05-18
[sstar] make local build on ubuntu 20 happy

ch999dev:


ZigFisher:
Good afternoon!
We are very grateful for the work you have done.
In the near future, we will test your patch and enable it to work
Thank you again very much !

ZigFisher:
Dear ch999dev
Please pay attention to the feedback on your patch, you may need your help
https://github.com/OpenIPC/openipc-2.1/issues/23

ch999dev:
Hello. There is nothing console related in this patch. It only fixes things related to Python 3. May be the previous version of the script didn't generate / apply the DTB at all?

operutka:
@ch999dev I suspect that the problem might be in `scripts/ms_bin_option_get_int.py` where the lines parsing value were originally commented out but your patch replaces them with an un-commented version:
```
if __name__ == '__main__':

     name=sys.argv[2]
-#    if sys.argv[3].upper().startswith( '0X' ):
-#        value=long(sys.argv[3],16)
-#    else:
-#        value=long(sys.argv[3])
+    if sys.version_info.major == 2:
+        if sys.argv[3].upper().startswith( '0X' ):
+            value=long(sys.argv[3],16)
+        else:
+            value=long(sys.argv[3])
+    else:
+        if sys.argv[3].upper().startswith( '0X' ):
+            value=int(sys.argv[3],16)
+        else:
+            value=int(sys.argv[3])
```

I'm currently building a new kernel image to check this hypothesis.

operutka:
There is also forgotten print in `scripts/ms_bin_option_update_int.py`.

ch999dev:
1 Yes. Good catch! This was commented out in the original script
2 This is not a forgotten print :) Just left this to check on build.

operutka:
Now I've checked the original `scripts/ms_bin_option_get_int.py` and the `value` isn't being used anywhere, so the commented/uncommented section should be irrelevant.

But still, if I build my kernel image with `02_infinity6b0_kernel_4.9.84_sdk-Ispahan-ILS00V020_python3_check.patch`, the console is dead. If I build the kernel image without this patch, the console works fine. Strange...

operutka:
@ch999dev Finally! I've got the devil. The problem is in the third file named `ms_builtin_dtb_update.py`. You've added a few lines that replace a built-in constant with whatever is passed as the third command line argument:
```python
name='#MS_DTB#'

...

if sys.version_info.major == 2:
    name=sys.argv[2]
else:
    name=sys.argv[2].encode()
```

Removing these four lines fixes the issue.

ch999dev:
As far as I remember, without these patches BR2 could not determine KERNEL_VERSION and stopped at the stage of finalizing the image. And the mstar DTB script just failed (but BR2 ignored the error).
(ubuntu 20 / python 3)
What versions are you building on?

operutka:
I'm building it using Docker. The base system is Debian Buster. I've also got into the problem where BR2 could not finalize the image but it was fixed by this patch: `03_infinity6b0_kernel_4.9.84_sdk-Ispahan-ILS00V020_fix_makefile.patch`.

ch999dev:
This is where the build stopped without a patch on local Ubuntu 20 / Python3
https://github.com/OpenIPC/openipc-2.1/runs/2719774667?check_suite_focus=true#step:7:127

operutka:
Yes, that's the same line where it stopped in my case but as I mentioned, `03_infinity6b0_kernel_4.9.84_sdk-Ispahan-ILS00V020_fix_makefile.patch` fixes it. However, the changes in `ms_builtin_dtb_update.py` seem to be unrelated and, at least in my case, it breaks things.

ch999dev:
What if the unpatched `ms_builtin_dtb_update.py` silently fails and the DTB is remains not patched? And because of this, you have a console? :)

operutka:
> What if the unpatched ms_builtin_dtb_update.py silently fails and the DTB is remains not patched? And because of this, you have a console? :)

I don't think so. I've added a few debug outputs into the `ms_builtin_dtb_update.py` to find out what you're actually putting into the `name` variable. The command line argument that you use equals to the path to the DTB (`arch/arm/boot/dts/infinity6b0-ssc009a-s01a.dtb` in my case). This string, obviously, cannot be found in the resulting kernel image. The script is supposed to find the "#MS_DTB#" placeholder in the kernel image and replace it with the actual size of the DTB. Your patch breaks this.

Finally, it really isn't about having or not-having console. It is about having a broken kernel image or a working one. :)

To sum it up:
1. `03_infinity6b0_kernel_4.9.84_sdk-Ispahan-ILS00V020_fix_makefile.patch` is fine. It fixes the image finalization.
2. `02_infinity6b0_kernel_4.9.84_sdk-Ispahan-ILS00V020_python3_check.patch`:
  a. changes in `ms_bin_option_get_int.py` are fine (even though not necessary because the original lines were commented out and the value isn't being used anywhere),
  b. changes in `ms_bin_option_update_int.py` are also good (though the extra print probably isn't necessary),
  c. changes in `ms_builtin_dtb_update.py` are not good because they replace a built-in constant with a value from command line and it results in a broken DTB.

ch999dev:
Ah. Ok. WRT 2/c 
This should be something like (seems typo)

```python
    if sys.version_info.major == 2:
        offset=fmap.find(name)
    else:
        offset=fmap.find(name.encode)
```

operutka:
Yes. That seems more like it. Though you probably meant `offset=fmap.find(name.encode())`.

ch999dev:
Yes, exactly

ch999dev:
pushed #25

PR #25, opened 2021-06-01
Typo in patch.

ch999dev:
Thanks to @operutka


ZigFisher:
Thank you very much for your reply !

PR #26, opened 2021-06-04
Add uClibc-ng config for SSC335

operutka:
I've found using a MUSL based system quite difficult on SSC335 given the fact that all SigmaStar libs and tools were built using a uClibc based toolchain. Even though I'd love to use MUSL instead of uClibc, it's been a real pain trying to make everything work. Switching to uClibc-ng can solve a lot of headaches.

I believe this might be also useful to other people who use OpenIPC for their projects.

I had to create a copy of the rootfs overlay because there are some symbolic links breaking the system in the default overlay.


ZigFisher:
I reviewed your edits and added a configuration file for uclibc. Thank you very much for the work you have done. However, I can't accept the patch in this form as we need to discuss what to do with the symbolic link between UCLIBC and MUSL in order to preserve the unity of the configurations.

ZigFisher:
As far as I understand, it is these two links that prevent us from reaching a consensus ?

overlay/lib/ld-uClibc.so.0 => /lib/libc.so
overlay/usr/bin/ldd => /lib/libc.so

operutka:
Let me answer your questions (including those sent via PM).

> Did you specify their changes for some reason after working with the documentation? Is your version more correct?
> 
> ...
>
> BR2_ARM_EABI=y
> BR2_ARM_INSTRUCTIONS_THUMB2=y
>
> I think these values are more correct, since I see them in the kernel configuration file from the manufacturer. What do you think?

It's hard to say what is correct. :) Regarding the hard float vs. soft float ABI - the user space libraries were built using a hard float toolchain and linking with a soft float code may cause problems. I faced this issue when I tried to link statically the libcam_os_wrapper.a lib. That's why I switched to hard float ABI. Note that SigmaStar also distribute a hard float toolchain in their SDK. I have it under:
```
01. MSTAR 335 SDK-20201123T084514Z-001/01. MSTAR 335 SDK/开发文档_V020/tool/arm-buildroot-linux-uclibcgnueabihf-4.9.4-uclibc-1.0.31.tar.xz
```

Regarding the instruction set - it shouldn't matter much. It should be possible to combine code compiled into Thumb2 and ARM together. I prefer the ARM instruction set a bit more because of a simple reason. Even though the Thumb2 instructions are shorter and, theoretically, it should result in smaller binaries, it often isn't the case. The instruction set is simpler and it often means that more instructions need to be generated to cover certain functionality. From my experience, it really differs between applications. You can end up having a smaller binary and you can also end up having a bigger binary.

If we ignore the binary size differences, there is also another point of view - performance. Using Thumb2 can negatively affect performance because the compiler won't be able to use more complex instructions that could make the execution faster.

Of course, all this is highly theoretical and we could have endless debates whether to use Thumb2 or ARM instruction set and which one is better. Personally, I'm fine with both. I prefer ARM a bit more but if you decide to use Thumb2 instead, I'm fine with it.

> Then the locale is enabled. I take it this need is solely related to uclibc or is it done for another additional purpose?

I disabled the locale because when you enable it, uClibc won't export the `__ctype_b` symbol. The user space libs from SigmaStar depend on this symbol, so it's very likely that they were built using a toolchain where locale was disabled.

> It remains to specify only the question of overlay-uclibc, whether a separate folder is needed. I would really hate to have a lot of the same things in the project.

I agree. I also hate duplicating things. Take it as a kick off. There are these options in the buildroot config (under System configuration):
* Custom scripts to run before creating filesystem images (BR2_ROOTFS_POST_BUILD_SCRIPT)
* Custom scripts to run inside the fakeroot environment (BR2_ROOTFS_POST_FAKEROOT_SCRIPT)

Maybe we could use one of these but I've never used them, so I'm not sure if it'll be helpful.

> I also use the SDK "Ispahan-ILS00V020" but the binaries and libraries are of a different size. This makes sense to understand... My version is dated November 2020.

My version is also from November 2020 (the exact version is 20201123T084514Z-001). I just stripped the kernel modules from debug symbols. The original modules were not stripped and they were huge (e.g. `mhal.ko` was 14MB). I'm not sure if I tried to strip also the user space libraries. It's possible though.

I've tested all my changes, built a streaming app and I've verified that everything works. I believe that this environment is really close to the original one from SigmaStar and there is no need to hack anything. You don't need to resolve missing symbols when linking your apps. Everything should be there.

operutka:
> As far as I understand, it is these two links that prevent us from reaching a consensus?

Yes. This is the only difference between the two overlays.

ZigFisher:
Super ! 
Thank you so much for taking the time to clearly explain your views. 
I will talk to my colleagues and give an answer to our questions here.

ZigFisher:
Copy that, thank you.
I'll report it done.

ZigFisher:
> I'm gonna close this PR now and open a new one once you finish the scripting for uClibc overlay.

Please open issue in this repo, and check my last commit
https://github.com/OpenIPC/openipc-2.1/commit/b4aad6f8cd1f2b5da66845e5d1c7f464139b301a

operutka:
Thanks. I have tested the changes and everything seems to work fine. You even renamed the uclibc defconfig to goodcam defconfig, so there is no need for me to open another PR.

PR #27, opened 2021-06-05
Small hisilicon defconfig fix to BR2_EXTERNAL_ variable

Dimonix:


ZigFisher:
Thank you so much for the corrections.
I will include your patch in the system soon.

PR #33, opened 2021-07-09
Disable QUAD for MXIC MX25L6433(6)F

dimerr:


dimerr:
tested, works.

ZigFisher:
You're gorgeous as always !
Thanks !

PR #35, opened 2021-08-07
Fix Hi3536 series support

dimerr:


ZigFisher:
Many thanks, Sir !  ;)

PR #53, opened 2021-08-30
fix(sysupgrade): use curl instead ping - icmp not always available

carnei-ro:
ICMP to github.com is not reliable. If no connection available, curl will send "000" status.


ZigFisher:
Thank you, Sir ! ;)

PR #114, opened 2021-12-19
add rtl8188eus driver; add hostapd with Realtek driver

nitr0man:
add driver rtl8188eus (which supports rtl8188eu); it has nl80211 API (kernel driver r8188eu hasn't). also it works with wifi adapters with disabled AP mode in EEPROM.
add hostapd patched for Realtek drivers which doesn't support nl80211 API


ZigFisher:
thank you !

PR #130, opened 2021-12-30
Added support for bridge and vlan initialization to /etc/network/interfaces

nitr0man:
needs busybox config review (make olddefconfig in buildroot with fresh busybox enables unneeded options like tc)


ZigFisher (approved):
Thank you !

PR #197, opened 2022-03-02
Update unknown_unknown_fh8856v100_openipc_defconfig

caliph9k:
Fixed BR2_LINUX_KERNEL_CUSTOM_CONFIG_FILE and BR2_LINUX_KERNEL_EXT_FULLHAN_PATCHER_LIST which were set to a directory that doesn't exist.


caliph9k:
With this patch in place my local repository compiled successfully and when I loaded the zImage on a FH8856 camera it booted (at least as far as mounting root, which failed as I haven't written anything to the flash yet).

PR #224, opened 2022-05-14
fix building scripts

cronyx:
Fix building scripts for the new kernel and rootfs sizes
Add option to select majestic sources to the build lite or ultimate (16ev300 only)


ZigFisher:
Thank you !

PR #246, opened 2022-06-24
Added OmniVision OV2710 .so and .ini files

mpissarello:


ZigFisher:
Many thanks, colleague ! ;)

PR #274, opened 2022-07-22
dont run 2 copies sysupgrade script

cronyx:


ZigFisher:
Many thanks !

khseal:
I ran the update twice through the web panel...

themactep:
Web interface runs the same script, so that fix will work anyway.

khseal:
Ok thanks !

PR #281, opened 2022-07-27
sysupgrade: dirty fix, do not check new version if -n option is used

cronyx:
TODO: Find a better solution


ZigFisher:
Thank you !

PR #404, opened 2022-08-24
Do not upgrade with SD card mounted.

themactep:
And more small fixes.


themactep:
Screenshot from 2022-08-24 19-48-18

PR #414, opened 2022-08-26
Fix ipctool building issues

widgetii:


themactep (approved):
From the top of my head, everything looks legit. IPCTOOL_VERSION takes the latest available commit hash from the master branch. How BR will update it that's the question though. I have pushed microbe-web from dev to master today, so if you have a fresh build, you can check on that particular package if the strategy works.

PR #416, opened 2022-08-27
Sync of work

ZigFisher:


themactep (commented):
I'd rather go with `diff -rN "rom${1}" "$1"` to be on a safe side, to avoid globbing and weird file names.

PR #417, opened 2022-08-27
Revert md5sum rearrange to fix broken sysupgrade

widgetii:


themactep (approved):
Right, you cannot calculate the hash first and then rename the file. Because hash goes along the filename in .md5sum. Once you renamed the file md5sum won't be able to find it.  Rename it first and then calculate the hash.

PR #436, opened 2022-09-05
Add ingenic-opensdk package

ZigFisher:


widgetii (commented):
I believe gpio.ko is not needed here because it compiled from source code anyway

ZigFisher:
We will clean up both binary files as soon as we make the package

PR #445, opened 2022-09-08
Disable thumb2 for development for CV500 platform

widgetii:


themactep (commented):
maybe replace [ ! -z something ] with [ -n something ] which is an equivalent?

PR #450, opened 2022-09-10
Remove F2FS support

ZigFisher:


widgetii:
Why are you doing this? I'm using f2fs on my cameras a lot for SD card storage

PR #465, opened 2022-09-16
add system tweak script to fpv builds

cronyx:


general/package/wifibroadcast/files/tweaksys
```diff
@@ -0,0 +1,73 @@
...
+family=$(ipcinfo -f)
+vendor=$(ipcinfo -v)
+sensor=$(ipcinfo -s)
+
+cli="yaml-cli -i /etc/majestic.yaml"
+
+all_other() {
+  # remove unneded modules
+  sed -i "s!insmod ${family}_rgn.ko!#insmod ${family}_rgn.ko!g" /usr/bin/load_${vendor}
+  sed -i "s!insmod ${family}_jpege.ko!#insmod ${family}_jpege.ko!g" /usr/bin/load_${vendor}
```
widgetii, line 13: ${family} will be `open` for gk and real family name for hisi, because first one is using OpenHisilicon modules and second one is using original binary modules

cronyx: thanks, it will be fixed in the next commit

PR #467, opened 2022-09-17
fix modules prefix

cronyx:


widgetii (approved):
For now, it should work on both platforms, I hope we can more further after openhilison develops in the future

PR #480, opened 2022-09-22
3516cv100+imx122 requires GPIO setup

carbofos:


FlyRouter (changes requested):
We need to finalize the final edits, maybe close this one and make a new PR from scratch.

PR #526, opened 2022-10-27
Added kernel module argument to select higmac phy addr: of_mdio.higma…

carbofos:
…cphy


FlyRouter:
Good research, thank you !

PR #533, opened 2022-11-01
Change release filename to names like openipc.hi3516cv300-ultimate-no…

widgetii:
…r.tgz


FlyRouter (approved):
Good job, thank you !

PR #547, opened 2022-11-07
detect flash type by ipcinfo

cronyx:
Attention to others: This changes does not contain backward compatibility! Do not merge it!


themactep (commented):
Don't we have exit codes in ipcinfo? Checking $? for zero would be more kosher. Or at least checking results against a list of valid expected values. If it is outside the list then employ the old method.

cronyx:
> Don't we have exit codes in ipcinfo? Checking $? for zero would be more kosher. Or at least checking results against a list of valid expected values. If it is outside the list then employ the old method.

ipcinfo don't have exit codes.

themactep (approved):

PR #549, opened 2022-11-12
Create mavlink.conf

anzhdan2:
Конфигурация Mavlink


cronyx:
This config comes in the mavlink-router package and should not be installed from the wifibroadcast package, port changes from this PR will be done in the next update.

PR #598, opened 2022-12-22
fix build rtl8188fu driver

cronyx:


FlyRouter (approved):
Good job !
Thank you !

PR #621, opened 2023-01-24
fix segfault in audio on v1 chips (SDK uses old version of "stat")

pianist:


widgetii (changes requested):
Please remove general/package/hisilicon-osdrv-hi3516cv100/libhisicompat/sTaT.o from this PR

widgetii (approved):

PR #652, opened 2023-02-28
LTE-FPV version with usb-modeswitch

whoim2:
remove wfb-ng and more packages
add usb_modeswitch
changed /etc/network/interfaces
changed /packages/majestic/majestic.mk (fix for compile as fpv)
tested at huawei e3372h


FlyRouter (changes requested):
Requires fine-tuning of sent PRs
- you need an argument for the chosen configuration file name, why the lte prefix is set before the firmware type
- remove the LTE interface comment, it is not needed, and also either disable its launch through the manual installation or remove it from the list altogether, don't forget that this config is used in all projects where no modems are connected
- replace the deleted patch - general/package/all-patches/wireguard-linux-compat/remove_fallthrough.patch

whoim2:
When compiling a new config, I ran into a situation where one of the name elements must be among ultimate , lite or fpv and it must be located last. This was about majestic.mk, although I still had to make corrections. But a number of commands for trimming the defconfig name, as I understand it, expect a mandatory part like fpv to be located last.

The very name of the config indicates that the assembly is intended for use in conjunction with a 4G LTE modem and at the same time with FPV, more precisely as a blank for such use. In fact, the majestic streamer and a number of tunnels / vpn remained there. And the means for the operation of the e3372 modem were integrated.

PR #655, opened 2023-03-02
added Archer T4U to wifibroadcast startup script

micbuh:
USB id 2357:0101 belongs to TP-Link Archer T4U
https://linux-hardware.org/?id=usb:2357-0101

consider to use the possible device IDs of th driver.


cronyx:
Thanks!

PR #670, opened 2023-03-14
Adjust wireless network configuration

viktorxda:
The idea is to replace the wifi configuration with a script that can be customized to support additional modules.
If a wifi module is found it will open an access point for five minutes, it shows up as OpenIPC with the password 12345678
When connected, a script can be used to adjust the credentials and connect it to a new access point:
```
wireless setup [SSID] [PASS]
wireless connect
```

Ethernet is still accessible at 192.168.2.10 for troubleshooting.

wpa_supplicant requires additional ap support:
`BR2_PACKAGE_WPA_SUPPLICANT_AP_SUPPORT=y`


FlyRouter (changes requested):
Thank you for your work.
Your idea is very interesting, but we can't accept it as it is now.
This is due to the fact that, in order not to break already working devices, you need to do a lot of testing and additions beforehand.
Your work is very valuable for us, I hope you understand the reason for freezing this commit. We can discuss this in the Issues to this PR.

FlyRouter (commented):
We also need to enable the option `CONFIG_UDHCPD=y` in the configuration file `general/package/busybox/busybox.config` to stop this service by default or to bind it to the WiFi mode AP/STA

viktorxda:
I can fully understand your reasoning.
I changed the interfaces file back to the default setting, so eth0 is unaffected by the script changes.
One option might be to add a new generic wifi package that will 'sed' the interfaces to the new scripts.

udhcpd will now also be terminated after the 5min timer.
The current changes are working fine with the SSC338Q, i will test this later with the HI3516EV300 as well.

FlyRouter:
Also remember to enable the DHCPD applet in busybox on this commit.

viktorxda:
Thanks.
I have checked both busybox configs, only busybox-initramfs.config is missing CONFIG_UDHCPD, but i can adjust that if needed.
I assume this could be solved with some additional checks, with ipcinfo / sysfs entries.

FlyRouter:
Great, thanks for the information.
It's just that it was off before, so someone has already fixed it. Add some info to the wiki and your patch will be included with thanks in the release.

FlyRouter (approved):

viktorxda:
I have another suggestion, instead of altering the current overlay files, the new wireless files could be copied from a new package.
This way it won't break the existing devices and the tested wirless devices can have the intended accesspoint functionality by default.

The suggested package would be:
`BR2_PACKAGE_WIRELESS_CONFIGURATION=y`

FlyRouter:
Thank you very much for the final PR and explanations after communication in person. 
After checking the CI, I'll be sure to do a merge.

PR #671, opened 2023-03-14
[Sigmastar] Add simple initramfs that boots rootfs from sdcard

viktorxda:
This commit adds a simple initramfs to utilize a IPL_CUST function that allows to boot a custom UBOOT image from a vfat formatted sdcard.
The booting chain is currently as follows:
IPL_CUST (nand) -> UBOOT (sdcard) -> uImage (sdcard) to a temporary ramfs -> load squashfs (from sdcard) -> switch root to rootfs

Additional needed settings are:
```
CONFIG_BLK_DEV_INITRD=y
CONFIG_INITRAMFS_SOURCE="initramfs/initramfs_list"

BR2_LINUX_KERNEL_EXT_SIGMASTAR_INITRAMFS=y
```


FlyRouter (commented):
Thank you very much for your work.
However, we want to note that it is not allowed to include in PR binary files which are not closed libraries or kernel modules without sources.
We need to think together about further steps to modify this work.

viktorxda:
I understand, i changed the script to have the tools build on demand with the current toolchain.

viktorxda:
I checked a few different configurations to see what is working best and doesn't affect the build when compiling from source.
One issue was that the TARGET_CC environment isn't ready when executing prepare_kernel, but the symlink would be set from per-package/toolchain/host/bin.

PR #673, opened 2023-03-15
Adjust makefile usage printout

viktorxda:


FlyRouter (commented):
Thank you for your work.
My colleagues will consider your suggestions in the near future

FlyRouter:
@themactep @dimerr
as you get a chance, please give your review of the proposed changes

Makefile
```diff
@@ -113,9 +121,10 @@ endef
 # -------------------------------------------------------------------------------------------------
 $(OUT_DIR)/.config:
 ifndef BOARD
-	@echo "Variable BOARD must be defined to initialize output directory" >&2 && exit 1
```
themactep, line 116: Why dropping exit here?

Makefile
```diff
@@ -92,8 +95,13 @@ toolname:
 	@$(SCRIPTS_DIR)/show_toolchains.sh $(FULL_PATH) $(BR_VER)
 
 list-configs:
+ifndef BOARD
+	@echo "Variable BOARD must be defined to list configs"
```
themactep, line 99: then maybe `$(error Variable BOARD must be defined to initialize output directory)` instead, so that it reports the error and exits.

Makefile
viktorxda: Basically for simplification, if the BOARD variable is not defined it just should show the echo.
I tested this with: make $PWD/output/.config

viktorxda:
Updated pull request:
- added proposed changes to use $(error) if variable BOARD is not defined
- increased buildroot version to 2023.02 for infinity6e family

PR #674, opened 2023-03-16
[Sigmastar] Add initial support for infinity6e platform

viktorxda:
Here is an initial implementation for the infinity6e platform, currently intended to work with the CamHi SSC338Q module.

The goal is to have it fully working from the sdcard without ever modifying the internal nand:
The user creates a FAT32 partition and copies UBOOT, uImage.ssc338q and rootfs.squashfs.ssc338q, the device boots and provides an accespoint to enter the setup.

#671 Adds an initial initramfs to load the actual rootfs from the sdcard, it then proceeds with switch_root to the new rootfs.
The initial initramfs doesn't allow pivot_root to change to the overlayfs, that is why this additional initramfs was added.
However, there might be a better solution to save changes/settings without relying on overlayfs.

#670 Adds automatic accesspoint functionality to have the device acessible without needing uart/ethernet.

A current drawback is the manual saving process by entering 'savesettings', it takes a copy of /overlay/root and move it to the sdcard. This copy will be restored before overlayfs is mounted.


FlyRouter:
Thank you very much for the final PR and explanations after communication in person. 
After checking the CI, I'll be sure to do a merge.

FlyRouter (approved):
Great work, thank you !

PR #682, opened 2023-03-18
Remove temporary sdcard scripts

viktorxda:


FlyRouter (commented):
We decided that we will finalize the solution, so this PR will be closed

PR #684, opened 2023-03-18
Move sdcard scripts to proper place

viktorxda:
- added some additional checks to make sure that nothing escapes
- moved both sdcard scripts to the initramfs package
- initramfs boot script is work in progress, but does the job
- wireless-configuration doesn't overwrite files anymore


FlyRouter (approved):
Thank you

PR #686, opened 2023-03-19
Remove restrictive shell history control.

themactep:


FlyRouter (commented):
I wouldn't do the deletion, I would just comment it out. 
Perhaps this feature will be needed someday, or we can use it to do something else.

themactep:
> I wouldn't do the deletion, I would just comment it out. 
> Perhaps this feature will be needed someday, or we can use it to do something else.

The code is in still in the git tree. It can be recovered from the commit, of needed.

PR #696, opened 2023-03-29
[Sigmastar] update infinity6e scripts

viktorxda:
Changes:
- add inotifyd script as part of initramfs to check for overlay changes
- consolidate wireless scripts
- ~~enable busybox wget https feature~~
- move and simplify initramfs build script
- slight adjustments for the initramfs init script
- use proper package path variables


FlyRouter (changes requested):
Using WGET for HTTPS is not encouraged here because it is the base package and enabling the encrypted file transfer option hooks LibOpenSSL.

Please use as in our project Curl + PolarSSL

CONFIG_FEATURE_WGET_TIMEOUT=y
**+ CONFIG_FEATURE_WGET_HTTPS=y**
# CONFIG_FEATURE_WGET_OPENSSL is not set

FlyRouter (approved):
Thank you

PR #703, opened 2023-04-02
Update package mt7601u-ap-openipc

viktorxda:
The native MT7601U kernel driver doesn't work too well with wifi ap mode.
This package isn't currently used and can be updated to a working driver version.

Additional changes:
- add wifi ap placeholder configuration for hi3516ev300_ultimate_defconfig
- enables CONFIG_CFG80211_WEXT to support wireless_send_event


FlyRouter (commented):
We are wary of repositories that belong to unknown users and are too recent. There may be a lot of bugs or some kind of abuse. If it is your repository and you will support it then you can do a transfer to the OpenIPC organization, or make any other suggestions. Thank you.

FlyRouter:
Thank you

PR #704, opened 2023-04-02
Add <gpio> command

jayfan0:
I wrote a simple GPIO script to make controlling GPIO pins more like the commands used in uBoot.

Makes interfacing with gpio simpler.

> Why do:
> 
> ```
> echo <pin> > /sys/class/gpio/export
> echo out > /sys/class/gpio/gpio<pin>/direction
> echo <0|1> > /sys/class/gpio/gpio<pin>/value
> ```
> 
> when you can do:
> 
> ```
> gpio <set|clear|toggle|unexport|read> <pin>`
>```

I'm surprised something like this wasn't already included in busybox. It'll not only make scripts easier to read, but also reduce their file size by ~ 101bytes for every gpio call (taking in to account the size of the script itself it'll pay itself off in ~30 gpio calls).

As suggested, I have also incorporated a search function based on [this script](https://github.com/OpenIPC/sandbox/blob/main/scripts/search-gpio/search-gpio)

```
gpio search <from gpio> <to gpio>
```


FlyRouter (commented):
Beautiful work, you are making the world of OpenIPC a better place

PR #709, opened 2023-04-03
hi3536dv100 as FPV GS

whoim2:
Added kernel config options for usb_joystick, CDC ethernet (usb modem). 
Added wifibroadcast package, zerotier package, mavlink-routerd and remove not-used in FPV GS packages.


FlyRouter (changes requested):
There is a lack of compliance with standards.

Patch is canceled until errors are corrected.

FlyRouter (approved):

PR #710, opened 2023-04-03
[fpv] improvements

cronyx:
* Added new atheros vid pid
* Added mcs change script. Thanks Tipo Man
* Replaced patched firmwares for atheros
* Add ability to set frequency insted of channel (wfb.conf) 
* Remove ralink support


FlyRouter (approved):
Thank you

PR #715, opened 2023-04-05
Added adapter info for T21

jayfan0:


jayfan0:
Thanks. 

- The status light command will not interfere if the gpio pins for each led are not defined. I'll sent a PR to update the defaults to empty. Did you still want to remove it from the adapter script completely? 
- Confirmed 8188fu loads cfg80211 with it. it does not load mac80211 but it seems i no longer need that for my wifi to work. I can remove the 2 extra lines

jayfan0:
all done

FlyRouter (approved):

PR #716, opened 2023-04-05
Parse output from qrscan for wifi

jayfan0:
@cronyx has a package for [quirc](https://github.com/OpenIPC/firmware/tree/master/general/package/quirc-openipc). Currently the package only scans qr codes to retrieve wifi credential data from them, however there is no logic behind it as yet.

This script takes the output of (/usr/sbin/)qrscan, parses the information and after some sanity checks outputs the separate information, ready to be written to the appropriate network files.

It has been tested and works on a clean/ clear QR code image of the [Zxing](https://github.com/zxing/zxing/wiki/Barcode-Contents#wi-fi-network-config-android-ios-11) format:

```
WIFI:T:WPA;S:mynetwork;P:mypass;;
```

Usage:

```
qrparse $(qrscan -p qrcode.jpg | head -n 1)
```
Full usage:

```
curl -s -o /tmp/img.jpg http://127.0.0.1/image.jpg ; qrparse $(qrscan -p /tmp/img.jpg | head -n 1)
```

Current issues with the entire pack:

- `qrscan` currently outputs an empty second line, therefore `| head -n 1` is required at the moment
- `qrparse` works on a clean, clear qr image, however on my camera the image sensor cannot capture a clear enough image for `qrscan` to successfully extract data from
- `qrparse` currently `echo`'s the fields to terminal output, stdout. An appropriate file needs to be decided. Possibly aligned with @viktorxda's [wireless-configuration ](https://github.com/OpenIPC/firmware/blob/454b15a1a2e62cd680a4c8a37b9a93b38b2482ff/general/package/wireless-configuration/files/script/wireless#L2) package (at `/etc/wireless.conf`), which is making good progress on the AP mode functionality.

Discussions on best course of action required

### DIFF
--- a/general/package/quirc-openipc/files/qrparse
+++ b/general/package/quirc-openipc/files/qrparse
@@ -1,0 +1,113 @@
+#!/bin/sh
+
+################################################################################
+#
+# THIS SCRIPT WILL PARSE AN ZXING FORMAT QR CODE PAYLOAD FOR WIFI CREDENTIALS
+#
+# INPUT MUST BE PASSED IN SINGLE QUOTES
+#
+# BARCODE CONTENTS CAN BE FOUND HERE
+# https://github.com/zxing/zxing/wiki/Barcode-Contents#wi-fi-network-config-android-ios-11
+#
+# THIS SCRIPT USES SIMILAR LOGIC TO:
+# https://github.com/zxing/zxing/blob/master/core/src/main/java/com/google/zxing/client/result/WifiResultParser.java
+#
+# @author jayfan0
+#
+# OPENIPC USAGE: qrparse $(qrscan -p /mnt/mmcblk0p1/qrcode2.jpg | head -n 1)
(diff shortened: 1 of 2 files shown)


FlyRouter (approved):

cronyx:
With reading QR codes in automatic mode, you need to be extremely careful and accurate, because there are many vulnerabilities and potential injections associated with it. It is necessary to think over the logic of work and the analysis of the read data before using the data received from the QR.

jayfan0:
Please do try the script. It only takes one argument and checks the validity of the format
`qrparse  'WIFI:T:WPA;S:mynetwork;P:mypass;;`

It's a shell translation of Zxing's own java parser on their github

I've tried adding commands in there to see if they execute. The logic seems fit, but that's what testing is for!

cronyx:
> Please do try the script. It only takes one argument and checks the validity of the format `qrparse 'WIFI:T:WPA;S:mynetwork;P:mypass;;'`
> 
> It's a shell translation of Zxing's own java parser on their github
> 
> I've tried adding commands in there to see if they execute. The logic seems fit, but that's what testing is for!

Thanks, I'll try it as soon as I find a suitable camera with wifi to tests.

PR #718, opened 2023-04-08
Add reset daemon for cameras with gpio buttons

jayfan0:
Some cameras have a general purpose reset button attached to a GPIO pin. This script is called in `/etc/init.d`, and if the reset button GPIO is configured, checks to see if it is pressed 4 times per second.

If the button is counted to have been held for 5 seconds, the `firstboot` command is called, resetting the camera.

### DIFF
--- a/general/overlay/etc/init.d/S10resetd
+++ b/general/overlay/etc/init.d/S10resetd
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+case "$1" in
+  start)
+    echo -e '\nLoading resetd...'
+    /usr/sbin/resetd &
+    ;;
+esac
--- a/general/overlay/usr/sbin/resetd
+++ b/general/overlay/usr/sbin/resetd
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+# Set Reset switch GPIO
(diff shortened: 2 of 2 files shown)


FlyRouter (commented):
I'll just leave it here...

https://github.com/larsks/gpio-watch

FlyRouter:
I'm not sure that the interrupt daemon written in pure C will always work and is needed by everyone. It's just that there are two ways to do it. Either way, your script is worthy of being saved and possibly implemented because it will work everywhere. The implementation as in my example at the link sometimes did not work correctly on some systems and caused the boards to freeze, all this needs to be tested.

FlyRouter:
We use the mdev daemon in the firmware to handle various events when mounting flash drives and cards.
It may be worth trying to implement support for gpio as well. Here are some links for you to think about.

http://derekmolloy.ie/kernel-gpio-programming-buttons-and-leds/
https://stackoverflow.com/questions/25507750/detecting-interrupt-on-gpio-in-kernel-module
https://habr.com/ru/articles/419273/

FlyRouter (approved):
Thank you

PR #721, opened 2023-04-10
Add SSC338Q to matrix_build.yml

viktorxda:


widgetii:
It is failing currently in 'Warning, rootfs.squashfs size exceeded by 2097152... ssc338q (lite)', there are two ways to fix it:
1) Fit rootfs.squashfs into 8Mb
2) Disable this check for the platform (and specify somewhere in docs that only 16Mb flash should be used with ssc338q)

PR #753, opened 2023-04-29
small updates for sysupgrade/wireless-configuration

viktorxda:
- Exposing an accesspoint with a camera is a potential security risk, limiting the time to 3 minutes should be sufficient for the initial setup.
- Summarize get_flash_type, puts the nand bootcmd and ipcinfo flashtype check in the same function.


FlyRouter (approved):
Thank you !

cronyx (approved):

cronyx:
With changes in the logic of sysupgrade, it is very desirable to make [bump version] (https://github.com/OpenIPC/firmware/blob/master/general/overlay/usr/sbin/sysupgrade#L6) so that the script is self updated before starting the firmware update.

PR #760, opened 2023-05-02
separate configs AIR and GS for datalink/wifibroadcast packages and add rcjoystick package

whoim2:
requested changes for '[separate configs AIR and GS for datalink/wifibroadcast packages](https://github.com/OpenIPC/firmware/commit/c8176a5f8be244beba268d9abbedbdfc1c6bb070)' and rcjoystick


cronyx:
Thanks!

PR #763, opened 2023-05-05
get rid of vendor-specific package directories

themactep:
Use packages directly from general/package/


themactep:
I would go further and made BR2_EXT_PACKAGES_PATH including the /package bit. It would save some code. But at the same time I am not a fan of the idea of adding more variables than really necessary. I wonder if buildroot already has its own variable for that.

Besides, my building aproach is to separate git sources and buildroot, and spit outputs per board, and when i did it for the first time BR2_EXTERNAL stopped working for me so I had to use vendor specific external directory as an anchor for futher traversing. @viktorxda, you might want to check Makefile in my repo if you are curios. Maybe you could have an idea how to amend it further.

themactep:
@viktorxda makes a better PR. We better use his.

PR #764, opened 2023-05-05
first stage to split majestic and vendor specific scripts

cronyx:
[All platforms affected in the future]: First stage to split majestic and load_vendor specific scripts.
[Ingenic T31]: Sync settings of resolution betwen streamer and sensor module (reboot after change required). GC4653 as example. Not all sensors have this feature.
[Other]: Mac checking moved out from vendor specific scripts to extutils. Load vendor specific drivers moved to S70vendor. For vendor detect use ipcinfo now.

All ideas for improvements are welcome.


FlyRouter (approved):
Great job, thank you !

general/package/ingenic-osdrv-t31/files/sensor/gc4653.yaml
```diff
@@ -4,3 +4,4 @@ sensor:
   width: 2560
   height: 1440
   bus: i2c
+  resolutions: 2560x1440x25_400,1920x1080x60_300,1280x720x60_100
```
themactep, line 7: If we want to keep it .yaml then syntax should be
```
resolution:
  - 2560x1440x25_400
  - 1920x1080x60_300
  - 1280x720x60_100
```
Otherwise there is no point to keep it .yaml, a simple .ini syntax would work better.

general/overlay/etc/init.d/S70vendor
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# load vendor specific drivers
+# 
+vendor=$(ipcinfo -v)
+
+case "$1" in
+  start)
+	logger -s -p daemon.info -t ${vendor} "Check MAC for ${vendor} devices..."
+	logger -s -p daemon.info -t ${vendor} "$(check_mac)"
```
themactep, line 10: Logging could be refactored further to a function. Something like
```
log_vendor() {
  logger -s -p daemon.info -t ${vendor} "$1"
}
```

general/package/ingenic-osdrv-t31/files/sensor/gc4653.yaml
cronyx: > If we want to keep it .yaml then syntax should be
> 
> ```
> resolution:
>   - 2560x1440x25_400
>   - 1920x1080x60_300
>   - 1280x720x60_100
> ```
> 
> Otherwise there is no point to keep it .yaml, a simple .ini syntax would work better.

I tried not to produce new configs, and immediately wrote it into the sensor config.

themactep: I didn't mean it a new config. Just the notation. An array of values in yaml is one per line with a dash at the beginning of the line item.

general/overlay/usr/sbin/extutils
```diff
@@ -21,6 +24,14 @@ case "${CMD}" in
...
 	fi
 	$IPCTOOL $@
+  ;;
+  check_mac)
+	if [ "$(fw_printenv -n ethaddr)" = "00:00:23:34:45:66" ]; then
+		status="The eth0 interface has a lousy MAC, please change it.."
+	else
+		status="The eth0 interface has a correct MAC - $(fw_printenv -n ethaddr)"
+	fi
+    echo ${status}
```
themactep, line 34: If `$status` is not used elsewhere, just drop the assignment and use two `echo` instead.

themactep (approved):

general/overlay/etc/init.d/S70vendor
dimerr: I would also replace messages.
`Check MAC ... `- that line is not necessary at all
`The eth0 interface has a correct MAC` - remove
Idea is to leave only 'error' message, and replace that message as follows:
`The eth0 interface has a lousy MAC, please change it..`  -> `Warning. Wired network interface has default MAC address, please change it.`

general/overlay/usr/sbin/extutils
cronyx: As i known, yaml-cli doesn't work with arrays, maybe need to add this functional.

PR #781, opened 2023-05-09
[General] remove external vendor unknown and placeholder folders

viktorxda:
Removal doesn't affect the build, so I assume they are only intended for make board-info.
It is planned to overhaul the buildscript/Makefile, so it should be fine to remove this function for now.


dimerr:
thank god someone did this finally

dimerr (approved):

PR #791, opened 2023-05-12
Makefile: overhaul usage and commands

viktorxda:
First draft of a makefile overhaul, build breakage is to be expected, need to see how feasible this approach is.


FlyRouter (commented):
Unfortunately, the binding of platforms to different versions of BR is broken here and more research will be needed.

PR #802, opened 2023-05-18
reuse dropbear package from buildroot

themactep:
Recent version of dropbear-openipc package is a copy of dropbear package from buildroot-2022.11.2. The only difference is DROPBEAR_SITE URL.


themactep:
Leave it till we have more recent buildroot everywhere.

FlyRouter (approved):
If you have done the tests and are confident in the action then I have no objections, thank you.

themactep:
There will be regression for builds using older buildroot, this is why i wanted to postpone it till later, but it will catch up as we transition to the recent version, which is way overdue.

PR #803, opened 2023-05-18
replace fwprintenv-openipc with uboot-tools from buildroot

themactep:
fwprintenv-openipc package supposedly was forked from uboot-tools package of buildroot-2020.04 to limit the number of installed tools. Modern uboot-tools package does that using buildroot variables, and by default installs only the two tools that fwprintenv-openipc wants. The extra patch file is moved to all-patches/uboot-tools.


FlyRouter (approved):
This is quite a serious and dangerous procedure, since this utility is the basis for changes and work. If you have done the tests and are confident in the action then I have no objections, thank you.

PR #804, opened 2023-05-18
[sysupgrade] 1. added key to select dev branch. 2. use downloaded ipcinfo…

cronyx:
… if it exist into tmp folder.


FlyRouter (approved):
Great work !
Thank you !

PR #808, opened 2023-05-19
[no ci] fix hi3536dv100 rootfs oversize

cronyx:


FlyRouter:
I wonder what the increase in size of rootfs has to do with it, before the WEB-interface seemed to fit normally.

PR #825, opened 2023-05-26
General: set separate release name for lte devices

viktorxda:


general/scripts/executing_commands.sh
```diff
@@ -19,9 +20,9 @@ if [ ${BR2_EXTERNAL_LIBC} != "glibc" ] && \
 fi
 
 if [ ${BR2_EXTERNAL_LIBC} = "musl" ]; then
-  exlist=${BR2_EXTERNAL_SCRIPTS}/excludes/${OPENIPC_MODEL}_${OPENIPC_RELEASE}.list
-  if [ -f ${exlist} ]; then
-    for e in $(cat ${exlist}); do
+  LIST=${BR2_EXTERNAL_SCRIPTS}/excludes/${OPENIPC_MODEL}_${OPENIPC_RELEASE}.list
+  if [ -f ${LIST} ]; then
```
themactep, line 24: What about replacing the loop and concatenation with working in-place with a list?
```
cd $TARGET_DIR
xargs -0 rm -f < $LIST
cd -
```
Not tested, just an idea!

PR #869, opened 2023-06-09
Package: use head branch by default for openipc

viktorxda:


cronyx:
In mavfwd the version is intentionally fixed, in the repo there are changes that should not yet be included in the builds.

PR #873, opened 2023-06-11
Linux: download archive based on git sha

viktorxda:
Extends PR #872


general/external.mk
themactep, line 1: Love the solution. My only concern is a branch name for the kernel. Let's replace wildcard `*$(OPENIPC_SPC_FAMILY)` with a `$(OPENIPC_SOC_VENDOR)-$(OPENIPC_SOC_FAMILY)`. Some vendors like Ambarella use very short names for their processors that could possibly cause a flase positive match somewhere down the road.

viktorxda: Indeed, this sounds reasonable. I added the suggested changes.

PR #879, opened 2023-06-24
Sigmastar: update infinity6b0 package

viktorxda:
First set of changes to sync the infinity6b0 package (sigmastar-osdrv-ssc335) with the general layout of infinity6e.


FlyRouter (approved):
I see a hell of a lot of changes and I have questions. ;)
It's not a complaint, no need to undo your edits, but I want more clarification on many points. I'll describe them below.

FlyRouter:
1. I see you deleted a bunch of old kernel configs. I guess you didn't find anything interesting in them, right? I collected them from various devices and experiments and I don't mind deleting them. Just have to keep in mind that someday you might need to look into them in the archive for some tricky function like RTC and so on that I implanted earlier.

2. I see you removed support for F2FS, I don't know if that's right, look at other cores on other CPUs. It's the only file system that works adequately on SD cards and doesn't kill them. We don't know how to format SD cards to F2FS yet but if we learn how, it will be the best choice for video recording.

3. I see you've converted support for RTL8188FU to an external repository. I want to emphasize that this chip is used in many types of cameras with different processors and we need not to lose compatibility in any way. i hope you have done some testing. And also I propose to move the repository with RTL8188FU driver from you to us in OpenIPC as we did for other chips, but that's when you have a chance.

Thank you !

viktorxda:
1. Yes, this is mostly intended to reduce the amount of configs, but I noted the commit id in case they might be later needed or come in handy. Maybe this can be added to the wiki instead as a general note, especially the rtc component could be important.

2. The idea is to replace F2FS with exfat instead, it uses an external kernel driver from the official exfat linux maintainer. The driver and tools are around ~200kB, I tested it on the SSC338Q with no major problems.

3. This source of the RTL8188FU is based on version v5.11.5 and doesn't require the firmware binary. I tested it on my Foscam X3E model (SSC337DE). I noticed it is used by some HiSilicon ultimate models, I have not the hardware available to test the compatibility with those, but I would assume it should work as well, since the driver is basically of the same origin.

About storing the driver source, I would like to suggest to rename/reuse https://github.com/OpenIPC/rtl8733bu as a general repo for realtek drivers. I have uploaded the source for both rtl8188fu/rtl8733bu.

FlyRouter:
Thank you for your answers
I put the firmware back together in Actions which ended badly

viktorxda:
Yes, the repo for the realtek wlan driver needs to be adjusted.
I have changed the repo address back to the previous values.

FlyRouter:
> Yes, the repo for the realtek wlan driver needs to be adjusted. I have changed the repo address back to the previous values.

I think we should create a separate package in which to specify the repository with the new driver and gradually translate all the devices testing, so that there is a choice. It will be a longer but at the same time correct way.

viktorxda:
The realtek drivers are separated now: rtl8733bu and rtl8188fu.
rtl8188fu-openipc is unchanged to allow compatibility with existing configurations.

PR #885, opened 2023-06-28
Ingenic T31 atbm603x wifi drivers + wireless-configuration support

cocus:
Adds the "wireless-configuration" package by default on T31 ultimate.
Adds the precompiled binary for the atbm603x sdio wifi module.
Modifies the "wireless-configuration" script that detects the WiFi module to add support for the atbm603x SDIO on T31's.

Tested on a T31X (clone of a Wyze Cam Pan V3) and a T31AL from @themactep (although with the lite version, and some packages removed to make the wifi driver fit).


cocus:
I guess the wifi doesn't fit for lite. I should make the script install the ko files only for ultimate

cocus:
With the new set of changes, the module will only be installed on the t31_ultimate. You also have a new config to select the mmc0 or mmc1 bus (mmc1 being the default on t31_ultimate). I'm using @themactep 's atbm kernel module for the mmc0, since both are built from the same source, with the only difference being

viktorxda:
I currently don't have a matching device to test the driver, but I appreciate the idea to have a potential solution for the altobeam wifi adapter.
One thing to note: there is some reworking for the wireless-configuration package in progress, so that might need some re-testing on a later time.

FlyRouter (changes requested):
@cocus 
Please correct the path to the new initialization data. 
The device profiles are now located here - firmware/general/overlay/etc/wireless
Thank you !

cocus:
Working on this. I really like the new mechanism for loading the appropriate driver

cocus:
So, there's a bug on the current wlan0 script on interfaces.d; see, my wifi SSID has a space on it (Cocus WiFi), but the script doesn't like it because of this:
`post-up wpa_passphrase $(fw_printenv -n wlanssid || echo openipc) $(fw_printenv -n wlanpass || echo openipc12345) > /etc/wpa_supplicant.conf`
As you can see, wpa_passphrase thinks "Cocus" is the ssid, and "WiFi" is the password (completely disregarding the actual password).
I've also noticed that even if the wlan0 device appears using the new `wlandevice`, it never gets configured (I have to manually run `ifup wlan0` to get to the wrong state I mentioned before)
I'll see if I can fix this as well, but I'm a little bit baffled as to why wlan0 didn't auto-up (wlan0 states `auto wlan0`, so it should work!)

FlyRouter:
We can coordinate these questions and actions with @viktorxda  because he is the author of the idea of switching to the new script

themactep:
> So, there's a bug on the current wlan0 script on interfaces.d; see, my wifi SSID has a space on it (Cocus WiFi), but the script doesn't like it because of this: `post-up wpa_passphrase $(fw_printenv -n wlanssid || echo openipc) $(fw_printenv -n wlanpass || echo openipc12345) > /etc/wpa_supplicant.conf` As you can see, wpa_passphrase thinks "Cocus" is the ssid, and "WiFi" is the password (completely disregarding the actual password). I've also noticed that even if the wlan0 device appears using the new `wlandevice`, it never gets configured (I have to manually run `ifup wlan0` to get to the wrong state I mentioned before) I think un-commenting the `#auto wlan0` to `auto wlan0`, and double-quoting the output of the fw_printenv calls should fix it. Do you want me to add this to this fix PR?

Try this.
`post-up wpa_passphrase "$(fw_printenv -n wlanssid || echo openipc)" "$(fw_printenv -n wlanpass || echo openipc12345)" > /etc/wpa_supplicant.conf`

cocus:
> Try this.
> `post-up wpa_passphrase "$(fw_printenv -n wlanssid || echo openipc)" "$(fw_printenv -n wlanpass || echo openipc12345)" > `

Yes, this works, but I still need to do an `ifup wlan0`, even if I have the "auto wlan0"

themactep:
Any errors/hints in dmesg?

cocus:
> Any errors/hints in dmesg?

Absolutely nothing. wifi kernel module loads fine, automatically, but no errors or weird messages. it's just that wlan0 doesn't get up automatically.

cocus:
Ah, I think  I know why! I've called modprobe with a & on the end so it "keeps working it in the background", because if the wlan driver fails to load, it gets to a never-ending loop, and modprobe just stalls. This almost always happens when you keep the UART TX plugged in and fails to load. I think I do need to wait for wlan0 to come up so the wlan0 gets ifup'd automatically.

viktorxda:
See the wireless init script: [S30wireless](https://github.com/OpenIPC/firmware/blob/master/general/overlay/etc/init.d/S30wireless), it checks for the exit code to modify the adapter files (although this is currently a placeholder solution).
Maybe you can strip down the wireless script a bit, to fit the overall 'theme' of the other devices.

themactep:
@viktorxda can we stick to a conventional syntax in sed and use slashes instead of pipes? Thank you.

viktorxda:
As far as I am aware it doesn't matter which delimiter to choose, the intention is to make it as clearly as possible.

themactep:
It is a question of clarity. Secondary delimiter is justified only when the primary one is used within regexp, and even then the most common choice would be either # or !. Using pipes is misleading at best.

viktorxda:
All right, I can agree with this explanation. I will adjust it with the next pull request.

general/overlay/etc/wireless
```diff
@@ -83,4 +83,42 @@ if [ "$1" = "l716-t31-camhi" ]; then
...
 
+# T31 ATBM603X on SDIO bus mmc0
+if [ "$1" = "atbm603x-t31-mmc0" ]; then
+	# PB10 in wifi mmc1 is in the pull-down state by default, it needs to be set to a high-impedance state
+	devmem 0x10011128 32 0x400
+	# set wifi mmc1 clk drive capability to 8mA
+	devmem 0x10011134 32 0x20000
+
+	# "insert" the bus mmc0 device
+	if [ "$(cat /sys/devices/platform/jzmmc_v1.2.0/present)" = "N" ]; then
```
cocus, line 94: this is important, it's not an "identification" procedure, just a mechanism to make the mmc (SDIO) driver enumerate the devices if it haven't.

same on line 112

cocus:
Can this be reviewed? I can update the branch again, but it's getting out of date constantly because this is not reviewed.

general/overlay/etc/network/interfaces.d/wlan0
```diff
@@ -1,4 +1,4 @@
-#auto wlan0
+auto wlan0
```
viktorxda, line 1: This needs to be changed back, having the interface active by default will cause compatibility issues with other devices.

FlyRouter (approved):

FlyRouter (approved):
I have no objection except for a few questions:
- you need to fix the interface status as Viktor wrote in the last comment
- how relevant will it be to have altobeam driver in ultimate firmware permanently for T31 ?
- will we be able to replace the binary drivers of the WiFi module in the future with a source build as we do everywhere else ?

The right of control over this PR is given to Viktor ;)

cocus:
> * how relevant will it be to have altobeam driver in ultimate firmware permanently for T31 ?
Well, it's mostly the same as having the MT and the ralink drivers. This seems to be useful for some cameras, including the Wyze Cam Pan v3.

> * will we be able to replace the binary drivers of the WiFi module in the future with a source build as we do everywhere else ?
Yes! As soon as Wyze releases the sources, I'll replace the precompiled binaries with our own kernel modules. (I have a ticket open with Wyze, and they were quite good and told me in a few days they'll release everything for this camera, including the wifi driver, I hope!). In the worst case, we just need to keep modifying the sources we've already gathered, until they load the firmware of the atbm and don't crash the system :).

I'll fix the wireless file and it should be good to go.

FlyRouter:
Thank you for your answers, I have no further questions or objections to the integration.
Thank you very much for your work !

viktorxda:
Personally, I think having the proper source is preferable, if you can later add it to replace the binary modules this would be an ideal solution.
I just wait for the CI to finish and then I think we can merge it.

cocus:
> Personally, I think having the proper source is preferable, if you can later add it to replace the binary modules this would be an ideal solution. I just wait for the CI to finish and then I think we can merge it.

I also think the sources are a better solution, but the problem is that the ones I tried, seem to load the firmware on the atbm, but crashes the kernel. Depending on how much I tinker with them, I get them to load it, but no wlan interface, so... they're "not" ready to got. That's why I'm relying on Wyze's sources, because these should work fine

viktorxda:
Yes, that is what I was referring to. If the wyze sources are actually sound, this would be ideal.

PR #894, opened 2023-07-01
Package: add global variable to request git sha

viktorxda:


FlyRouter (changes requested):
I think there is a problem because the method of getting the source code is not specified, in our case it should be "git".

FlyRouter (commented):
We need to get the sources through GIT, because only there are blanks and labels we need in some cases, as well as you can track the latest version.
We experimented for a long time and settled on the current version.

viktorxda:
The same technique is already being used by the kernel, tarballs are downloaded by wget, which is the default setting.

FlyRouter (approved):
Thank you

PR #902, opened 2023-07-06
Ingenic: add preliminary support for CIP-37210AT

viktorxda:


FlyRouter (commented):
I don't really like the action here where the kernel includes WiFi and VFAT modules because there are modules on this processor without Wireless and these options would be superfluous, and also we need to stick to some rules about these modules. Nowadays, as far as I remember, everything is done by modules everywhere and not in the kernel. What do you think, @cronyx

viktorxda:
- Adding the modules back to the kernel is intended for #865
- If no wireless modules are included it gets automatically removed: [linux-parcher](https://github.com/OpenIPC/firmware/blob/master/general/package/linux-patcher/linux-patcher.mk)
- t21.generic.config is only used by one configuration which uses wireless by default.

viktorxda:
Kernel changes were reverted.

PR #904, opened 2023-07-07
Overlay: update wireless-settings

viktorxda:


general/overlay/etc/network/interfaces.d/wlan0
```diff
@@ -1,6 +1,6 @@
 #auto wlan0
 iface wlan0 inet dhcp
-    post-up wpa_passphrase $(fw_printenv -n wlanssid || echo openipc) $(fw_printenv -n wlanpass || echo openipc12345) > /etc/wpa_supplicant.conf
+    post-up wpa_passphrase "$(fw_printenv -n wlanssid || echo OpenIPC)" "$(fw_printenv -n wlanpass || echo password)" > /etc/wpa_supplicant.conf
```
themactep, line 3: Length of "password" password is insufficient to satisfy WPA(SAE) requirements. We need at least 8 chars.

themactep: lol. scratch that. seems i am bad at math.

themactep (approved):

general/overlay/etc/network/interfaces.d/wlan0
viktorxda: Using 'password' is only mostly intended to show the usage of the command.
However, it could be a (unlikely, but possible) security risk, so it might be reasonable to change it to something entirely different.

FlyRouter (approved):
Great patch, great job, thank you all !

PR #913, opened 2023-07-14
Add a new T31 kernel module for motors.

cocus:
Set default GPIO values for the Wyze Pan Cam V3.
Modify the motors package to build the newly added "t31-kmotor" app.

This requires https://github.com/OpenIPC/motors/pull/7 to be merged.


FlyRouter:
Thank you very much, I made a merge of your work in motors repo.

Maybe I can join your experiments with motors, I have found some hardware, they just need to be placed on stands.

Unfortunately, what you propose in the current PR doesn't look good due to the large number of variables. Maybe you should merge them into one line with some separator ?

if you need a separate repository feel free to PM me on telegram as @FlyRouter.

Thank you for your work we really appreciate it !

cocus:
I bet you're right and I should pass the gpios as parameters on the actual module rather than on build time. It's just the way they were programmed on the original sources. I can change that to make it more user friendly.

FlyRouter:
Thank you for your understanding.
I can already see in the end that it will be a good job, sir !

cocus:
Okay, this is pretty cool. I figured I don't need the PWM stuff to get the motors running, since the motors driver actually uses a single timer (seems like tcu channel 2.2), and no PWM is required. So I took the PWM code out.
Secondly, I made all the GPIOs as arguments. These are:
* hmin, vmin: Motor start point
* hmax, vmax: Motor stop point
* hlevel, vlevel: Motor IRQ "style" (not used, so I don't know what's the deal)
* hst1, vst1: Motor Phase A
* hst2, vst2: Motor Phase B
* hst3, vst3: Motor Phase C
* hst4, vst4: Motor Phase D
By default, the values would be "good to go" for the Wyze Pan Cam v3, but if you have another setup, you do need to pass these arguments while modprobing (or insmoding) the driver, like for example:
`modprobe sample_motor hmaxstep=2130 vmaxstep=1600 hst1=57 hst2=49 hst3=51 hst4=54 vst1=61 vst2=60 vst3=63 vst4=62`

You are still required to pass the `hmaxstep` and `vmaxstep`, because these are the maximum steps allowed for each motor.

themactep:
No rebuke means approval :) Merge it, and if it does not work or if breaks something, there will be chance to fix it.

cocus:
> No rebuke means approval :) Merge it, and if it does not work or if breaks something, there will be chance to fix it.

Right ahead Sir!

ZigFisher (commented):
Thank you

PR #919, opened 2023-07-20
Package: update realtek wlan drivers

viktorxda:
The updated RTL8188FU driver was successfully tested with two different boards, it should be working with the remaining boards as well.


FlyRouter (approved):
I have no objection.

However, then I propose to move the driver repo to our project, so that it would be identical to everything else.

viktorxda:
All right, that seems reasonable.
Please rename https://github.com/OpenIPC/rtl8733bu to realtek-wlan and delete the master branch, thanks.

PR #929, opened 2023-07-31
[no ci] Overlay: separate wireless device list

viktorxda:


FlyRouter (commented):
I have some objections and suggestions for this PR.

1. The objection is due to the fact that the wlan0 configuration file is moved outside /etc/network/interfases.d. This directory is for all network configurations and this is a policy position, they should only be there.
These files are used in many packages, in the WEB-interface and it is impossible to break the order. Thanks.

2. I have no objections to the device class division of the configuration files, but I would suggest using the name "rndis" or "modem" for your current "serial" class. Thanks.

3. it seems a bit wrong to me that the scripts are now inside the /etc/network/* directory. Their work could be related not only to the network, but also to some other devices connected to the camera that require detection. Today we have SDIO, USB and RNDIS/MODEM, and tomorrow something else may appear. Please consider suggesting something like /etc/profiler where we can put these scripts-configs. This is not a required title, just a suggestion. Thanks

4. There are already a lot of projects using OpenIPC as a chassis and it's time for us to stop with the constant redesign of the networking structure, as it's already causing a number of questions and confusion. We all need to learn to work more on concepts and be able to announce changes if they are really needed. I hope you understand correctly, this is in no way a complaint. Just need stability since we are dealing with embedded devices. Thanks

viktorxda:
1. I restored the original wlan0 interface file.

2. Serial was renamed to modem.

3. The 'wireless' folder was moved to /etc/wireless. /etc/profile is unfortunately executed after the initial boot, so the variables are not initialized during the network setup.

4. No problem, that is what the PR is for, to gather feedback and suggestions. The current wireless functionality is not going to be changed, this is more intended to separate the devices before we gather too many devices inside one script/list.

FlyRouter:
Thank you very much for your understanding and great work !

viktorxda:
> 3. it seems a bit wrong to me that the scripts are now inside the /etc/network/* directory. Their work could be related not only to the network, but also to some other devices connected to the camera that require detection. Today we have SDIO, USB and RNDIS/MODEM, and tomorrow something else may appear. Please consider suggesting something like /etc/profiler where we can put these scripts-configs. This is not a required title, just a suggestion. Thanks

I understand now, the profiler script is under /usr/sbin and works in a similar fashion as the usb/modem/sdio scripts.
It doesn't seem like the profiler script is currently in use or called by any of the init.d scripts. If you wish/prefer, I can integrate the wireless scripts into the profiler.

PR #940, opened 2023-08-09
Make sd CARD work

naksper:
NLS ISO8859 is required for SD card to work
If it is not enabled, it gives an error when mounting:
FAT-fs (mmcblk0p1): IO iso8859-1 charset not found


ZigFisher (approved):
Thank you, sir

PR #942, opened 2023-08-10
[skip ci] Update wlan0, fix typo

ZigFisher:


ZigFisher:
I realized where we got the problem. 😊

You're looking at it from the perspective that there's a profile for the device, it's executed and everything works, I agree with that.

But the user above had a custom configuration where he manually added commands fixing "post-up". of course it won't work and he needed to fix it to "pre-up".

So everyone was right in their own way, it's fine.

PR #947, opened 2023-08-11
[no ci] Overlay: export soc environment variable

viktorxda:


FlyRouter:
Hi !
When you get a chance, please give a little more explanation

- I understand that now there are only two cores for i6b, the rest are symlinks.
- What are the options changes in ssc338 core related to, just wondering.
- As far as I understand, the firmware package for mt7601 is not used anymore.
- What is the reason for removing the SOC=$(fw_printenv -n soc) variable?

I have no objections to the actions but would like to be kept up to date on the details, thanks ;)

viktorxda:
Hello, I checked the differences between the families on the kernel configs and they are basically identical with minor differences. For the SSC335 and SSC337 is only boils down to the used device tree file.

The changes made on the SSC338 config are mostly intended to have both families as similar as possible in settings and options.

The external MT7601U driver needs the MT7601USTA.dat file to read some of the configuration data from it, this is not needed for the built-in kernel driver.
The firmware file mt7601u.bin is needed by both, the package is still available but was simplified, as we only need three small files from a ~500MB archive: https://github.com/OpenIPC/firmware/commit/73dd5a6182b239dabcc78efe32cc9782ab1f4efb

The SOC variable was introduced to identify the Sigmastar chipsets with the environment variable instead of the ipctool output.
However, if microbe-web uses ipcinfo/ipctool, it cannot read the environment variable and outputs the default guessed chipsets based on the family id (SSC336X/SSC338X/SSC339X).
This was changed on ipctool itself, so it can now properly read the uboot 'soc' variable.

FlyRouter:
Thank you so much for your lovely comment.
It is very nice to see not only your work, but also explanations that are easy to understand.
Good luck with your experiments and research !

viktorxda:
One thing I forgot to mention, the MT7601U is currently not used by Sigmastar, if this changes in the future we can enable the driver and add the firmware back in.

PR #961, opened 2023-08-20
rtl8189fs-openipc: add support for Ingenic T31 to wlan driver

gtxaspec:
On T31 devices,  we need to modify the wlan driver to force re-scan the SDIO module on the mmc bus.

These changes enable the `rtl8189fs-openipc` package to work with the T31 platform.

( note, you may need to enable the GPIO first before loading the wlan driver.  Normally, the driver would enable the GPIO, but this is disabled to enable generic comparability across a broad range of devices, as the GPIO's would otherwise have to be hard-coded into the driver.  )


gtxaspec:
Going to revise, let's close for now

PR #963, opened 2023-08-21
[no ci] add sc4236 sensor driver

gtxaspec:
add sc4236 sensor driver to load_ingenic (driver module is already included), sensor bin, and sensor yaml


ZigFisher (approved):
Thank you

PR #964, opened 2023-08-22
[no ci] overlay: add rc.local capability

gtxaspec:
Improved Startup with rc.local Scripts.  Added S99rc.local to /etc/init.d/ and created /etc/rc.local. This lets users easily run their own commands at the end of startup. It's a straightforward way to customize and optimize the system's start-up process.


ZigFisher (approved):
Thank you

PR #965, opened 2023-08-22
overlay: wlan0: set mac address from uboot env variable

gtxaspec:
Introduced a feature that lets the user customize the MAC address of the wireless device (wlan0). Here's how it works:

Added a setting in U-Boot called wlan_addr to `/etc/network/interfaces.d/wlan0` and `/etc/wireless/interfaces/wlan0`.  

If the user provides a MAC address in wlan_addr, the wlan0 device will use that address when it starts up. 

This is useful because sometimes the MAC address printed on a device doesn't match what the device's wireless hardware actually uses, example Wyze and Atom device vendors. 

The user can choose their own MAC address if desired. If wlan_addr is empty or not set it at all, the device will just start up normally with the default MAC address.


ZigFisher (approved):
Thank you.
This is similar to the principle we use for Ethernet, but in reverse.
`hwaddress ether $(fw_printenv -n ethaddr || echo 00:00:23:34:45:66)`

PR #966, opened 2023-08-22
ingenic-osdrv-t31: Improve image quality by replacing gc2053-t31.bin sensor param file from stock fw

gtxaspec:
Update `gc2053-t31.bin` with sensor param file from stock FW.  The result is Improved image quality: better contrast, saturation and luminance.


ZigFisher (commented):
Thank you.
This patch will require additional testing as devices with this sensor are already in use and we need to make sure that nothing will be broken.

gtxaspec:
@ZigFisher Thank You!

If anyone else would be able to test on their devices, the feedback would be appreciated.  I tested on: 

Wyze V3
Wyze Doorbell
AtomCam 2

viktorxda:
Just from very quick test on a Wyze V3, the image quality is indeed improved.
I assume this is only a image parameter calibration file, something similar is used for Sigmastar as well.

ZigFisher:
Test passed, thank you all

PR #970, opened 2023-08-24
Add support for jxf23 sensor for T20

bengris32:


ZigFisher (commented):
I have no objections to the work done, thank you.

However, there is a question about the relevance of the sinfo.ko file and whether there will be problems and collisions with other sensors.

Thank you.

bengris32:
That's perfectly understandable, I did check the sensors present in the older prebuilt sinfo module and found that all the sensors present in there are also present in the module from openingenic (If that's what you're referring to here).
But yes, this should really be checked to see if it breaks other sensors.

gtxaspec:
is the sinfo.ko bundled with openipc not the version from openingenic? https://github.com/OpenIPC/openingenic/blob/master/kernel/sinfo/sensor_info.c

bengris32:
I don't think so, the sinfo module for T20 currently seems built straight from the T10/T20 SDK, at least that's what I concluded when looking at what sensors it supported.

gtxaspec:
it appears that the sinfo.ko supplied with openipc supports:
```
sp1409
ov2735b
ov9712
cgu_cim
ov9732
ov9750
jxh42
sc1035
sc1135
sc1045
sc1145
ar0130
jxh61
gc1024
gc1064
gc2023
bf3115
imx225
ov2710
imx323
sc2135
jxh62
bg0806
ov4689
jxf22
imx322
imx291
ov2735
sc3035
ar0237
sc2145
jxh65
sc2300
jxv01
ps5230
ps5250
ov2718
ov2732
sc2235
jxk02
ov7740
hm2140
gc2033
jxf28
os02b10
os05a10
sc2232
sc4236
sc1245
gc1034

but doesn't support sc2336
```

the openingenic sinfo.ko supports all of the above, AND sc2336.   @ZigFisher maybe we can have more people test before inclusion, it seems the openingenic version is more versatile in it's sensor support.

gtxaspec:
@bengris32 maybe you could make a seperate PR for the sinfo module, that way we can get the sensor module stuff into the firmware sooner

ZigFisher:
The decision rests with you, colleagues.
As far as I have seen our variant supports more sensors and maybe it should be used everywhere.
Now I will write to try to make more tests.

bengris32:
> @bengris32 maybe you could make a seperate PR for the sinfo module, that way we can get the sensor module stuff into the firmware sooner
> 
> I can confirm the sensor module, bin, and changes to load_ingenic work well, tested on T20 wyze v2.

We could certainty do it that way, with the only obvious side effect being that people with jxf23 sensor would have to set it manually instead of the sensor being auto detected.

If you're fine with that, then I'm fine with it too.

FlyRouter:
There may be some misunderstanding or language barrier, I'll try to break it down and give explanations

- no problem adding your edits to the load_ingenic and ingenic-osdrv-t20.mk files, thank you.

- no problem adding sensor_jxf23.ko and jxf23.bin files if you can say with certainty that we can't compile them ourselves, thanks

- there is a problem with adding the sinfo.ko file because it is desirable to build it from the openingenic repo sources so that it contains all sensors at once.

So if you remove the last file or give arguments to apply the merge can be immediately. Thank you.

FlyRouter (approved):

FlyRouter:
Thank you

PR #972, opened 2023-08-25
add support for WiFi on LifeSmart LS179

BenoitDuffez:
this device has a HI3518EV200 SoC
the RTL8188 is connected over USB
the GPIO 54 is used to power the module


FlyRouter (approved):
Thank you

PR #974, opened 2023-08-27
Enhanced system timezone configuration via U-Boot.

gtxaspec:
Introduced capability to derive system timezone from U-Boot's 'timezone' environment variable, ensuring persistence even if the overlay is reset.
Added a new utility, timezone.sh, located in /usr/sbin.
Updated /etc/rcS to invoke timezone.sh, guaranteeing the updated timezone settings are accessible to all programs and scripts during boot.

This change was implemented to provide users with a more resilient and flexible approach to timezone management, especially in scenarios where system overlays might be frequently modified or reset, or for deployment configuration to multiple devices.

### DIFF
--- a/general/overlay/etc/init.d/rcS
+++ b/general/overlay/etc/init.d/rcS
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 # Setting the TimeZone for all processes
+/usr/sbin/timezone.sh
 export TZ=$(cat /etc/TZ)
 
 # Set the firmware creation time as the base system time
--- a/general/overlay/usr/sbin/timezone.sh
+++ b/general/overlay/usr/sbin/timezone.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 
-#Set the system timezone based on the u-boot environment variable
+# Set the system timezone based on the u-boot environment variable
 
-#Convert the timezone from lowercase, replace underscores with spaces for proper searching
+# Convert the timezone from lowercase, replace underscores with spaces for proper searching
 convert_timezone() {
     local converted=""
     IFS="/"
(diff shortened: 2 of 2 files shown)


themactep:
Let's clarify first, what is the primat, the config in /etc/ or the setting is the env? Should the /etc/ config file get created with every reboot overriding the one existing, or do we first check and compare the values? What's the order of updating the value when we make changes from Web UI, as they need to be synchronized yet remain optional for, say, FPV systems which does not have Web UI.

Also, why the case conversion? There is a standard of the timezone format, so it should be either matched or missed. The only thing I could agree on is replacing spaces with underscores but even that is overkill. That would make parsing matching the value from the list way simpler.

gtxaspec:
Thanks for the feedback @themactep!

1. As observed in `time-config.cgi`, the TZ is prioritized before timezone (https://github.com/OpenIPC/microbe-web/blob/master/files/var/www/cgi-bin/time-config.cgi#L21C75-L21C75). To ensure consistency, the `timezone.sh` script should be adjusted to always update both values in this order.

2. To optimize the write operations, we should implement a differential check mechanism. The values should only be written if there's a change detected, thus avoiding redundant writes.  Great Suggestion! =D

3. The case conversion mechanism exists to accommodate the current u-boot environment's predilection for lowercase. While u-boot predominantly employs underscores in lieu of spaces (except in command arguments), our script robustly handles this format. Users have the flexibility to employ the lowercase+underscore convention, but our script ensures compatibility irrespective of their choice. Extensive testing has been done on the entire timezone file to ensure accurate parsing, even with varied use of spaces, underscores, mixed case, and other symbols like '+' or '-'.

themactep:
>     1. As observed in `time-config.cgi`, the TZ is prioritized before timezone (https://github.com/OpenIPC/microbe-web/blob/master/files/var/www/cgi-bin/time-config.cgi#L21C75-L21C75). To ensure consistency, the `timezone.sh` script should be adjusted to always update both values in this order.

The part of code you refer to is only valid on updating the timezone. There's a validation of values above that, and if all components validate then saving of both parties ensues. The actual checking and assigning of default values happens here https://github.com/OpenIPC/microbe-web/blob/57f945c7a0f6998dae5709e0099ba8eea6321de6/files/var/www/cgi-bin/p/common.cgi#L729-L735

>     3. The case conversion mechanism exists to accommodate the current u-boot environment's predilection for lowercase. While u-boot predominantly employs underscores in lieu of spaces (except in command arguments), our script robustly handles this format. Users have the flexibility to employ the lowercase+underscore convention, but our script ensures compatibility irrespective of their choice. Extensive testing has been done on the entire timezone file to ensure accurate parsing, even with varied use of spaces, underscores, mixed case, and other symbols like '+' or '-'.

From what I see, uboot saves capitalized values just fine.
```
fw_setenv timezone America/Toronto
# fw_printenv -n timezone
America/Toronto
```
I think we can skip case conversion shenanigans and demand a true final value in env.

What we really care about for setting up a correct timezone is TZ, a string of values defining time change settings for the particular zone.

Timezone name (America/Toronto) is a mere convinience for the user. We could not care less if it is omitted or just lost. We could just take the first matching value and it still would work :)

themactep:
I think we can merge it as it is now. If there is something to be corrected, we can always do it later.

FlyRouter:
Okay, thank you all, I'll do it now.
I paused it to hear more opinions and confirmation of readiness since there was a discussion above.

PR #976, opened 2023-08-27
add t20_ultimate_defconfig

gtxaspec:
add t20_ultimate_defconfig to support t20 devices, including wyze cam v2 and wyze pan cam v1


FlyRouter (approved):
Thank you !

PR #986, opened 2023-09-07
[no ci] Overlay: move manual wlan mac setting to init script

viktorxda:


gtxaspec:
tested, working well.

PR #990, opened 2023-09-10
Package: logcat-openipc

gtxaspec:
logcat-openipc: added userspace programs (logcat, logwrapper) for reading the Android logging facility used by Ingenic IMP libraries


FlyRouter:
Thanks for sending PR but I have a question - why don't you make it so that the source code of the program is taken from your repository, as it is done in other packages. It would be more correct and efficient. We try to add program code directly into Firmware repo only if it contains not more than a dozen files. The link can go either to your personal repository on GitHub or you can make a transfer to the OpenIPC organization of any of your repos while retaining the rights of the main developer. Please think about it. So to add a new package you only need two new files, the first in your PS and minimal configuration edits. Thanks.

gtxaspec:
@FlyRouter yes, I agree.  Stand by for PR edit.

gtxaspec:
@FlyRouter changes completed

FlyRouter (approved):
Thank you

PR #998, opened 2023-09-20
overlay-files: Refactor wlan0 ifup Networking Configuration

gtxaspec:
- Incorporated a 15-second timeout within udhcpc_opts, ensuring adequate time for the device to aquire a dhcp lease from the network before forking to the background.
- Transitioned all wpa_supplicant parameters to the `pre-up` directive, ensuring wpa_supplicant initialization precedes DHCP client execution.

Collectively, these modifications optimize the DHCP lease acquisition process, enhancing the device's network initialization speed.

PR #1013, opened 2023-09-30
[no ci] Overlay: correct wireless interface files

viktorxda:
Commit #998 introduces some problems:

1. wlan0 should not be enabled by default, it is not needed for ethernet only devices and requires a driver in almost all cases (that is what the wlandev interface is intended for).

2. ~post-up is the preferred choice for wpa_supplicant, so we can be sure the [interface is initialized.](https://github.com/mirror/busybox/blob/master/networking/ifupdown.c#L1216)~

3. `-t 15` increases the discover counter to 15, removing it sets it to the [default value of 3](https://github.com/mirror/busybox/blob/master/networking/udhcp/dhcpc.c#L1175).


FlyRouter:
Continuation of a discussion that was previously started here
https://github.com/OpenIPC/firmware/pull/998

themactep:
i don't have a ready solution yet, but the way wifi acts now is very annoying. so i am all for changes as long as they make the interface to appear sooner and connect faster :)

FlyRouter (approved):
I agree with points 1 and 3.
If it is verified, I agree with point 2 as well

Also we need to disable or remove the eth0 interface if there is no cable connection on the device, so that it does not hang unconnected and does not get fallback ip and does not conflict with WiFi. Point 4.

ZigFisher:
Thank you.
I think it will take us some time to realize the changes and perhaps some colleagues will want to comment on this issue

gtxaspec:
In commit #998, I made WiFi connections work better and faster. Here were the main changes:

wpa_supplicant Timing: Most Linux systems start wpa_supplicant before trying to connect. Starting it after, like in openipc, slows things down. In #998, I fixed this.  I am not aware of another linux distribution or embedded linux system that places it in post-up.  wpa_supplicant can absolutely be started before the wireless interface is fully "up" or activated:

Doing so can indeed speed up the authentication process, especially in environments where faster network connection establishment is desired.  By starting wpa_supplicant early, it can begin its tasks (like scanning for networks, setting up credentials, etc.) while other system or network-related processes are also initializing.

Once the interface is up, if wpa_supplicant has already been running, it's in a ready state to authenticate immediately, rather than having to start up and then authenticate.

In many systems, especially embedded systems or IoT devices as we handle, reducing the time it takes to establish a network connection after boot or wake can be critical. Starting wpa_supplicant early contributes to minimizing this delay.

Timeout Change: We set the timeout to 15 seconds.  It gives udhcpc enough time to work properly, making sure devices connect without issues.  The dhcp lease is aquired before udhcpc moves into the background, saving time.

Testing Results: Before my changes, connecting to WiFi on ALL my openipc devices took around 30 seconds, longer than any IoT device I have operated.  With #998, it takes about 5 seconds. I checked and tested this on different WiFi modules, including RTL8189FS, RTL8189ES, and ATM603x, all using the Ingenic T20/31 platforms.

@victorxda, did my changes cause real issues for you on your devices, or are they just observations of the scripts? I'd hate to revert such a time saving effort.

viktorxda:
I made some boottime tests with the previous and current firmware to check for the actual time difference.

- Old firmware with post-up:
`--- boot time: 11 seconds`

- Old firmware with pre-up:
`--- boot time: 9 seconds`

- Current firmware:
`--- boot time: 22 seconds`

The pre-up vs. post-up makes indeed a difference, so we are keeping this change.

The main culprit seems to be the discover packets, which sometimes stacks up to the complete 15 queries:
```
udhcpc: broadcasting discover
udhcpc: broadcasting discover
[...]
udhcpc: broadcasting discover
udhcpc: broadcasting discover
```

> Timeout Change: We set the timeout to 15 seconds. It gives udhcpc enough time to work properly, making sure devices connect without issues. The dhcp lease is aquired before udhcpc moves into the background, saving time.

I don't see a timeout change, but there two options that come into question:
`Pause between packets -T` and `Send up to N discover packets -t` which is mentioned on point 3.
Both should be reduced to the minimum value to reduce the setup time.

gtxaspec:
` -x hostname:$(hostname) -T 1 -A 0 -R -b -O search` solves the delay (20 seconds default) after udhcpc forks to the background.

ZigFisher (approved):

PR #1017, opened 2023-10-01
overlay-files: Fix rc.local output display during boot

gtxaspec:
Small change, changed `printf` to `echo` in S99rc.local to ensure correct output on the system console during boot.


ZigFisher (approved):

gtxaspec:
updated with stop case statement

PR #1022, opened 2023-10-05
Workflow: add ssc338q initramfs

viktorxda:


FlyRouter (approved):
I have no objection, now all that matters is that it works.

PR #1023, opened 2023-10-05
Files to start venc

MarioFPVdev:
Added /etc/init.d/S98venc
Added /root/channels.sh for mavfwd
Fixed telemetry


FlyRouter (approved):
Thank you

PR #1024, opened 2023-10-05
Update venc-openipc.mk

MarioFPVdev:
created commands to include the new files for venc


FlyRouter (approved):
Thank you

ZigFisher (approved):
Yes, right ;)

PR #1027, opened 2023-10-09
[no ci] Overlay: update sysupgrade nand detection

viktorxda:
1. Currently flash_type is only necessary to determine the flash_eraseall -j option.
2. As far I am aware, only the Sigmastar NAND firmware can be updated with the sysupgrader (by utilizing gluebi). This requires the nor firmware package as well, so it should be possible to set it to nor for all devices by default.


ZigFisher (approved):
Thank you

PR #1032, opened 2023-10-10
Update datalink.mk

MarioFPVdev:
channels.sh should be in /root

### DIFF
--- a/general/package/datalink/datalink.mk
+++ b/general/package/datalink/datalink.mk
@@ -26,7 +26,7 @@ define DATALINK_INSTALL_TARGET_CMDS
 	cp $(DATALINK_PKGDIR)/files/$(DATALINK_FPATH)/telemetry $(TARGET_DIR)/usr/bin
 
 	$(INSTALL) -m 755 -d $(TARGET_DIR)/usr/sbin
-	cp $(DATALINK_PKGDIR)/files/$(DATALINK_FPATH)/channels.sh $(TARGET_DIR)/usr/sbin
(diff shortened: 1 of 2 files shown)


ZigFisher (commented):
1. The correct path to the scripts is /bin, /sbin, /usr/bin, /usr/sbin

2. If the file in /tmp is of significant size, it will eat up your memory. 
Consider using the logging command "logger" with the parameters

Example wrire:

```
root@openipc-gk7205v300:~# echo Test | logger -t fpv
```

Example read:

```
root@openipc-gk7205v300:~# logread | grep fpv

Oct 10 19:36:32 openipc-gk7205v300 user.notice fpv: Test
```

general/package/datalink/datalink.mk
```diff
@@ -26,7 +26,7 @@ define DATALINK_INSTALL_TARGET_CMDS
 	cp $(DATALINK_PKGDIR)/files/$(DATALINK_FPATH)/telemetry $(TARGET_DIR)/usr/bin
 
 	$(INSTALL) -m 755 -d $(TARGET_DIR)/usr/sbin
-	cp $(DATALINK_PKGDIR)/files/$(DATALINK_FPATH)/channels.sh $(TARGET_DIR)/usr/sbin
+	cp $(DATALINK_PKGDIR)/files/$(DATALINK_FPATH)/channels.sh $(TARGET_DIR)/root
```
cronyx, line 29: In [mavfwd code](https://github.com/OpenIPC/mavfwd/commit/feac9dad090bca82e3b63de14174ca9aada0f8b1), channels.sh run without custom path. It can be located only in /bin /sbin /usr/bin or /usr/sbin

PR #1042, opened 2023-10-12
ingenic-osdrv t20/t31/t40: add PWM core and hal driver modules

gtxaspec:
added sample_pwm_core.ko and sample_pwm.hal.ko driver modules for t20/t31/t40 to support PWM hardware (ex. pwm driven IR led) in those devices.


ZigFisher (commented):
How about making them buildable from source code in the https://github.com/OpenIPC/openingenic ?

gtxaspec:
> How about also making them buildable from source code in the https://github.com/OpenIPC/openingenic ?

That is the plan 😁 I will post a PR there later today, @ZigFisher.  Thank you!

PR #1047, opened 2023-10-14
add sc4236 param file to makefile script

gtxaspec:
add sc4236-t31.bin param file to ingenic-osdrv-t31 package makefile so it copies it to the device.


FlyRouter (approved):
Thank you

PR #1049, opened 2023-10-15
package: add altobeam atbm6441 wireless driver

gtxaspec:
add altobeam atbm6441 wireless driver commonly used with ingenic t31 zeratul zrt platform devices


ZigFisher (approved):
Thank you

PR #1050, opened 2023-10-15
ingenic-osdrv: delete pwm driver files

gtxaspec:
since we now build these from source, for all ingenic t-series cpu ( t20/ t31 / t40 ), we no longer need the pre-compiled files.

added to openingenic repo https://github.com/OpenIPC/openingenic/pull/19/files


FlyRouter (approved):
Thank you

PR #1051, opened 2023-10-15
overlay-files: update wireless sdio for wyze atbm603x

gtxaspec:
update wireless sdio for wyze atbm603x:

for atbm603x on ingenic t31, gpio must be cleared and then set for atbm to properly init.  

update sdio script, tested on wyze pan v2, and atomcam 2.


FlyRouter (approved):
Thank you

PR #1053, opened 2023-10-16
ingenic: kernel config - enable CONFIG_DYNAMIC_DEBUG

gtxaspec:
Enable CONFIG_DYNAMIC_DEBUG for Ingenic kernel configs

enable logs with:
`echo 'file * +p' > /sys/kernel/debug/dynamic_debug/control`

disable logs with:
`echo 'file * -p' > /sys/kernel/debug/dynamic_debug/control`


themactep:
> Enable CONFIG_DYNAMIC_DEBUG for Ingenic kernel configs
> 
> enable logs with: `echo 'file * +p' > /sys/kernel/debug/dynamic_debug/control`
> 
> disable logs with: `echo 'file * -p' > /sys/kernel/debug/dynamic_debug/control`

What about making these into a script or a shell alias? Say, debug_on and debug_off.

gtxaspec:
@themactep I like that idea.  I'll submit a different PR updating overlay-files with a script for consideration.

PR #1066, opened 2023-10-19
Disable watchdog

MarioFPVdev:
This fixes the 30 minutes reboot.
https://github.com/OpenIPC/sandbox-fpv/blob/master/note_nvr_wdt.md


FlyRouter (approved):
Thank you

FlyRouter:
@cronyx 
Please check Mario's suggestions
I think you need to tweak the path and put the module in the right place
You also need to correct the path when calling from the script.
Thank you

PR #1068, opened 2023-10-19
Default port for WFB-ng

MarioFPVdev:
The default port for WFB-ng is 5600.
In /etc/init.d/S98vdec it is already set correctly.


FlyRouter (approved):

FlyRouter:
@cronyx 
Please check Mario's suggestions
Thank you

PR #1069, opened 2023-10-19
ingenic: remove logcat from lite defconfig

gtxaspec:
ingenic: remove  logcat from lite defconfig for space saving.   available to users only in ultimate releases for now.


FlyRouter (approved):
Thank you

PR #1070, opened 2023-10-19
simplified air and ground scripts

cronyx:


FlyRouter (approved):
Thank you very much for the edits

Please note that I have already put the wdt.ko kernel module in the correct path

And what we need to clean up here is
https://github.com/OpenIPC/firmware/blob/797f21fba70f87ed333e9e64c31b7da3b5db0e7a/general/package/vdec-openipc/files/S98vdec#L8

This PR from Mario created a little bit of chaos, but your version will of course be more correct, thank you.

FlyRouter:
I cleaned out the file
https://github.com/OpenIPC/firmware/pull/1071/files

I intervened on a tablet not exactly pretty as everything stopped and had to make quick edits due to bad actions due to PR.
Everything should be OK now.

PR #1074, opened 2023-10-20
[no ci] add switch_debug function to profile

gtxaspec:
**Integrate the `switch_debug` command function into the Default User Profile**

Building upon the advancements in [PR #1053](https://github.com/OpenIPC/firmware/pull/1053), we have facilitated the `CONFIG_DYNAMIC_DEBUG` feature within the Linux kernel. This is tailored to suppress most engineering debug logs by default, as to not display them to everyday users.

Users now have the flexibility to toggle these debug messages through simple commands:
- To enable: `switch_debug on`
- To disable: `switch_debug off` or just `switch_debug`

While the majority of debugging filters are active, we've deliberately excluded  the following modules: `mmc_core`, `avpu`, `squashfs`, `mtdblock`, and `fs/sysfs/file.c`. These modules are known to generate an overwhelming number (thousands) of debug lines. However, users who wish to tap into these specific debug messages can manually enable them via `/sys/kernel/debug/dynamic_debug/control`.

### DIFF
--- a/general/overlay/etc/profile
+++ b/general/overlay/etc/profile
@@ -49,6 +49,17 @@ show_wlan() {
 	grep -r '$1..=' /etc/wireless | cut -d '"' -f 4 | sort | grep -e $(fw_printenv -n soc) -e generic
 }
 
+switch_debug() {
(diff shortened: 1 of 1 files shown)


FlyRouter (approved):
Thank you

PR #1084, opened 2023-10-23
[no ci] overlay-files: wireless - wyze pan changes

gtxaspec:
Add support for Wyze Pan V2 and correct Pan V1 driver name

- Introduced support for Wyze Pan V2 due to distinct GPIOs for the wireless device.
- Fixed the driver name for Pan V1.


FlyRouter (approved):
Thank you

PR #1100, opened 2023-11-01
[no-ci] ingenic-osdrv: don't claim gpio 0

gtxaspec:
Don't claim any GPIO `0` (or a negative value), retrieved from the `majestic.yaml` configuration for any GPIO setting.

`GPIOS=".nightMode.irCutPin1 .nightMode.irCutPin2 .nightMode.irSensorPin .nightMode.backlightPin .audio.speakerPin"`


gtxaspec:
Let's Revisit at a later time

PR #1102, opened 2023-11-01
[no-ci] overlay-files: update rc.local init script

gtxaspec:
Fixes:

- Change the name of S99rc.local to S94rc.local. By doing this, any tasks within this file will run before tasks in S95majestic. This fixes issues like recording to network storage failing because network share is not mounted before S95majestic starts.

Enhancement:

- Introduce an optional file named rc.local.stop. If you have tasks you need to run when the system is rebooting (like disconnecting from a network storage), you can add them to this file.

### DIFF
--- a/general/overlay/etc/init.d/S94rc.local
+++ b/general/overlay/etc/init.d/S94rc.local
@@ -3,7 +3,6 @@
 # Start rc.local
 #
 
-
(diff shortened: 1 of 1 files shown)


FlyRouter (approved):
Thank you

PR #1122, opened 2023-11-07
[no ci] fix timezone.sh to write complete timezone data string

gtxaspec:
- Fix: timezone.sh should write the entire POSIX timezone string instead of just the timezone's short code.  This will make certain that the time is set correctly on NTP sync, ensuring that the system knows both the standard offset and when to apply daylight saving adjustments.

before:
```
# cat /etc/TZ
PST8PDT
```

after:
```
# cat /etc/TZ
PST8PDT,M3.2.0,M11.1.0
```


FlyRouter (approved):
Thank you

PR #1123, opened 2023-11-08
fix rtl8812au driver build after the last code update

cronyx:


FlyRouter (approved):
Thank you

bwees:
Is there a reason the RTL8821 is no longer supported after this commit?

PR #1136, opened 2023-11-12
ingenic-osdrv-t31: fix driver module path

gtxaspec:
Fix:  T31 platform, no longer uses pre-compiled binaries for kernel modules. Vendor kernel modules are currently placed in `/lib/modules/$KERVEL_VERSION/extra`, they should be in `/lib/modules/$KERVEL_VERSION/ingenic`, which was currently empty.  

Correct the makefiles to place the kernel modules in the appropriate directory.


FlyRouter (approved):
Thank you

PR #1153, opened 2023-11-21
ingenic-motors: correct module install dir

gtxaspec:
Fix: 
- sample-motor.ko should be installed in the ingenic modules directory, lets adjust the make file to correct this.
- correct module name in makefile header


FlyRouter (approved):
Thank you

PR #1180, opened 2023-12-03
Added DVR recording

MarioFPVdev:
Value for filename is stored in /etc/rec.
If the /etc/rec doesn't exist then we create the file.


cronyx:
Hi, you can't enable disk writing by default for everyone. What happens if the disk is not connected?

cronyx:
#1190 Added automatic names generation without any hardcode in the startup scripts.

Setup in the /etc/vdec.conf full path to records archive. For example: 

`records=/mnt/sda1/records`

the recording files will look like this: video_XXXX.h265

PR #1188, opened 2023-12-06
[no ci] overlay-files: update resetd log messages

gtxaspec:
currently, resetd logs this to the console: `GPIO not set. Exiting`

Enhance the messaging to be more informative as to where the message originates from.


FlyRouter (approved):
Thank you

PR #1191, opened 2023-12-10
[no ci] overlay-files: add watchdog option to init.d

gtxaspec:
Implement optional system watchdog in init.d scripts:

Introduced an optional hardware-based system watchdog, functioning independently of the majestic program. This feature provides an alternative layer of reliability, allowing users to utilize the system's hardware watchdog independently from majestic's internal operations. 

This enhancement ensures that the system watchdog remains operational even in scenarios where majestic may not be functioning correctly, for instance, hanging on reboot.

Users can activate this feature by modifying the init.d script. To prevent conflicts, this system watchdog is designed not to run if majestic's own internal watchdog is active, even if the system watchdog is enabled.  It is disabled, by default.

### DIFF
--- a/general/overlay/etc/init.d/S99watchdog
+++ b/general/overlay/etc/init.d/S99watchdog
@@ -1,0 +1,42 @@
+#!/bin/sh
+
+DAEMON="watchdog"
+
+WATCHDOG_ENABLED=false
+WATCHDOG_TIMEOUT=60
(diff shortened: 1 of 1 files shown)


ZigFisher (approved):
We need extensive testing.
We'll discuss the details in Telegram chat.

FlyRouter (approved):

PR #1198, opened 2023-12-17
ingenic-osdrv-t31: remove binary kernel modules

gtxaspec:
binary sensor modules for t31 have been moved to the sensors repo for archival since we build using openingenic now.

https://github.com/OpenIPC/sensors/pull/4


FlyRouter (approved):
Thank you

PR #1200, opened 2023-12-17
[no ci] update load_ingenic script to add audio module parameters

gtxaspec:
by default, audio.ko will claim GPIO63, regardless if audio is enabled or not within majestic or anywhere else.

```
# cat /sys/kernel/debug/gpio 

GPIOs 32-63, GPIO B:
 gpio-63  (gpio_spk_en         ) out lo
```
Not all devices use GPIO63 for audio out, and this interferes with devices that could use GPIO63 for other purposes, for example for motors, leading to overheating motors.

specify `spk_gpio=-1` in `load_ingenic` so the audio.ko module does not claim any GPIO by default ( let majestic do this if the user specifies a speaker out gpio pin )  Now, other devices may use the GPIO63 pin if they need to:

```
# cat /sys/kernel/debug/gpio 

GPIOs 32-63, GPIO B:
 gpio-63  (motor_st4_gpio      ) out lo
```


FlyRouter (approved):
Thank you

PR #1201, opened 2023-12-17
enable libimp-control across remaining ingenic platforms

gtxaspec:
currently we enable libimp on t10, t31, enable libimp-control across the remaining ingenic platforms


FlyRouter (approved):
Thank you

PR #1204, opened 2023-12-19
[no ci] add go2rtc package

gtxaspec:
add go2rtc streamer package


FlyRouter (approved):
Thank you.
It seems to me that the lines indicating video streams need to be shifted to the right by one character in accordance with the yaml standard.

FlyRouter (commented):
Also, now this application will be compiled only for the MIPS platform.
It is advisable to add the ability to detect and select, or for now indicate in the selector that it will only work with Ingenic
I built this application and launched it, it works well, but it requires quite a lot of resources. Just a little less than ffmpeg.

gtxaspec:
I'll revise the makefile to allow compilation on whatever the target is, stand by. =D

FlyRouter (approved):
Thank you

FlyRouter (approved):

PR #1207, opened 2023-12-21
ingenic-osdrv-t40: remove binary kernel modules

gtxaspec:
binary sensor modules for t40 have been moved to the sensors repo for archival since we build using openingenic now.

https://github.com/OpenIPC/sensors/pull/5


FlyRouter (approved):
Thank you

PR #1223, opened 2024-01-02
Ingenic t20: use openingenic sdk

gtxaspec:
openingenic sdk has been updated with T20 sources, switch to that instead of pre-compiled binaries.


FlyRouter (approved):
Thank you

PR #1227, opened 2024-01-03
[no ci] update module parameters for atbm

gtxaspec:
add `atbm_printk_mask=0` to the generic atbm profile to prevent spamming of the dmesg and system logs


FlyRouter (approved):
Thank you

PR #1264, opened 2024-01-18
[no ci] Ingenic: update t20/t30 osdrv libraries

viktorxda:


viktorxda:
The libraries have been properly set now, please check it again for the T20 and T30. Thanks.

gtxaspec:
T20 confirmed working now ( though still with corrupted video on T20 since `master+afd484b` , could be another other issue though )

PR #1281, opened 2024-01-26
add ov9750 for T20/T10

johndi3:


FlyRouter (commented):
@johndi3 
Thank you

@gtxaspec 
Can we build a driver for this sensor from source?

PR #1283, opened 2024-01-27
enable eth0 by default with highest priority, monitor ethenet state changes

johndi3:
Tested several devices with and without Ethernet, need more tests

### DIFF
--- a/general/overlay/etc/init.d/S40network
+++ b/general/overlay/etc/init.d/S40network
@@ -11,7 +11,6 @@ set_wireless() {
 	elif $path/modem "$dev"; then
 		ifup usb0
 	fi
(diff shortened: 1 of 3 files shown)


FlyRouter (commented):
Thank you very much for your work
I think our colleagues 
@widgetii @cronyx @p0i5k @viktorxda @zigfisher @themactep @ialexlog @gtxaspec @dimerr 
will consider and check your proposal and give feedback in the near future.

FlyRouter:
Since this PR is yours, interesting and at the same time critical, preliminary experiments have been submitted to the Sandbox.
https://github.com/OpenIPC/sandbox/tree/main/scripts/network

After passing all tests, edits and collecting feedback, a new PR will be created in the Firmware repository.

PR #1285, opened 2024-01-29
Package: filter plugins by vendor

viktorxda:


FlyRouter (approved):
Thanks, good solution.

PR #1287, opened 2024-01-30
Update Makefile

longcat99:
For 16mb's NOR flash, more kernel space should be added!


FlyRouter (commented):
Contradicts OpenIPC unification

FlyRouter:
Does not meet accepted kernel size standards in all OpenIPC firmwares.

PR #1300, opened 2024-02-05
add jxq03 for sigmastar i6b0

johndi3:


FlyRouter (commented):
Thank you

@viktorxda
Please look at the status of the sensor, maybe we can build a driver from the source?

PR #1321, opened 2024-02-17
Update infinity6e-ssc012b.config

m0nk1111:
usb lte modem modules added


FlyRouter (approved):
Thank you for your work, but it seems to me that some of the modules can be disabled as they will take up extra space.

FlyRouter (changes requested):
https://github.com/OpenIPC/firmware/blob/master/general/package/linux-patcher/linux-patcher.mk

Please pay attention to this configuration file, in which you can enable some functions only for FPV profiles.
We kindly ask you to sort out the necessary blocks and send the patch here to enable RNDIS modems.
Thank you very much.

PR #1331, opened 2024-02-29
RubyFPV Integration - Part 1

PetruSoroaga:
Added the basic building blocks:
1. A new empty package, just with basic setup, no real code, specific for FPV stuff;
2. Default new configs for Goke and Sigmaster chipsets cameras, configs that are just clones of the existing fpv flavor with few packages removed from build and this new package added.


FlyRouter (approved):
Thank you

PR #1332, opened 2024-03-01
RubyFPV integration, part 2

PetruSoroaga:
Added placeholder binaries for all binaries that will part of RubyFPV package;
Added RubyFPV package to the list of known packages.


PetruSoroaga:
I kept the PR very small in scope, to be easy to follow what's going on.

PR #1336, opened 2024-03-05
Ruby FPV Integration, part 3

PetruSoroaga:
* Changed the linux kerner timers resolution to 1000 hz, for RubyFPV builds;
* Reogranised the RubyFPV package content


PetruSoroaga:
> Adding several similar kernel configurations will clutter the folders over time, please add your kernel changes here: https://github.com/OpenIPC/firmware/blob/master/general/package/linux-patcher/linux-patcher.mk

Done. I removed the kernel config files I just added before.

PetruSoroaga:
Done. I removed the kernel config files I just added before.

PetruSoroaga:
> Please note that for Sigmastar we use the majestic lite build (fpv is basically lite with audio support removed due to size restrictions). For Sigmastar we have 16MB+ flash by default, so we can use majestic lite to support audio as well.

Done, updated to use lite majestic variant for sigmaster rubyfpv variant builts

PetruSoroaga:
Done, integrated the feedback about executable files access rights flags.

PR #1388, opened 2024-04-08
vtund

augin:
Adding support for storing tunnel settings in the environment


FlyRouter (commented):
I have a suggestion to change the variable name from "vtund" to "vtun" which is more correct.

augin:
I think it will be right

skilur:
PR for the name change:
https://github.com/OpenIPC/firmware/pull/1411

FlyRouter:
Thank you !

PR #1421, opened 2024-04-26
Safely exiting in case of error

reloadman:
In case of incomplete download of the firmware file, erasing the flash may start before the reboot

telegram-cloud-photo-size-2-5280575734412337330-y


FlyRouter:
@cronyx 
Hello
Look what an interesting situation, I think we need to consider it the same way.
Some process prevented the device from rebooting and the process of executing script commands continued.
I believe that the fuse proposed by the guys from Faceter should be used, there are no questions about it.
However, maybe we should make the “reboot” command more strict, for example with the -f switch?

FlyRouter (approved):
Thank you

PR #1424, opened 2024-04-29
gst1-plugins-bad-openipc: Shadow gst1-plugins-bad

ajlennon:
...to support disabling nvcodec

This plugin can't be disabled via OpenIPC with the upstream buildroot package. When we pull in the package building for the Ingenic T31 we get an error building nvcodec. So shadow and add a new set of buildroot configuration options to support disabling build of nvcodec


FlyRouter (approved):
Thank you

ajlennon:
Thanks @FlyRouter  - I am a bit worried this is causing the CI to fail though?

PR #1429, opened 2024-05-01
WIP: Support Ingenic T31 GLIBC toolchain

ajlennon:


FlyRouter (commented):
Perhaps we should think of some more beautiful structure for this particular case. 
We can consider any proposals.

ajlennon:
Hi @FlyRouter @viktorxda could you please kick off a toolchain workflow to test creation of the T31 GLIB toolchain? Thanks!

FlyRouter (approved):
Thank you.
I don’t quite agree with the name of the profile and the structure at this moment, but I hope we can work on it some more.

ajlennon:
Thanks. One thought I had was to break out the naming to include the code library such as $platform_$library_$distro_defconfig 🤷‍♀️

PR #1438, opened 2024-05-10
Add rtp patch for baresip

skilurius:


FlyRouter (approved):
Thank you monsieur
I checked, it works great !

PR #1441, opened 2024-05-12
Package: update osd server makefile

viktorxda:


FlyRouter (approved):
Thank you.
In the future, I think we will be able to add a collection for Ingenic ;)

PR #1472, opened 2024-06-17
Update majestic.mk

skilurius:


flyrouter (commented):
Sure

flyrouter:
https://github.com/OpenIPC/wiki/blob/master/en/majestic-config.md

PR #1482, opened 2024-06-28
Update ssc335_fpv_defconfig

zhngchng:


flyrouter (commented):
Changes rejected
1. This is not an FPV configuration
2. All descriptions must be in English only

PR #1490, opened 2024-07-14
Update gc4023.yaml

NVOAN:
Because the sensor source code is 0x29,
Here 0x21 will not generate an image, just change it to 0x29. I have verified it on the hardware.
https://github.com/OpenIPC/openingenic/blob/master/kernel/sensors/t31/gc4023/gc4023.c


flyrouter:
Information accepted, thank you very much

PR #1520, opened 2024-08-13
Add more sensor modules and settings for gm8136

csev1755:
Pulled from stock firmware of a Faleemi FSC880.


flyrouter (approved):
Thank you

PR #1534, opened 2024-08-22
fix watchdog when majestic crashes

zerog2k:
per wdt.ko:
```
parm=nodeamon:By default, a kernel deamon feed watchdog when idle, set 'nodeamon=1' to disable this. (default=0)
parmtype=nodeamon:int
```
as documented in https://github.com/OpenIPC/majestic/issues/171, when majestic crashes, without `nodeamon=1` param to wdt kmod, camera does not automatically reboot.

Expected behavior is for majestic to feed the watchdog, and if majestic crashes, watchdog timer would expire and reset cpu, forcing host reboot.

### DIFF
--- a/general/package/hisilicon-osdrv-hi3516cv100/files/script/load_hisilicon
+++ b/general/package/hisilicon-osdrv-hi3516cv100/files/script/load_hisilicon
@@ -345,7 +345,7 @@ insert_ko() {
 	insert_audio
 	echo "==== Your input Sensor type is $SENSOR ===="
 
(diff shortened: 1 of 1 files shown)


Flole998:
I've discovered this a few months ago, performed the change and for me it was an improvement as now the camera reboots when there's an issue.

flyrouter:
Good day
Thank you for waiting. 
I will try to check it on my device during the weekend.

Flole998:
@flyrouter Anything new? This has also been reported as #1803

PR #1543, opened 2024-09-04
Update ssc378de_lite_defconfig, experimental enable WireGuard

flyrouter:


NicksonYap:
I am interested to have Wireguard running in OpenIPC as well to better secure my camera's streaming to a remote server

PR #1565, opened 2024-10-07
[no ci] Package: remove log reader

viktorxda:


flyrouter (approved):
The package has been disabled because there is no longer a need to use it.

PR #1582, opened 2024-10-19
Install vtxmenu files

henkwiedig:
This PR will install new vtxmenu files as well as change msposd's default out port to send data through the mavlink telemetry.

This PR needs: https://github.com/OpenIPC/msposd/pull/8


flyrouter (commented):
I have no objections if it is tested and works.

henkwiedig:
I've changed the the out port to not send over the default mavlink stream.
Possible mavlink listeners will probably not understand msp.

PR #1605, opened 2024-11-12
use msposd HEAD again

henkwiedig:
This should fix: https://github.com/OpenIPC/msposd/issues/19
We will introduce a testbuild in msposd against https://github.com/OpenIPC/firmware to safeguard this will not happen again in the future.


flyrouter (approved):
thank you

PR #1609, opened 2024-11-17
Update fh8858v210.generic.config

caliph9k:
Enabled CONFIG_MACH_FH8858v210 which is needed for build to succeed


flyrouter (commented):
Thank you.
Then what is line 302 for?

caliph9k:
I know its a duplicate line, mere lines apart but it does make a difference (at least for me).
But your comment got me thinking more, and I considered it wasn't the additional line that made it work but the removal of '# CONFIG_MACH_FH8858V210 is not set'. As such I tried building with the current repo file and it failed in linux-custom as usual. I then simply removed that specific comment and it built successfully. Somehow this is impacting the content of the build/linux-custom/.config file. If the comment is there, the .config doesn't have the required CONFIG_MACH_FH8858V210=y. I will close this pull request as it is not accurate in light of the new information. I will double-check everything and open a new even smaller pull request if the fix is confirmed..

flyrouter:
Thank you for your reply and understanding sir.
Yes, that is exactly what I wanted to say. that the patch may need to be changed.
We will wait for your corrections, we are really glad that you will take part in solving the problem, thank you.

PR #1613, opened 2024-11-22
Update fh8858v210.generic.config

caliph9k:
made CONFIG_MACH comments consistent with other FH885xV200 variants which enables successful build of fh8858v210


caliph9k:
For some reason the presence of the not set comment '# CONFIG_MACH_FH8858V210 is not set' alters some logic, which results in the build/linux-custom/.config file not having the requisite CONFIG_MACH_FH8858V210=y line. Removing that comment results in the right line and thus the build completes successfully.

PR #1623, opened 2024-11-29
[no ci] Package: update busybox.config

viktorxda:


flyrouter:
As far as I understand, we no longer use the RNGD applet and RANDOM generation for the first entry occurs alternatively?

We used the HDPARAM applet to work together with NVRs to which disks were connected. Since there are few such installations, I think it is normal that we disabled it.

PR #1664, opened 2025-01-15
Add OS02G10 driver for Gk7205v200 (as seen on Tiandy boards)

dimerr:


flyrouter (approved):
As always, thanks bro ;)

PR #1665, opened 2025-01-16
Add GC4023 for Gk7205v200

dimerr:


flyrouter (approved):
As always, thanks bro ;)

PR #1668, opened 2025-01-18
Limit the number of channels supported by the VENC module.

dimerr:
Some of the OS memory can be saved


flyrouter (approved):
Thank you very much, my friend
We will do the testing soon!

PR #1693, opened 2025-01-29
Allow switching MSPOSD from air to ground and vice-versa via config

ramkrishna757575:
Added the ability to switch MSPOSD from rendering on the air or on the ground based on value configured in /etc/wfb.yaml

Also added the ability to configure MSPOSD AHI type via /etc/wfb.yaml config file


ramkrishna757575:
> Hello, we could merge this for wifibroadcast. For wifibroadcast-ng there are currently a lot of ongoing changes so it might be adopted at a later time.

Thanks @viktorxda for reviewing it. In that case I think its better to decline this PR for now as a similar patch has already been created here https://github.com/OpenIPC/firmware/compare/master...henkwiedig:firmware:msposd-gs-rendering

PR #1702, opened 2025-02-11
Update generate_vtx_info.sh

snokvist:
Removed yaml-cli updates until yaml-cli is updated.


flyrouter (commented):
@cronyx 
Please look at this PR, should we close it or implement it.
Thanks.

snokvist:
Leaving it open will cause vtx_info.yaml to grow 2kb each reboot due to an unforseen yaml-cli bug. Removing these lines will eliminate the problem for now.

PR #1723, opened 2025-02-23
[no ci] General: move additional packages to legacy

viktorxda:


flyrouter (approved):
Thank you very much for this solution

PR #1724, opened 2025-02-23
YAML-relase candidate

snokvist:
- Temperature throttle checker
- wifi profile integration
- vtx_info generation
- auto set radio parameters based on bitrate.
- .... and more


snokvist:
I forgot compile options for src files.

PR #1473, opened 2024-06-18
Update load_grainmedia to work on GM8135 SOC. 

willersonsp:
This pull request needs another repository pull request to be approved first.

https://github.com/OpenIPC/linux/pull/22

PR #1735, opened 2025-03-04
Fix audio on GM8136

csev1755:
I found my GM8136 audio broke if this line executed and noticed it was added for the GM8135 specifically in #1473


flyrouter (approved):
Thank you

PR #1739, opened 2025-03-05
Consolidate motor packages - create gpio-motors

csev1755:
This PR consolidates code from gk7205v200-motors and sigmastar-motors as their code was almost completely identical. Additionally, I've added support for the sysfs GPIO interface for kernels older than 4.8.0. 

Perhaps in the future we could either move this package to the motors repository and consolidate further or retire the motors repository and split those packages out into the firmware repository. Interested to hear opinions.


csev1755:
I've converted this to a draft as I realized I had not properly imported both of the gpiochip configurations from both packages. 

Also, while trying to install this package onto a different camera than I developed on, I ran into some serious performance issues that I'm working through at the moment.

flyrouter:
Hello
Thank you for your interest in our project.
Yes, the idea of ​​making a universal package looks interesting.
However, I have a suggestion for you.
Make and add the gpio-motors package first and include all components from goke, sigmastar and your grainmedia.
And after testing, the next step will be to remove the sigmastar-gpio-motors and goke-gpio-motors packages.
One-time addition and removal will look dubious without testing and is unlikely to be accepted.
Thank you for understanding.

csev1755:
Thanks for your input, I added it as a new package. I ended up changing quite a bit of the code and now I'm getting much better performance on my hi3516cv200 camera than I was before. 

I also ended up going with just sysfs for wider compatibility. The complexity of detecting gpiochips seemed unnecessary when during my testing I found no performance benefit to going with chardev. Maybe this could be added later. 

I also added the option to set the motor delay as an argument.

flyrouter:
Am I right in understanding that you have tested this code on your two devices hi3516cv200 and gm8135 and now we should test it on other types?
How do you feel about writing a short note in the Wiki on motors and attaching the pictures that you posted in the motors repo?
I see that you are a big ethusiast and are passionate and interested in this, and that is why I suggest you do these experiments.

csev1755:
I could absolutely update the wiki! I have tested those two devices but I do have some other devices I could try out. I thought I might as well get the code out there so people could try it out and give feedback.

flyrouter:
Your description here is quite clear and readable, I think this can be taken out into the package description.
https://github.com/OpenIPC/motors/pull/13/files

flyrouter (approved):
Thank you.
We will merge this stage of work and then try to add the transfer of GPIO configuration variables via ENV as you planned.
Also, please try to use full device names if known from the builder repository, example:
ssc337de-foscam => ssc337de_ultimate_foscam-x5
and so on

PR #1744, opened 2025-03-10
Update gpio-motors

csev1755:
This update allows configuration of GPIO pins via U-Boot environment variables along with extra info on confirmed working SOCs

### DIFF
--- a/general/package/gpio-motors/Config.in
+++ b/general/package/gpio-motors/Config.in
@@ -1,6 +1,6 @@
 config BR2_PACKAGE_GPIO_MOTORS
 	bool "gpio-motors"
 	help
-	  Control stepper motors via sysfs GPIO. This code was developed 
-	  on a GrainMedia GM8136 board with 24BYJ48 stepper motors and a ULN2803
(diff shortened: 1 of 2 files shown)


flyrouter (commented):
Good day
Thank you for your suggestions, they look quite interesting, we will consider them soon.

@viktorxda  @cronyx  please join to view this PR

My only personal suggestion is that perhaps we should name the variable as "gpio_motors" by analogy with future other "gpio_*" as planned.
But this is only a discussion proposal for now and other guys should also speak out on this matter.

Thank you.

flyrouter:
I also think that as a backup source of information we should write down the list of GPIOs from already tested devices in the description of the Config.in file

csev1755:
Would it be appropriate to put the list of configs in the wiki instead? I feel like a user would be more likely to look there.

flyrouter:
Good day
I like the final version, thank you
We are waiting for comments from @cronyx and @viktorxda

cronyx:
Hello, looks good.

PR #1746, opened 2025-03-10
Shutdown Ruby, if present, during sysupgrade procedure

PetruSoroaga:
Stop all Ruby processes, if present, when launching a sysupgrade procedure.


ystinia (approved):
Thank you

flyrouter (approved):
Thank you very much.
I think this will solve the problem.

Issue #58, opened 2021-10-03
Compile problems for hi3518ev300


001Tycoon:
Hello

export PLATFORM=hisilicon;make prepare;make BOARD=unknown_unknown_hi3518ev300_openipc clean distclean
export PLATFORM=hisilicon;make prepare;make BOARD=unknown_unknown_hi3518ev300_openipc all

```
Error: arch/arm/boot/dts/hi3518ev300-demb.dts:135.11-12 syntax error
FATAL ERROR: Unable to parse input tree
make[4]: *** [scripts/Makefile.lib:313: arch/arm/boot/dts/hi3518ev300-demb.dtb] Fehler 1
make[3]: *** [arch/arm/Makefile:348: dtbs] Fehler 2
```

ZigFisher:
Fixed, thanks

001Tycoon:
thanks

PR #1748, opened 2025-03-11
fix system time from sd card files

robin-debug:
Add little shell code to restore system time from sd card file after boot up.It's woking in my rc.local, but I'm not sure there's a better place to put it.

---

The root reason for modifying the time is that the sky-side card recording DVR stores files according to the system time. When offline, the system always use defaults time, resulting in chaotic connection of video files. Possible solutions include:

1, Modify the time
1.1, Manually modify the time
1.2, Automatically correct the time
1.2.1, Use GPS data to synchronize the time
1.2.2, The receiving end (Android) pushes the time
1.2.3, Use the existing DVR to infer the time, at least avoid duplication
2, Modify the algorithm
2.1, Detect the existence of the file and append a number
2.2, Do not use time as the file name
This pr is method 1.2.3, method 1.1 send in msposd pr #58

---

修改时间的根本原因是天空端卡录DVR按系统时间存储文件，离线时系统总是默认时间，导致视频文件混乱连接在一起。可能的解決方式包括：
1,修改时间
1.1,手动修改时间
1.2,自动修正时间
1.2.1，使用GPS数据同步时间
1.2.2，接收端（安卓）推送时间
1.2.3，借助已有DVR推测时间，至少避免重复
2,修改算法
2.1，检测文件存在，追加数字
2.2，不适用时间作为文件名
本pr为方法1.2.3,方法1.1已经在msposd pr #58 提交

### DIFF
--- a/general/overlay/etc/rc.local
+++ b/general/overlay/etc/rc.local
@@ -11,4 +11,18 @@
 #
 # By default this script does nothing.
 
+target_dir="/mnt/mmcblk0p1"
+[ ! -d "$target_dir" ] && echo "not find" && exit 1
+
+latest_time=$(find "$target_dir" -exec stat -c '%Y' {} \; 2>/dev/null | sort -nr | head -1)
+[ -z "$latest_time" ] && echo "no sub dir" && exit 0
+echo latest_time $latest_time
+
+current_time=$(date +%s)
+echo current_time $current_time
+
+if [ "$latest_time" -gt "$current_time" ]; then
+    date -s "@$latest_time" >/dev/null && echo "ok" || echo "fail"
+fi
+
 exit 0


robin-debug:
> Please check the current implementation here: https://github.com/OpenIPC/firmware/blob/master/general/overlay/etc/init.d/S30customizer#L5

As I said before, the original implementation is to set the system time to the firmware release time every time it starts, such as 25-1-7 1:00. After I fly twice, the DVR file will be in 2025-1-7 of the SD card, and the video fragments will be mixed together.
And my code checks the existing files on the SD card, including the latest time of the DVR file. For example, the first time the DVR has video from 1-7 1:00 to 1:10, the next time it will continue to save from 1:10, not from 1:00.
Although this cannot obtain accurate time, it can ensure that the DVR file will not be confused.

正如我前面说的，原有实现是每次启动将系统时间设置到固件发布时间，比如我的25-1-7 1:00。当我飞行两次后，dvr文件将会在sd card的2025-1-7里面，视频碎片混合在一起。
而我的代码检查sd卡已有的文件，包含dvr文件的最新时间。比如第一次dvr从1-7 1:00到1:10有视频，下回就会从1:10开始继续保存，不会从1:00开始。
这样虽然不能获得准确的时间，但能保证dvr文件不会混乱。

robin-debug:
> Hello Thank you for your interest in our project. Unfortunately, we will not be able to accept your PR as a default in the main branch, as it will conflict with the ability of various implementations and approaches to setting the time (batteries, GPS, synchronization via the Internet). You can put this script in the Sandbox repository and upload it or use the Builder profiler if you have a customized device. Thank you for understanding.


I understand your choice. What I want to explain is that,
1. This code is written in rc.local. According to init.d, it will be after all the initialization codes and should not affect other functions.
2. The code protects against path failures, and the system time will only be modified if the file time is newer than the system time.
3. As the air/ground end of the aircraft, it is a rare scenario to have an Internet connection, and more work is done offline. The existing Internet connection to update the time is almost useless. I sincerely hope that the function of effectively correcting the time can be completed as soon as possible.

Finally, I posted the code here. Even if it is not accepted, it can be seen by more people, which is also good.


我理解你的选择。我要说明的是，
1，这个代码写在rc.local，根据init.d，他会在所有初始化代码之后，应该不会影响其他的功能。
2，代码中保护了路径失败，也只有文件时间比系统新，才会修改系统时间。
3，作为飞行器的天空端/地面端，有会联网是很稀有的情景，更多的离线工作。现有联网更新时间几乎没用。我衷心希望能早日完成有效修正时间的功能。

最后，我把代码贴在这里，即便不被接受，也能被更多的人看到，这也不错。

flyrouter:
Thanks for your clarification.
I think for the purposes of the FPV system your solution is more suitable for integration into the tweaker that @cronyx runs than into the main rc.local, which should be empty by default. if we are talking about a global profile.
In any case, we will discuss this within the development team. Thanks.

flyrouter:
Good day
As a final universal solution, we suggest you consider launching a command via a tunnel from the Ground Station to the device during startup and synchronization.
This will be the most correct and acceptable from a technical point of view.

general/overlay/etc/rc.local
mikecarr, line 1: This works on my FPV drone

flyrouter:
Good day

Starting with Majestic master+9f0308f, 2025-05-11, a new "notime" option is supported that allows creating files with incremental names, which is more correct.
After testing and reporting, this option can be enabled by default for all FPV systems.

Where it is possible to transmit the correct time, for example in the Android attachment or Windows, you can set the status of this option to "false" to save files with the time name.

```
cli -s .records.enabled true
cli -s .records.split 1
cli -s .records.notime true
```


```
root@openipc-robot:~# tree /mnt
/mnt
└── mmcblk1p1
    ├── 00000.mp4
    ├── 00001.mp4
    ├── 00002.mp4
    ├── 00003.mp4
    ├── 00004.mp4
    ├── 00005.mp4
    ├── 00006.mp4
    └── 00007.mp4
```

PR #1750, opened 2025-03-11
Update to Ruby 10.6, update sysupgrade procedure

PetruSoroaga:
Fix sysupgrade bug, updated Ruby code.
The /etc/rc.local.stop (to stop Ruby) is created at runtime by Ruby, if missing. (to cover OTA updates of existing installs too).


flyrouter (commented):
Thank you

PR #1758, opened 2025-03-17
Update onvif-simple-server

csev1755:


flyrouter:
Good evening.
Have you checked this version. Is everything ok?
What interesting things will it give us?

PR #1770, opened 2025-03-24
[no ci] Update S40network

flyrouter:


flyrouter:
It is suggested to add a static MAC from the ENV bootloader settings, since otherwise it will be constantly random, which causes many problems.
The variable name can be changed, everything is negotiable. 
The proposal is given as a basic concept of the option.

PR #1773, opened 2025-03-26
osd-openipc: adapting the package to the revised project

wberube:
Based on the package build files for divinus, to match the renewed osd project that uses dynamic linking


flyrouter (approved):
Thank you

PR #1806, opened 2025-04-30
Allow changing TX power dynamically on 8812eu and 8733bu

zhouruixi:
Now when loading the 8812eu and 8733bu drivers, use `rtw_tx_pwr_by_rate=2` and `rtw_tx_pwr_lmt_enable=2` parameters, which will cause the user to be unable to modify the tx power to the expected value. Please check @viktorxda 

https://github.com/libc0607/rtl88x2eu-20230815/blob/v5.15.0.1/README.md?plain=1#L46-L47
https://github.com/OpenIPC/firmware/blob/master/general/package/legacy/wifibroadcast/files/wifibroadcast#L44

### DIFF
--- a/general/package/wifibroadcast-ng/files/wifibroadcast
+++ b/general/package/wifibroadcast-ng/files/wifibroadcast
@@ -50,8 +50,8 @@ load_modules() {
 	sleep 1
 
 	if [ "$driver" != "88XXau" ]; then
-		opt1="rtw_tx_pwr_by_rate=2"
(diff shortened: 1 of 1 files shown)


zhouruixi:
> However, not all adapters support or include efuse settings, so it will be reverted.

Yes, a friend in the community uses Runcam WiFiLink and sets txpower to 63. If efuse is used, the maximum output power is only ~13dbm (20mw), and if efuse is not used, the power can reach ~27dbm (525mw). Thanks for merging

PR #1814, opened 2025-05-10
Added default resolution and corrected exposure for default 60fps

MarioFPVdev:
Added default resolution
Added default gop
Corrected exposure for default 60fps

### DIFF
--- a/general/package/wifibroadcast-ng/files/wifibroadcast
+++ b/general/package/wifibroadcast-ng/files/wifibroadcast
@@ -127,8 +127,10 @@ start_telemetry() {
 video_settings() {
 	sensor=$(ipcinfo -s)
 	cli -s .isp.sensorConfig /etc/sensors/"$sensor"_fpv.bin
-	cli -s .isp.exposure 5
+	cli -s .isp.exposure 16
+	cli -s .video0.size 1920x1080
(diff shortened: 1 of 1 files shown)


flyrouter (commented):
Thank you for your suggestions, they will be discussed soon.

MarioFPVdev:
Currency the static resolution of the Thinker Tiny is 704x576. I don't know why, you may ask Dmitry or Viktor.

I also agree that the default resolution should be defined based on the sensor but for the time being it should be 1920x1080 since all the bin file sensors have been calibrated based on that resolution. 

The 1440x1080 resolution will require different bin files for good video quality. I have tested the 4:3 1440x1080 resolution with the imx335 Thinker Tiny with the existing sensor bin files and the video was overexposed. It was also in box view on a 16:9 resolution monitor. The bin files in 1920x1080 resolution was fine as we would expect.

flyrouter:
Also discussion here - https://t.me/c/1809358416/28287/218235

MarioFPVdev:
> To handle sensor variants, this change will be proposed in /usr/bin/wifibroadcast for sbc 2.0 release in 2-3 weeks.

Is this proposal now included in the SBC 2.0?

PR #1826, opened 2025-06-06
[no ci] Update automount.sh

viktorxda:


flyrouter (commented):
Thank you very much for your research.
I have no objections, you can do Merge.
If someone needs to do a fail-safe record, he can enable sync

PR #1853, opened 2025-08-08
Ingenic: initial support for T23

viktorxda:


flyrouter (approved):
As always, great job
Thank you.

PR #1857, opened 2025-08-10
[no ci] Package: move motor sample to legacy

viktorxda:


flyrouter (commented):
Thank you
I will add some wishes tonight

fss-hacks:
Is there any way to move this out of local and provide this like other packages? I went thru hell and back trying to create a build that includes this - I know it's legacy and niche etc but it made everything severely complicated/

PR #1878, opened 2025-09-04
support for ec25 series and sim7600g

danielbanar:


danielbanar:
worked for me

flyrouter (commented):
Good day
Thank you very much, your patch will be reviewed soon
Why do you raise the usb0 interface in the script?
If the modem type is specified, then it should rise automatically.

danielbanar:
@flyrouter sorry I use modified S40network, because if i have modem and ethernet connected it DOSes my home network because it doesnt like 192.168.2.10
ill remove it to match the other modems

PR #1886, opened 2025-09-16
fake-hwclock implementation

danielbanar:
Hi,

I noticed that when the camera boots, it starts with the firmware compilation time.

This causes problems when the device is not connected to the internet. When recording videos to a file, the system time is always the same at boot, which results in new recordings being concatenated into the same file instead of creating separate files.

Currently, the time is set in `S30customizer` from `/etc/os-release`. I added a script that restores the system time from a saved timestamp file on boot and updates this file periodically. This ensures that the time at boot is set to the last known time the camera was powered on, although the restored time might be outdated.

Personally, I don’t like the name `fake-hwclock`, but it’s called that on Debian systems, including Raspberry Pi. I would have preferred something like `sw-rtc` instead.


flyrouter:
Good afternoon

Thank you for your suggestion; it's quite valuable because you write good comments outlining your thoughts.

Yes, the package can be considered and integrated, but I think we should remove the "-openipc" ending as it's too long and unnecessary. I think fake-hwclock or sw-rtc would be sufficient as they don't conflict with other packages. Perhaps this simple sample will be useful to someone, but we won't include it in the firmware by default.

I'd also like to point out that the streamer settings include a "notime" flag in the file writing section, which allows you to be completely independent of the time and avoid using an incorrect date or a lagging time. You might find this interesting.

openipc-ai:
The idea from this PR has landed on master via #2336, which builds directly on the diagnosis here — thank you @danielbanar, you're credited as co-author on the merged commit. The final design keeps your load/save/boot-restore shape but bounds the flash writes by the overlay medium (hourly on a NOR jffs2 overlay, rather than every 5 s), adds busybox ntpd's `-S` hook so the first sync is checkpointed immediately, restores forward-only from the newest of the checkpoint, `/etc` mtimes and the SoC RTC, and runs by default on every camera. Closing this as superseded.

PR #1988, opened 2026-04-10
hisilicon-opensdk: fix module install order for hi3516cv500

widgetii:
## Problem

opensdk modules are not overwriting vendor osdrv modules because in buildroot per-package mode, install order depends on package dependencies. Without an explicit dependency, osdrv's per-package target wins during the final merge.

## Fix

Add `HISILICON_OPENSDK_DEPENDENCIES += hisilicon-osdrv-hi3516cv500` (conditional on CV500 being enabled) so opensdk builds and installs after osdrv. This ensures opensdk's renamed modules overwrite the vendor ones in `hisilicon/`.


widgetii:
Superseded by #1987 which is already merged.

PR #2012, opened 2026-04-16
sysupgrade: add NAND flash support

widgetii:
## Summary

- `sysupgrade` previously hardcoded NOR/squashfs firmware downloads and used `flashcp` for all flash types, which bricks NAND cameras (boot expects `ubifs`, not `squashfs`)
- `download_firmware()` now selects the correct NAND package (`$soc-nand-$osr`) when `ipcinfo -F` reports `nand`
- `do_update_kernel()` uses `flash_eraseall` + `nandwrite -p` for NAND instead of `flashcp`
- New `do_update_rootfs_nand()` function: pivot_roots to tmpfs, detaches UBI, erases the MTD partition, writes the UBI image with `nandwrite`, and reboots
- `do_wipe_overlay()` uses `ubiupdatevol -t` for NAND overlay wipe instead of `flash_eraseall -j` (JFFS2-specific)

## Known limitation

NAND rootfs update erases the entire UBI partition, which includes both rootfs and rootfs_data (overlay). This means **overlay data (settings, SSH keys, etc.) is lost** during rootfs update. Preserving overlay would require `ubiupdatevol` with a bare UBIFS image, which is not currently included in release packages. Re-attaching UBI after `nandwrite` to restore overlay data is unsafe — it causes `image_seq` mismatch and kernel panic.

## Test log (hi3516av200, 128MB SPI NAND)

<details>
<summary>sysupgrade -k -r -f -z output</summary>

```
OpenIPC System Updater v1.0.48

Vendor  hisilicon
SoC     hi3516av200
Kernel  23:23:49 2026-04-15
RootFS  master+1783691, 2026-04-15

Stop services, sync files, free up memory
Stopping crond: OK
Stopping ntpd: OK
Stopping klogd: OK
Stopping syslogd: OK

Firmware
Download from https://github.com/OpenIPC/firmware/releases/download/latest/openipc.hi3516av200-nand-ultimate.tgz
Received and unpacked

Kernel
Update kernel from /tmp/uImage.hi3516av200
Erasing kernel partition...
Erasing 128 Kibyte @ 300000 - 100% complete.
Writing kernel...
Writing at 0x001c0000
Kernel updated to 11:17:22 2026-04-16

RootFS
Update rootfs (NAND) from /tmp/rootfs.ubi.hi3516av200
UBI partition: /dev/mtd3 (mtd3)
Preparing tmpfs root for NAND upgrade...
Performing pivot_root to tmpfs and flashing NAND...
Unmounting old root filesystems...
[...]

PR #2022, opened 2026-04-26
sysupgrade: NAND user-data preservation + SquashFS-on-UBI for HiSilicon

widgetii:
> Supersedes #2012, which is being closed in favour of this comprehensive rework.

## Why

#2012 added basic NAND support to `sysupgrade` but every `-r` wiped the entire UBI MTD and the user's overlay with it. The maintainer's review captured the dilemma:

> «оно там ломает все. или надо переходить на пивот везде и отказываться от gluebi или использовать как есть»

This PR takes the "do it properly" path. The changes here let a NAND camera survive `sysupgrade -r` with config intact, the same way NOR cameras already do — and they bring the HiSilicon NAND firmware in line with the architecture sigmastar/rockchip already use.

Two coupled root causes need to be fixed together; neither stands on its own:

1. **Wrong upgrade primitive.** `flash_eraseall + nandwrite` on a UBI MTD leaves UBI's autoresize unable to reclaim tail PEBs cleanly. On reboot the kernel panics with `bad image sequence number in PEB N`. That brick is reproducible on this branch — see the recovery write-up below.
2. **Wrong rootfs layout.** Default HiSilicon `ubinize.cfg` shipped a single read-write UBIFS rootfs volume, so even with a correct primitive there was no clean place to land preserved user data: any upgrade rewrote the same filesystem the user was editing.

## What this PR achieves

- `sysupgrade -r` on NAND preserves user config across upgrades by default. `-n` opts in to wiping. (Same semantics as NOR.)
- HiSilicon NAND ships SquashFS-on-UBI rootfs + UBIFS `rootfs_data` for the overlay — matching sigmastar (`ubinize_sigmastar.cfg`) and rockchip (`ubinize_rockchip.cfg`).
- `sysupgrade` uses `ubiformat -y -f` (the OpenWrt `nand.sh` pattern), which handles erase counters and writes one fresh `image_seq` across every PEB. No more `image_seq` panics.
- Existing UBIFS-rootfs cameras get a one-shot bootargs migration via `fw_setenv` — runs **after** the flash succeeds so a flash failure can never leave the env pointing at a non-existent rootfs.
- `/etc/sysupgrade.conf` is a user-extensible keep-list (one path per line), mirroring OpenWrt.

## How it works

```
[old camera] sysupgrade -r
    │
    ├─ tar /etc/passwd /etc/dropbear /etc/majestic.yaml /root … → /tmp/sysupgrade.tgz
    ├─ compute new bootargs (idempotent; empty if no change)
    ├─ pivot_root → tmpfs (busybox + ubiformat + fw_setenv + fw_env.config + rootfs.ubi + sysupgrade.tgz)
    │
    ├─ ubidetach
    ├─ ubiformat -y -f rootfs.ubi /dev/mtdN          # consistent image_seq across all PEBs
    ├─ fw_setenv bootargs "<migrated>"               # only after flash success
    ├─ ubiattach + wait for autoresize to mat rootfs_data
    ├─ mount ubi0:rootfs_data /tmp/data
    ├─ cp sysupgrade.tgz /tmp/data/
    └─ reboot

[new firmware /init]
    ├─ mount ubi0:rootfs_data /overlay
    ├─ if /overlay/sysupgrade.tgz: tar -xzf … && rm -f      # one-shot, then no-op forever
    ├─ overlayfs lower=/ (squashfs ro) upper=/overlay (UBIFS rw)
    └─ pivot_root → /sbin/init
```

Reference: [OpenWrt `package/system/procd/files/nand.sh`](https://github.com/openwrt/openwrt/blob/master/package/system/procd/files/nand.sh).

### Online upgrade from a previous OpenIPC NAND camera

```sh
sysupgrade -k -r --force_all
```

That's it. The script:

- backs up the keep-list paths,
- replaces the rootfs UBI MTD via `ubiformat`,
- migrates U-Boot `bootargs` automatically (`root=ubi0:rootfs rootfstype=ubifs` → `root=/dev/ubiblock0_0 rootfstype=squashfs ubi.block=0,0`),
- stages the user-data tarball into `rootfs_data`,
- reboots.

After reboot, the new firmware's `/init` extracts the tarball into the overlay and the camera comes up with config intact.

### To wipe overlay on upgrade (factory-reset semantics)

```sh
sysupgrade -k -r -n --force_all
```

`-n` skips the keep-list step; `rootfs_data` is empty after the flash.

### Custom keep-list

Drop additional paths to preserve into `/etc/sysupgrade.conf` (one path per line, files or directories):

```
/opt/myscripts/
/etc/cron.d/extra
```

## Migration from the legacy UBIFS-rootfs layout

Existing deployed cameras don't ship `ubiformat` (it's introduced by this PR via `BR2_PACKAGE_MTD=y`). The first hop from old → new firmware therefore runs the legacy `flash_eraseall + nandwrite` path. This single upgrade is **destructive to user data** (the script logs a clear warning). Every subsequent upgrade on the new firmware preserves correctly because `ubiformat` is now installed.

The bootargs migration is idempotent (a re-run on already-migrated env is a no-op), so it's safe to invoke from automation.

## Recovery if something goes wrong

If the camera fails to boot after upgrade, U-Boot TFTP recovery still works:

```
=> setenv serverip <your-tftp-host>
=> setenv ipaddr <camera-ip>
=> run urnand
```

`urnand` re-flashes the same `rootfs.ubi.<soc>` image; combined with this PR's `ubiformat` semantics it lands cleanly. The flash MTD partition is never touched until *after* `ubiformat` validates the input image, so a network/checksum failure during download cannot brick the camera.

## Scope

| Affected | Layout | sysupgrade flow |
|---|---|---|
| HiSilicon NAND ultimate (av100/av200, cv300, dv100, ev200/ev300, hi3518 ev200/ev300) | switched to SquashFS-on-UBI | preserving by default |
| Sigmastar / Rockchip NAND | already SquashFS-on-UBI; layout untouched | preserving by default (uses same code path) |
| All NOR boards | unchanged | unchanged |

## Out of scope (follow-ups)

- Dropping `CONFIG_MTD_UBI_GLUEBI` from kernel configs — this PR removes sysupgrade's *reliance* on gluebi-exposed MTDs; unsetting the kconfig globally touches every NAND-capable kernel config and risks breaking unrelated consumers (notably the JFFS2-on-UBI fallback in `/init`).
- Lite NAND variants — same recipe applies once defconfigs exist.

## Files modified

| File | Change |
|---|---|
| `general/scripts/ubifs/ubinize_hisilicon.cfg` | new — SquashFS rootfs + autoresize rootfs_data |
| `br-ext-chip-hisilicon/configs/{hi3516av100,hi3516av200,hi3516cv300,hi3516dv100,hi3516ev200,hi3516ev300,hi3518ev200,hi3518ev300}_ultimate_defconfig` | repoint at new ubinize cfg + `BR2_PACKAGE_MTD=y` (ships `ubiformat`) |
| `general/overlay/usr/sbin/sysupgrade` | `build_keep_list` + `compute_new_bootargs` helpers; `do_update_rootfs_nand` rewritten to ubiformat + tarball pattern |
| `general/overlay/init` | regex broadened to enter UBI overlay branch on squashfs+ubi.block layouts; one-shot `/overlay/sysupgrade.tgz` extractor |
| `general/overlay/etc/sysupgrade.conf` | new empty file with format docs |
| `general/package/hisilicon-osdrv-hi3516ev200/files/script/set_allocator` | additive form — preserves all bootargs keys except `mem`/`mmz_allocator`/`mmz` so `ubi.mtd=`/`ubi.block=` survive `load_hisilicon` invocations |

## Test plan

- [ ] `make BOARD=hi3516av200_ultimate` produces a `rootfs.ubi.hi3516av200` whose first volume is SquashFS (verify with `ubinize -v` or `unubi`).
- [ ] On a hi3516av200 NAND camera at known IP: customise (set root password, add `/root/canary.txt`, edit `/etc/majestic.yaml`, append `/opt` to `/etc/sysupgrade.conf`, drop `/opt/marker.txt`).
- [ ] SCP the new `sysupgrade` to the camera; run `sysupgrade -k -r --force_all`; capture UART.
- [ ] Verify on reboot: no `bad image sequence number` panic, root password persists, `/root/canary.txt` exists, `/etc/majestic.yaml` retains edits, `/opt/marker.txt` exists. `ubinfo -a` shows two volumes (`rootfs` SquashFS, `rootfs_data` autoresized). `/proc/cmdline` shows `root=/dev/ubiblock0_0 rootfstype=squashfs`. `mount` shows `overlayfs` with `lowerdir=/` and `upperdir=/overlay`.
- [ ] Wipe path: `sysupgrade -k -r -n --force_all` then verify root password reset and `/root/canary.txt` gone.
- [ ] Idempotent re-run from new firmware: `sysupgrade -k -r --force_all` — `fw_setenv` is a no-op, bootargs unchanged.
- [ ] Brick recovery via U-Boot TFTP `run urnand` works (kept compatible).


widgetii:
## End-to-end test results on hi3516av200 NAND

Tested on a real hi3516av200 NAND camera via UART + SSH. The fix-up commit (`ecdf7f90`) lands several bugs that surfaced during testing.

**Working ✅**

- `make BOARD=hi3516av200_ultimate` builds with the new SquashFS-on-UBI layout (`rootfs.ubi` 6.7 MB; `rootfs.squashfs` 6.4 MB; `uImage` 1.9 MB).
- Legacy first-hop upgrade (camera without `ubiformat`) flashes new firmware via `flash_eraseall + nandwrite`. Camera reboots into the new layout.
- `compute_new_bootargs` migrates `root=ubi0:rootfs rootfstype=ubifs` → `root=/dev/ubiblock0_0 rootfstype=squashfs ubi.block=0,0 ubi.mtd=N,2048 init=/init`. Idempotent on already-migrated cameras (verified — no-op re-run).
- `fw_setenv` writes the migrated bootargs after a successful flash (binary + `/etc/fw_env.config` staged into tmpfs pre-pivot).
- `/init` mounts `ubi0:rootfs_data` and overlays squashfs → camera comes up with `overlay on / type overlay (rw, lowerdir=/, upperdir=/overlay/root)`. Confirmed via `mount`.
- `ubiformat -y -f` (with leading `flash_eraseall`) writes consistent `image_seq` across all PEBs. No more `bad image sequence number` kernel panics on subsequent boots.
- `set_allocator`'s additive form preserves `ubi.mtd=`/`ubi.block=` across `load_hisilicon` runs.
- `BR2_PACKAGE_MTD=y` ships `/usr/sbin/ubiformat` in the new firmware.

**Known limitation ⚠️**

The post-pivot UBIFS write that stages `/tmp/sysupgrade.tgz` into `rootfs_data:/root` returns `EINVAL` from the busybox shell on this kernel (3.18.20)/firmware combination. The `/init` overlayfs path itself works perfectly in normal runtime — only the post-pivot tmpfs context hits this. Investigation pointed at filesystem-level issues that would need significant kernel-side debugging to resolve cleanly.

The script logs `Warning: rootfs_data mount failed; user data not preserved.` when this happens; the rest of the upgrade still completes and the camera boots into the new firmware. **No data destruction beyond the rootfs replacement itself** — `rootfs_data` survives, just isn't pre-populated with preserved files.

**Recommended follow-up**

Drop the post-pivot UBIFS write entirely and switch to `ubiupdatevol /dev/ubi0_0 < rootfs.squashfs` to update **only** the rootfs volume, leaving `rootfs_data` untouched throughout. Requires shipping `rootfs.squashfs.${soc}` alongside `rootfs.ubi.${soc}` in the release tarball (small Makefile change to `REPACK_FIRMWARE`). With that pattern there is no UBIFS write during sysupgrade at all and user data is preserved by construction.

**Net effect of this PR**

Even with the known limitation, this PR is a strict improvement over PR #2012:

1. Layout is correct (squashfs+overlay, matching sigmastar/rockchip).
2. Flash mechanism is correct (`ubiformat`, no more `image_seq` panics).
3. First-boot init pivots overlayfs cleanly.
4. Bootargs migration is automatic and idempotent.
5. The infrastructure for user-data preservation (`/etc/sysupgrade.conf`, keep-list, tarball) is in place — the follow-up just swaps the staging mechanism.

Brick recovery via U-Boot TFTP `run urnand` works (tested), with the caveat that the `urnand` env on existing deployed cameras targets `0x400000` while the current mtdparts has UBI at `0xA00000`. Operators recovering an existing camera need to use explicit offsets:

```
=> tftpboot ${baseaddr} rootfs.ubi.${soc}
=> nand erase 0xa00000 0x7600000
=> nand write ${baseaddr} 0xa00000 ${filesize}
=> reset
```

widgetii:
## Update: user data preservation now works end-to-end ✅

Following maintainer review and the prior comment's recommended follow-up, switched to the **single-volume update** pattern: ubiupdatevol on `/dev/ubi0_0` to rewrite only the rootfs squashfs, leaving the entire `rootfs_data` UBI volume — and the overlayfs upper layer it backs — untouched.

This sidesteps every UBIFS write quirk hit on the previous attempt and preserves user data **by construction** rather than by tarball stash.

### What changed in commit `1353b24a`

- `Makefile`: NAND release tarball now also ships `rootfs.squashfs.${soc}` alongside `rootfs.ubi.${soc}` (extra-file slot added to `PREPARE_REPACK` / `REPACK_FIRMWARE`).
- `sysupgrade` picks one of three upgrade modes for `do_update_rootfs_nand`:
  - **`updatevol`** — when `rootfs.squashfs` is in the release tarball **AND** `rootfs_data` already exists. `ubidetach + ubiattach + ubiupdatevol /dev/ubi0_0`. Fastest, preserves `rootfs_data` by construction. **This is the steady-state path.**
  - **`ubiformat`** — first-hop migration from legacy UBIFS-rootfs firmware to the new layout. Full UBI MTD rewrite. Destructive to `rootfs_data` (which doesn't exist on legacy cameras anyway, so nothing to preserve).
  - **`legacy`** — pre-this-PR firmware that doesn't ship `mtd-utils`; `flash_eraseall + nandwrite`. Same one-time destruction as `ubiformat`; subsequent upgrades hit `updatevol`.
- The `build_keep_list` helper and `/etc/sysupgrade.conf` overlay file are dropped — with `rootfs_data` preserved wholesale, an explicit keep-list is unnecessary.

### Verification on hi3516av200 NAND

Staged before upgrade:
- root password changed to a non-default value
- `/root/canary.txt` (custom file)
- `/opt/marker.txt` (file in user-only directory)
- `/etc/persistent_test` (overlay-modified system file)

After `sysupgrade --archive=…`:
- ✅ SSH login with the custom password works
- ✅ All three files present with original content and timestamps
- ✅ `uname -r` shows the new kernel build (#10, was #8)
- ✅ `ubinfo` confirms `rootfs_data` is still 655 LEBs (its previously-autoresized size — the volume itself was never touched)
- ✅ No `image_seq` panic on reboot
- ✅ `/proc/cmdline` unchanged (camera was already migrated; `compute_new_bootargs` correctly returned empty)

Sysupgrade output excerpt:
```
RootFS
UBI partition: /dev/mtd3 (mtd3)
Update rootfs (NAND, updatevol) from /tmp/rootfs.squashfs.hi3516av200
Preparing tmpfs root for NAND upgrade...
Performing pivot_root to tmpfs and flashing NAND...
[post-pivot, on UART]
Detaching UBI from mtd3...
Re-attaching UBI to update rootfs volume in place...
ubiupdatevol /dev/ubi0_0 with rootfs.squashfs...
Rootfs volume updated; rootfs_data left intact.
NAND upgrade complete. Rebooting...
```

The "known limitation" note in the prior comment is now resolved.

PR #2043, opened 2026-05-04
txw8301-openipc: add Taixin TXW8301 802.11ah HaLow Wi-Fi FMAC driver package

csvke:
Adds a Buildroot package under `general/package/txw8301-openipc` with firmware blobs, default config, and init script. Also enables TXW8301 on `gk7205v300_ultimate` and `ssc338q_ultimate`. 

**TXW8301 FMAC firmware:** https://github.com/TXW8301/TXW8301-FMAC-SDK
**TXW8301 Linux Driver:** https://github.com/TXW8301/TXW8301-FMAC-linux-driver

Tested on a `XM IVG-G5F` `GK7205V300` camera board using it's `J3 Extension ports` for both SDIO and USB interfaces connecting with `TAIXIN-AH-RX00P_EVB_V1.7`

IMG_3680
IMG_3582 2

@cronyx @FlyRouter @viktorxda


flyrouter (commented):
Good afternoon

Thank you very much for your interest in our project and your contribution.
Your PR will be reviewed shortly by our team.

A brief review of your files

Files that can be accepted into the Firmware repository immediately:

- general/package/Config.in
- general/package/txw8301-openipc/Config.in
- general/package/txw8301-openipc/files/S35txw8301
- general/package/txw8301-openipc/hgicf.conf.default
- general/package/txw8301-openipc/hgicf_sdio.bin
- general/package/txw8301-openipc/modprobe-txw8301.conf
- general/package/txw8301-openipc/txw8301-openipc.mk

Regarding the two files:

- br-ext-chip-goke/configs/gk7205v300_ultimate_defconfig
- br-ext-chip-sigmastar/configs/ssc338q_ultimate_defconfig

We would not like to include them in the current configuration of the Ultimate firmware for the following reasons:

1. It is possible that after some time the Ultimate firmware strategy itself will be changed or they may be abolished altogether.
2. If we accept your PR for these two files, any users who have installed the Ultimate firmware will have drivers inside that they may never need.

We suggest an alternative approach:

1. Remove these two files from your PR and it will be accepted immediately.
2. Please take a look at the Builder repository, and it might be more appropriate to create your two HaLow configurations there and maintain them under your guidance and management. We believe this is the right approach and will give you more freedom of action.

Please do not hesitate to comment on the suggestion.

Thank you for your contribution!

csvke:
Thank you for the thorough review and for the thoughtful suggestion regarding the defconfig files. Sorry I have been tied up with other stuff.

I have removed br-ext-chip-goke/configs/gk7205v300_ultimate_defconfig and br-ext-chip-sigmastar/configs/ssc338q_ultimate_defconfig from the PR. The branch now contains only the package files as listed in your accepted list.

Regarding the Builder repository approach — I agree this makes more sense for board-specific HaLow configurations, as it keeps them decoupled from the Ultimate firmware strategy and allows them to be maintained independently. I will look into creating the two HaLow configurations there.

The PR should now be ready for acceptance at your convenience. Please let me know if any further changes are needed.

Thank you again for your guidance!

widgetii (commented):
Thanks for this — the package itself is clean and follows the repo's kernel-module conventions well. With the two `*_ultimate_defconfig` files now dropped per @flyrouter's review (board enablement moving to the Builder repo), the integration side looks good. A few things I'd like to see addressed or clarified before merge:

**1. `/etc/hgicf.conf` is installed but never consumed — the device won't actually associate.**
`hgicf.conf.default` ships `ssid` / `wpa_psk` / `key_mgmt` / `freq_range`, and the `hgicf`/`hgpriv` tools are installed to `/usr/bin`, but `S35txw8301` only loads the module, brings `hg0` up, and runs `udhcpc`. Nothing applies the association parameters, and the `modprobe.d` override only sets `fw_file=` (no path pointing the driver at the conf). As written, a fresh board loads the module but joins no network. Could you either wire `hgicf`/`hgpriv` into the init script to apply the config, or document how the firmware/driver ingests `/etc/hgicf.conf`?

**2. `choice` defaults to USB, but the hardware was tested on SDIO.** Minor — just flagging the default won't match the documented bring-up setup.

**3. Test-app output path is slightly fragile.** `TXW8301_OPENIPC_CC_NAME = $(notdir $(TARGET_CC))` matched against `bin/$(notdir $(CC))/` will break if `TARGET_CC` ever carries a ccache prefix (`notdir` would yield `ccache gcc`). Fine for the current toolchain config, but pinning the test_app output dir explicitly would be more robust.

**4. Cosmetic:** `LICENSE = GPL-2.0, proprietary (hgicf firmware blob)` isn't standard SPDX — convention here is just `GPL-2.0` (`REDISTRIBUTE = NO` already covers the blob's status). No `LICENSE_FILES` set either.

**Verified working as intended:** the custom `BUILD_CMDS`/`INSTALL_TARGET_CMDS` don't clobber the module build/install (the infra appends those via `POST_*_HOOKS`); `MODULE_SUBDIRS = hgic_fmac` + `CONFIG_HGIC_SDIO/USB` forward correctly; the `.ko` installs to `/lib/modules/<ver>/extra/` matching the init script's `insmod` fallback; and `fw_file=taixin/hgicf.bin` matches the blob install path.

Item **#1** is the main one — happy to merge once the association path is wired up or explained.

csvke:
> Thanks for this — the package itself is clean and follows the repo's kernel-module conventions well. With the two `*_ultimate_defconfig` files now dropped per @flyrouter's review (board enablement moving to the Builder repo), the integration side looks good. A few things I'd like to see addressed or clarified before merge:
> 
> **1. `/etc/hgicf.conf` is installed but never consumed — the device won't actually associate.** `hgicf.conf.default` ships `ssid` / `wpa_psk` / `key_mgmt` / `freq_range`, and the `hgicf`/`hgpriv` tools are installed to `/usr/bin`, but `S35txw8301` only loads the module, brings `hg0` up, and runs `udhcpc`. Nothing applies the association parameters, and the `modprobe.d` override only sets `fw_file=` (no path pointing the driver at the conf). As written, a fresh board loads the module but joins no network. Could you either wire `hgicf`/`hgpriv` into the init script to apply the config, or document how the firmware/driver ingests `/etc/hgicf.conf`?
> 
> **2. `choice` defaults to USB, but the hardware was tested on SDIO.** Minor — just flagging the default won't match the documented bring-up setup.
> 
> **3. Test-app output path is slightly fragile.** `TXW8301_OPENIPC_CC_NAME = $(notdir $(TARGET_CC))` matched against `bin/$(notdir $(CC))/` will break if `TARGET_CC` ever carries a ccache prefix (`notdir` would yield `ccache gcc`). Fine for the current toolchain config, but pinning the test_app output dir explicitly would be more robust.
> 
> **4. Cosmetic:** `LICENSE = GPL-2.0, proprietary (hgicf firmware blob)` isn't standard SPDX — convention here is just `GPL-2.0` (`REDISTRIBUTE = NO` already covers the blob's status). No `LICENSE_FILES` set either.
> 
> **Verified working as intended:** the custom `BUILD_CMDS`/`INSTALL_TARGET_CMDS` don't clobber the module build/install (the infra appends those via `POST_*_HOOKS`); `MODULE_SUBDIRS = hgic_fmac` + `CONFIG_HGIC_SDIO/USB` forward correctly; the `.ko` installs to `/lib/modules/<ver>/extra/` matching the init script's `insmod` fallback; and `fw_file=taixin/hgicf.bin` matches the blob install path.
> 
> Item **#1** is the main one — happy to merge once the association path is wired up or explained.

Hi @widgetii ,

Thank you for the detailed review — all four items have been addressed:

1. — `/etc/hgicf.conf` wiring: The kernel module reads `/etc/hgicf.conf` natively at `insmod` time via an internal `conf_file` module parameter (default value is `/etc/hgicf.conf`). To make this binding unambiguous, I've added `conf_file=/etc/hgicf.conf` explicitly to `modprobe-txw8301.conf` alongside the existing `fw_file=` line, so the config path is visible regardless of how the module is loaded.

2. — Default interface: Changed `Config.in` default from `USB` to `SDIO`, matching the validated hardware setup. (I have also tested `USB` interface and it works.)

3. — `CC_NAME` fragility: Changed to `$(notdir $(lastword $(TARGET_CC)))` so a ccache prefix is stripped before the path basename is taken.

4. — LICENSE cosmetic: Simplified to `GPL-2.0` — `REDISTRIBUTE = NO` already covers the blob's status.

widgetii (commented):
Confirmed all four are in `0d26ea0` — `conf_file=/etc/hgicf.conf` pinned in modprobe.d, SDIO default, ccache-safe `CC_NAME` via `$(lastword ...)`, and clean `GPL-2.0`. Since you've validated association on real hardware for both SDIO and USB, #1 is satisfied. Thanks for the quick turnaround — LGTM from my side. 🚀

widgetii:
@csvke Now that the package is merged — I've written a wiki article to help newcomers with HaLow in general and the TXW8301 specifically: https://github.com/OpenIPC/wiki/blob/master/en/wifi-halow.md

Since you're the one with the actual hardware, I'd really appreciate a review and any amendments:

- **Sanity check** the hookup and configuration sections — especially the J3 header / SDIO wiring notes and the `hgicf.conf` key descriptions.
- **Real bring-up artifacts** would strengthen it a lot: a `dmesg` capture of the driver loading + `hg0` appearing, sample `hgpriv` status output, and rough iperf3 numbers over the link at your `bss_bw` — so readers know what "working" looks like.
- Your test-rig **photos** from this PR would be a great addition to the hardware section, if you're OK with including them (with credit).

Feel free to PR the wiki directly (edits welcome on https://github.com/OpenIPC/wiki — the page is `en/wifi-halow.md`), or just drop corrections/logs here and I'll fold them in. Also happy to link your Builder-repo HaLow configs from the article once they land.

PR #2100, opened 2026-05-17
hisilicon-osdrv: drop vendor .ko install, modernise cv300 load script

widgetii:
## Summary
- Drop the prebuilt vendor `.ko` install paths from `hisilicon-osdrv-hi3516cv300` and `hisilicon-osdrv-hi3516cv500`. Both packages now ship sensor configs, load scripts, and vendor-only userspace `.so` libs only — kernel modules come from `openhisilicon-opensdk` (`open_*.ko`). The 3.18-era fallback was dead code on every current build.
- Rewrite cv300's `load_hisilicon` on the cv500 template: `modprobe open_*` throughout instead of vendor `hi3516cv300_*.ko` / `hi_acodec.ko` / `hi_mipi.ko` / `hi_osal.ko` / `cma_osal.ko` / `sys_config.ko`. This is the exact mismatch that has the cv300 qemu-boot row in openhisilicon CI marked \`allow-failure: true\`.

3 files, +254 / -640 lines. The 26 vendor \`.ko\` binaries under \`files/kmod/\` remain on disk; a follow-up can \`rm\` them once nothing references them.

## Why now
Prerequisite for upcoming cv300_neo (Linux 7.0 kernel) enablement work. Without this cleanup, cv300_neo would need either (a) a kludgy band-aid script to bring \`lo\`/\`eth0\` up since vendor osdrv's \`/etc/network/interfaces\` isn't installed when osdrv kernel-modules-build is assumed to fail, or (b) carrying the wrong-version 3.18 \`.ko\` files into a 7.0 build. Mirroring cv500's clean shape sidesteps both — cv300_neo just enables osdrv like av300_neo does.

Net side-benefit: closes the long-standing \`cma_osal.ko\` vs \`hi_osal.ko\` qemu-boot mismatch on cv300_lite.

## Test plan
- [ ] cv500 lite + av300_neo: osdrv install reduced to what the \`ifeq=y\` branch was already producing; expect no behavioural change. Sysupgrade + boot + camera streams.
- [ ] cv300_lite (real IMX291 board): **consequential test** — new \`load_hisilicon\` does \`modprobe open_*\` for the full module set; verify all \`open_*.ko\` load cleanly via \`lsmod\`, dmesg has no Oops/panic, majestic captures a snapshot, network is up.
- [ ] CI qemu-boot rows for \`hi3516cv300\` and \`hi3516cv500\` — should turn green; if so, the \`allow-failure: true\` markers in openhisilicon \`.github/workflows/build.yml:580-605\` can come off in a follow-up.


widgetii:
Closing — found an unannotated rename-on-install trick in opensdk.mk (cv300 case): opensdk renames open_*.ko to vendor names (hi3516cv300_base.ko etc) at install time, overwriting osdrv's vendor blobs. My load_hisilicon rewrite using modprobe open_* breaks because the modules don't exist with those names on cv300_lite.

The .mk cleanup half of the change is still mostly correct (opensdk overwrites the duplicate paths), but a handful of vendor-only blobs (hi_adv7179, hi_piris, hi_sample_ist, hi_ssp_*, hi_tlv320aic31, virtualhifb) have no opensdk equivalent and would silently disappear. Need an audit and a tighter PR scope.

Reopening with narrower fix shortly.

PR #2135, opened 2026-05-25
[no ci] Change sensor type from stSnsOv02g10Obj to stSnsOs02g10Obj

flyrouter:


widgetii:
The image quality has degraded, so the sensor driver should be replaced with a binary form instead

PR #2143, opened 2026-05-27
Remove duplicated rtl8188fu Config.in

conke:
There are 2 same lines for rtl8188fu in general/package/Config.in:
```
source "$BR2_EXTERNAL_GENERAL_PATH/package/rtl8188fu-openipc/Config.in"
```


widgetii:
Thanks for the catch on the duplicate `rtl8188fu` line — that part of the change is clearly correct.

However, this PR's diff is much wider than the title/description suggests: 117 files, +231/-253, 7 commits. Looking at the commit list, only the last commit (`remove duplicated rtl8188fu Config.in`) matches what the description promises. The others appear to be unrelated downstream changes that came along because the branch was cut from a fork master that had drifted from upstream:

- **`remap OPENIPC_KERNEL url to maxwit-openipc`** rewrites `BR2_TOOLCHAIN_EXTERNAL_URL` and `BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION` in 113 board defconfigs from `github.com/openipc/{firmware,linux}` to `github.com/maxwit-openipc/{firmware,linux}`. Upstream builds need to keep pointing at the OpenIPC org — the MaxWit mirror won't host every kernel/toolchain tag the upstream defconfigs reference, so this would break the matrix.
- **`remove uboot.yml`** deletes the `.github/workflows/uboot.yml` CI workflow with no rationale given.
- **`add /usr/share/openipc/customizer.sh`** introduces a new overlay script that hard-codes MaxWit-specific gpio labels (`wled_pin`, `ircut_pin1`, `ircut_pin2`) and a `fw_printenv -n SN` lookup. Nothing in the base init invokes it, so it would ship as dead code on every board.
- **`add menuconfig support`** (the 4-line Makefile addition) is actually a nice standalone improvement and would be welcome on its own.

Could you split this into two PRs against a fresh branch cut from `OpenIPC/firmware:master`?

1. The 1-line `general/package/Config.in` dedup (this PR's stated goal).
2. The `menuconfig` Makefile target as its own PR.

Everything else (URL remap, `uboot.yml` removal, `customizer.sh`) looks like it belongs in the MaxWit downstream fork, not upstream. Easiest way is usually:

```
git fetch upstream
git checkout -b rtl8188fu-dedup upstream/master
# cherry-pick just the one commit, push, open new PR
```

Happy to re-review once it's narrowed down.

PR #2151, opened 2026-05-27
kc110-board-support: new package for TP-Link Kasa KC110 PTZ camera (hi3518ev200 + OV2735)

moontwister:
## Summary

Adds a new Buildroot package `kc110-board-support` for the **TP-Link Kasa KC110** PTZ IP camera — a HiSilicon Hi3518EV200 board with an OV2735 MIPI sensor and 24BYJ48 stepper motors driven through a ULN2803 darlington array. The package bundles the three small board-specific bits that don't fit any existing package, all of them necessary to get pan/tilt/IR functionality working.

## What it installs

| File | Purpose |
|---|---|
| `/etc/init.d/S94pinmux-pan-ir` | Writes `0x1` to pinmux registers `0x200f00e8/ec/f0/fc/100/104`. The KC110's pan + IR-cut + IR-LED pads (chip7/chip8 region) need this; the kernel pinctrl driver does not set it automatically, and the SoC pinmux state survives soft reboots. Without this script, GPIO writes succeed silently but the pads stay in the wrong function — looking 100 % like a hardware fault. |
| `/usr/bin/ptz` | Small C binary (PL061 mmap, vendor "base-10" GPIO numbering — `bank=N/10`, `pin=N%10`, which on Hi3518EV200 maps cleanly onto OpenIPC's existing PL061 chips). Drives the 4-coil pan and tilt steppers and the IR-cut/IR-LED solenoids. |
| `/usr/bin/gpio-motors` | 451-byte shell shim that forwards the `gpio-motors X Y N` calls used by the web UI's `motor.cgi` to `ptz pan/tilt`. |
| `/usr/bin/ir-control.sh` | Convenience wrapper for the IR-LED + IR-cut filter. |

## KC110 GPIO map (vendor base-10 → OpenIPC PL061 pin)

```
pan motor coils A/B/C/D : 65 / 60 / 59 / 58   (vendor: 81/74/73/72)
tilt motor coils A/B/C/D: 0 / 1 / 2 / 36      (vendor: 0/1/2/44)
IR-cut filter solenoid  : 64                  (vendor: 80)
IR-LED                  : 63                  (vendor: 77)
```

Note that vendor pin `44` translates to OpenIPC pin `36` (chip4 base 32 + pin 4), not pin `44`.

## Conflicts with existing packages

This package installs `/usr/bin/gpio-motors` and therefore conflicts with the existing `BR2_PACKAGE_GPIO_MOTORS`, which writes a sysfs-based binary at the same path. The two cannot be enabled simultaneously. The `Config.in` text mentions this; adopters need to unset `BR2_PACKAGE_GPIO_MOTORS` in their defconfig.

## No defconfig changes

The package is hardware-specific to the KC110 and should not be enabled on generic Hi3518EV200 boards. This PR adds the `source` line to `general/package/Config.in` but does not modify any defconfig — KC110 owners would add the option to their own defconfig:

```
BR2_PACKAGE_KC110_BOARD_SUPPORT=y
# BR2_PACKAGE_GPIO_MOTORS is not set
```

## Hardware testing

Built and live-flashed on a real KC110 device. Verified end-to-end:

- All four package files install to the expected paths in the squashfs (md5-checked).
- `ptz` binary builds reproducibly (same md5 across rebuilds, `e70f78d73c91248ba044082e3f21c54b` in our testing).
- `ptz status`, `ptz pan ±N`, `ptz tilt ±N`, `ptz home` all functional.
- The web UI's motor.cgi buttons correctly invoke the gpio-motors shim → ptz path.
- `S94pinmux-pan-ir` runs at boot; without it, pan + IR fail silently (confirmed by removal + reboot in a separate test).
- IR-cut filter toggle and IR-LED control work via `ir-control.sh` and the majestic `/night/*` endpoints.

## Out of scope

This PR does not include:

- **OV2735 sensor support** for cv200 osdrv's `load_hisilicon` (a separate 3-line patch we maintain downstream — happy to send that as a follow-up PR if there's interest).
- **An LD_PRELOAD MIPI shim** (`libfixmipi.so`) needed on V2A boards because upstream majestic builds `combo_dev_attr_t` with `raw_data_type=0` for MIPI sensors. That's a workaround for an unrelated upstream bug and lives in a separate `kc110-mipi-shim` package downstream, gated behind that root cause.

## Reverse-engineering background

Full notes on how the GPIO map, pinmux constants, and chip layout were reverse-engineered:
<https://github.com/moontwister/kasa-kc110-openipc>


widgetii (commented):
Thanks for the very thorough writeup and the hardware-tested results — the package layout, reproducible build, and reverse-engineering notes are all excellent.

That said, I don't think `OpenIPC/firmware` is the right home for this, and I'd like to suggest you move it before any maintainer time is spent on the merge path here:

- This package is, by your own description, hardware-specific to a single retail model and **must not** be enabled on generic `hi3518ev200_*` boards. The PR adds no defconfig wiring, so on `OpenIPC/firmware` it ships as a `Config.in` entry that no build in this repo's CI matrix will ever exercise.
- It installs `/usr/bin/gpio-motors`, conflicting with the existing `BR2_PACKAGE_GPIO_MOTORS`. Two packages writing the same path that must be manually mutually excluded is exactly the kind of board-pinned divergence we try to keep out of the main firmware tree.
- You already maintain [`moontwister/kasa-kc110-openipc`](https://github.com/moontwister/kasa-kc110-openipc) downstream for the same reason.

The canonical home for exotic / single-board configurations in the OpenIPC ecosystem is **[OpenIPC/builder](https://github.com/OpenIPC/builder)**. It already hosts per-board defconfigs and board-specific packages for variants that don't fit the standard `lite` / `ultimate` / `neo` split — this is also reflected in `OpenIPC/firmware`'s own `sysupgrade` logic, which routes any non-`{lite,ultimate,neo}` variant to the builder repo's releases.

Concretely, I'd suggest:

1. Re-target this PR at `OpenIPC/builder` — drop `general/package/kc110-board-support/*` into the equivalent location under `devices/common/br-ext-chip-hisilicon/` and add a `kc110_<variant>_defconfig` that selects this package (and un-selects `BR2_PACKAGE_GPIO_MOTORS`).
2. Send the OV2735 cv200 `load_hisilicon` patch to the same place so KC110 owners get a single coherent build target.
3. Once that's merged, KC110 firmware images flow naturally through the builder release channel, and the `sysupgrade` path on the camera will already pick them up.

On the LD_PRELOAD MIPI shim — note that an out-of-tree `libfixmipi.so` is unlikely to be accepted anywhere in the OpenIPC tree; the `combo_dev_attr_t` / `raw_data_type` issue should be fixed in upstream `majestic` rather than worked around in a preload. Worth raising that with the majestic maintainers directly.

Happy to help review on the builder side once it's moved over. Again — really nice work on the reverse engineering and on the package itself.

widgetii:
Closing — please re-open against [OpenIPC/builder](https://github.com/OpenIPC/builder) as outlined in the review above. Happy to look at it there.

PR #2152, opened 2026-05-27
kc110-mipi-shim: LD_PRELOAD workaround for majestic's broken MIPI ioctl construction (Hi3518EV200/V2A + OV2735)

moontwister:
## Summary

Adds a new package `kc110-mipi-shim` that builds `libfixmipi.so` — an `LD_PRELOAD` shim that intercepts `HI_MIPI_SET_DEV_ATTR` and `VIU_DEV_ATTR_SET` in `majestic` and patches the arguments in flight.

**This is a workaround, not a real fix.** The real fix belongs in `majestic` itself, which today builds `combo_dev_attr_t` with `raw_data_type=0` and `VI_DEV_ATTR_S.enIntfMode=DC` for MIPI sensors. `hi_mipi.ko` rejects both, and without something correcting the values majestic dies before any frame flows.

## The bug

On a Hi3518EV200/V2A board with an OV2735 MIPI sensor (TP-Link Kasa KC110), upstream majestic without any shim:

```
ioctl HI_MIPI_SET_DEV_ATTR failed
Cannot set VI Mipi attr
Cannot init sensor
Cannot start SDK
```

`/proc/driver/hi_mipi` reports the uninitialized defaults:

```
LaneNum   Vc0   Vc1   Vc2   Vc3    DataType    PhyMode
   4       0     0     0     0       N/A        CMOS
```

…instead of the correct `LaneNum=2, RAW10, MIPI` for OV2735.

## What the shim patches

| ioctl | Patch |
|---|---|
| `HI_MIPI_SET_DEV_ATTR` (0x41306d01) | `raw_data_type=2` (RAW10), `lane_id=[0, 1, -1, -1, -1, -1, -1, -1]` |
| `VIU_DEV_ATTR_SET` | `enIntfMode=VI_MODE_MIPI=5`, `au32CompMask[0]=0xfff00000` |

After interception, `/proc/driver/hi_mipi` shows the right values and majestic boots normally.

## Activation

The package installs `/usr/lib/libfixmipi.so` only. Activation requires `LD_PRELOAD=/usr/lib/libfixmipi.so` in the majestic init script (typically `/etc/init.d/S95majestic`). The package intentionally does **not** modify that init script — board adopters either patch their own `S95majestic` or carry a small overlay that does the export.

## Why upstream

This bug isn't theoretical for me — without this shim my KC110 won't start majestic. Bundling the shim as a Buildroot package makes it:

1. **Reviewable and versioned** — others looking at the same `HI_MIPI_SET_DEV_ATTR failed` line in their logs can pull a packaged workaround instead of rediscovering it.
2. **Easy to disable and delete** — `BR2_PACKAGE_KC110_MIPI_SHIM=n` and the LD_PRELOAD line goes away. The README has a removal checklist.
3. **A real fix candidate** — if a maintainer can sketch what the right majestic patch looks like, the symbol offsets the shim hooks (`HI_MIPI_SET_DEV_ATTR`, `VIU_DEV_ATTR_SET`) and the correct values for an OV2735 MIPI sensor are right here in source form, ready to be ported into majestic itself.

## Hardware tested

TP-Link Kasa KC110 (Hi3518EV200 + OV2735 MIPI 1080p). Live end-to-end:

- Without the shim → `Cannot start SDK`, no RTSP/JPEG.
- With the shim → RTSP and JPEG functional, sensor i2c writes land (post-#2149), AE converges (`mean RGB ~122/129/141`, normal exposure).

The shim's logic and Buildroot wiring are simple — `nm -D --defined-only` confirms it exports `ioctl` at the expected offset, and `strings` carries the two debug lines (`[libfixmipi] MIPI dev_attr: rdt=2 lanes=[0,1]`, `[libfixmipi] VIU dev_attr: intf=5 mask0=0xfff00000`) that fire once each at majestic startup.

## Naming

I kept `kc110-mipi-shim` to be honest about where it was developed and tested, but the failure mode is generic to Hi3518EV200 + any MIPI sensor going through OpenIPC majestic. Happy to rename to something more general (`hi3518ev200-mipi-shim`, `majestic-mipi-shim`, etc.) if reviewers prefer.

## Reference

Reverse-engineering write-up + companion `kc110-board-support` package (PR #2151): <https://github.com/moontwister/kasa-kc110-openipc>


flyrouter:
Good afternoon
Thank you for your interest in our project
Any use of LD_PRELOAD disrupts the logic of certain systems in the current OpenIPC+Majestic firmware. 
We will try to review the issue and resolve it internally as soon as possible.
If you enjoy working with source code, please take a look at our open streamers - Divinus, Mini, Venc, Waybeam, and perhaps you can implement your ideas there without hacks. 
Thank you for your understanding.

PR #2154, opened 2026-05-28
hisilicon-osdrv-hi3516cv200: add OV2735 case to load_hisilicon

moontwister:
## Summary

Adds OV2735 to the existing `ov2710_mipi` case in `cv200/load_hisilicon`. OV2735 is a MIPI sensor (1- or 2-lane, RAW10) used on several Hi3518EV200 cameras (TP-Link Kasa KC110, Lenovo Snowman variants, generic Foscam-class hardware). Its MIPI pinmux + SoC clock requirements are identical to OV2710_MIPI — only the libsns / .ini differ.

Without this case, setting \`fw_setenv sensor ov2735_i2c_1080p\` (or \`ov2735\`) on cv200 falls through to the \`xxxx Invalid sensor type xxxx\` branch, vendor modules don't load, and \`/proc/driver/hi_mipi\` never gets configured.

## Test plan

- [x] Built \`hi3518ev200_ultimate_tplink-kasa-kc110\` (OpenIPC/builder#99) with this patch applied. \`load_hisilicon\` enters the new case correctly, modules load, \`/proc/driver/hi_mipi\` shows the configured lane / data type.
- [x] cv300's \`load_hisilicon\` already handles OV2735 (combined with OV2718 in the same case) — this brings cv200 to parity. No behaviour change for existing sensors.

## Context

This is one of three small fixes split out from the closed #2151 (KC110 board package, re-filed as OpenIPC/builder#99 per @widgetii's review). Each lands a generic OV2735 cv200 fix without any board-specific scope.


flyrouter:
Good afternoon 
Thank you, we will review your proposals.

Issue #1930, opened 2026-01-04
Gk7205V200 Unable to operate normally os02g10_i2c_1080p


yooadmin:
> root@openipc-gk7205v200:~# majestic
> 
> 06:20:48  <       majestic> [    main] main@238                      Majestic Lite for HiSilicon, version master+094aaec, built on 2025-12-18
> 06:20:48  <       majestic> [watchdog] watchdog_start@94             Detected Watchdog
> 06:20:48  <       majestic> [watchdog] watchdog_start@115            Watchdog timeout set to 300 seconds
> 06:20:48  <       majestic> [   httpd] start_web_server@373          Set server path: /var/www
> 06:20:48  <       majestic> [   httpd] new_http_server@359           HTTP server started on :::80
> 06:20:48  <       majestic> [    main] init_sdk@66                   Loading main configuration: /etc/majestic.yaml
> 06:20:48  <       majestic> [     sdk] start_sdk@384                 Using /etc/sensors/os02g10_i2c_1080p.ini sensor configuration file
> 06:20:48  <       majestic> [     sdk] start_sdk@391                 App was built with MPP version: Gk7205V200_MPP_V1.0.1.0.B020 Release
> 06:20:48  <       majestic> [     sdk] start_sdk@394                 Current MPP version: VERSION=MPP_V1.0.1.0 B020 Release
> 06:20:48  <       majestic> [sensor_c] parse_sensor_config@961       Set custom sensor clock mode: 27MHz
> 06:20:48  <       majestic> [     sdk] start_sdk@422                 sensor stSnsOs02g10Obj
> 06:20:48  <       majestic> [     sdk] start_sdk@428                   input_mode MIPI, WDR NONE
> 06:20:48  <       majestic> [     sdk] start_sdk@435                   dev [1920x1080]@200x20 25fps, BGGR
> 06:20:48  <       majestic> [  sensor] try_to_load@19                trying to load /usr/lib/sensors/libsns_os02g10.so
> 06:20:49  <       majestic> [     hal] hisi_free_mem_ex@753          Free MMZ mem before allocation: 32736KB
> 06:20:49  <       majestic> [     sdk] dump_vb_configuration@2205    VB configuration:
> 06:20:49  <       majestic> [     sdk] dump_vb_configuration@2215      [0]: 3110400 x 3
> 06:20:49  <       majestic> [     sdk] dump_vb_configuration@2215      [1]: 2592000 x 1
[...]

Issue #1404, opened 2024-04-16
GK7205V210+sc223a (branded sc5239s) i2c not working


miguelangel-nubla:
Device [https://www.aliexpress.com/item/2251832668606248.html](https://www.aliexpress.com/item/2251832668606248.html) 4MP 48V POE, sensor branded as sc5239s which should be a sc223a.

Flashed with `burn` just fine, but the sensor is not detected on ipctool.

```
root@openipc-gk7205v210:~# dmesg
Booting Linux on physical CPU 0x0
Linux version 4.9.37 (runner@fv-az1501-923) (buildroot-gcc-13.2.0) #1 Tue Apr 16                                                                                               17:38:40 UTC 2024
CPU: ARMv7 Processor [410fc075] revision 5 (ARMv7), cr=10c53c7d
CPU: div instructions available: patching division code
CPU: PIPT / VIPT nonaliasing data cache, VIPT aliasing instruction cache
OF: fdt:Machine model: Goke GK7205V200 DEMO Board
cma zone is not set!
Memory policy: Data cache writeback
On node 0 totalpages: 8192
free_area_init_node: node 0, pgdat c05072d8, node_mem_map c1fbb000
  Normal zone: 64 pages used for memmap
  Normal zone: 0 pages reserved
  Normal zone: 8192 pages, LIFO batch:0
CPU: All CPU(s) started in SVC mode.
pcpu-alloc: s0 r0 d32768 u32768 alloc=1*32768
pcpu-alloc: [0] 0
Built 1 zonelists in Zone order, mobility grouping on.  Total pages: 8128
Kernel command line: mem=32M console=ttyAMA0,115200 panic=20 rootfstype=squashfs                                                                                               root=/dev/mtdblock3 init=/init mtdparts=sfc:256k(boot),64k(env),2048k(kernel),5                                                                                              120k(rootfs),-(rootfs_data) mmz_allocator=gk
PID hash table entries: 128 (order: -3, 512 bytes)
Dentry cache hash table entries: 4096 (order: 2, 16384 bytes)
Inode-cache hash table entries: 2048 (order: 1, 8192 bytes)
Memory: 27056K/32768K available (3856K kernel code, 150K rwdata, 912K rodata, 19                                                                                              2K init, 181K bss, 5712K reserved, 0K cma-reserved)
Virtual kernel memory layout:
    vector  : 0xffff0000 - 0xffff1000   (   4 kB)
[...]

PR #2206, opened 2026-06-24
goke: SC223A bit-bang i2c sensor support for gk7205v200 GPIO-wired boards

AKoo7:
## What

Adds an **opt-in** package, `goke-sns-bbi2c`, that brings up image sensors which are wired to **GPIO bit-bang i2c** on `gk7205v200` boards instead of the SoC hardware i2c controller.

Concretely it makes the **SmartSens SC223A** (chip-id `0xcb3e`) work on a **CP Plus CP-UNC-TA21L2C** (Dahua-OEM) camera, producing correct color 1080p video with the existing `libsns_sc223a.so`.

## Why

On some `gk7205v200`/`v210` cameras the sensor's i2c is **not** on any hardware i2c controller (`0x12060000/0x12061000/0x12062000`) — it is bit-banged over GPIO. So OpenIPC's hardware-i2c probe never gets an ACK and the sensor reads as *"not detected"* / `I2C_WRITE error`. This is the failure reported in:

- #1930 — gk7205v200 + os02g10, `I2C_WRITE error`
- #1404 — gk7205v210 + sc223a, "i2c not working"

Reverse-engineering the CP Plus unit (stock Dahua/Xiongmai-LiteOS fw) showed the sensor sits on a software i2c bus:

| signal | pin | pad / reg |
|---|---|---|
| SDA | GPIO7_2 | `0x112c0060` |
| SCL | GPIO7_3 | `0x112c0064` |
| sensor reset / enable | GPIO8_7 → high | `0x112c0048` |
| MCLK 27 MHz | CRG | `0x120100f0 = 0x19` |
| MIPI-RX ref clock | CRG | `0x120100f4 = 0x10340` |

Two things were needed beyond a normal sensor probe: (1) the sensor **reset/enable** line `GPIO8_7` must be driven high, and (2) the **MIPI-RX reference clock** `f4` must stay at `0x10340` while streaming — if it gets disturbed (e.g. by per-transfer clock re-init during majestic's AE writes) the PHY never locks (permanent ECC storm).

## How

- `goke_sns_bbi2c.ko` — registers a real Linux i2c adapter (`/dev/i2c-N`) that bit-bangs SDA=GPIO7_2 / SCL=GPIO7_3, drives `GPIO8_7` high, sets the 27 MHz MCLK, and **holds `f4` stable** (re-asserts MCLK + pads per transfer but never touches `f4`).
- `S96sensor-bbi2c` — loads the module, wakes the sensor, runs `load_goke -a -sensor sc223a -board demo`, remaps `/dev/i2c-0` (where `libsns` opens) to the bit-bang bus, and starts majestic. Replaces the generic majestic start for this board.

Opt-in via `BR2_PACKAGE_GOKE_SNS_BBI2C` (depends on `goke-osdrv-gk7205v200`); nothing changes for other boards.

## Testing

Verified on hardware (CP Plus CP-UNC-TA21L2C, SC223A):
- `MIPI DETECT 1920x1080`, **all CRC/ECC/CIL error counters 0**.
- majestic streams correct, color-balanced 1080p H.264 + MJPEG; RTSP + HTTP snapshot OK.
- The module compiles cleanly against the `gk7205v200` 4.9.37 kernel + OpenIPC toolchain (vermagic `4.9.37 mod_unload ARMv7 p2v8`).

## Notes / open questions for maintainers

- The sensor (SC223A) was already supported by `libsns_sc223a.so`; this PR only adds the missing **board glue**.
- GPIO/CRG addresses are `gk7205v200`-specific. A device-tree `i2c-gpio` adapter would be the more idiomatic long-term form, but this vendor 4.9 kernel has no DT for these pins and `CONFIG_I2C_GPIO` is not enabled — hence a small self-registering module. Happy to rework toward `i2c-gpio` / a cleaner integration (e.g. how best to slot the bring-up + `/dev/i2c-0` remap into the standard sensor flow) if preferred.
- Marked **draft** pending that feedback.


flyrouter:
Good afternoon,
 thank you very much for your interest in the project and the work you have done, it's a very interesting solution. 
 We will review your proposals soon and plan our next steps together. 
 Thanks.

widgetii:
## Correction: these pads are the SoC's hardware I²C0, not GPIO-only

Going back over this against the gk7205v200 pinmux table, the premise in the description doesn't hold up and I want to correct it.

The SDA/SCL pads this module bit-bangs — `GPIO7_2` / `GPIO7_3` (IOCFG_VIO `0x112c0060` / `0x112c0064`) — are the **hardware I²C0** pads in their function-2 mux. The OpenIPC SDK does exactly this routing itself, in [`openhisilicon` `kernel/sys_config/sys_config.c`](https://github.com/OpenIPC/openhisilicon/blob/master/kernel/sys_config/sys_config.c):

```c
void i2c0_for_vi_raw_sensor_pin_mux(void)
{
    SYS_WRITEL(reg_iocfg_vio_base + 0x0060, 0x1002);  // 0x60 -> func2 = I2C0
    SYS_WRITEL(reg_iocfg_vio_base + 0x0064, 0x1002);  // 0x64 -> func2 = I2C0
}
```

So the pad map is `func0 = GPIO7_2/7_3`, `func2 = I2C0_SDA/SCL`. That means the stated root cause — *"the sensor i2c is not on any hardware i2c controller, it's bit-banged over GPIO"* — is wrong, and the reverse-engineering detail about the stock firmware using a software i2c bus is not something I can substantiate; the pinmux contradicts it. Please disregard that part of the write-up.

The more likely reason the hardware probe never ACKs: OpenIPC's stock bring-up muxes I²C0 onto the **default** mipi-sensor routing (`0x30/0x34`, func1) and probes there, but this board wires the SC223A to the **vi-raw** routing (`0x60/0x64`, func2). The controller transactions simply come out the wrong pins.

If that's it, the fix is a pad-mux + the existing hardware I²C0 controller — no kernel module, no `/dev/i2c-0` remap:

```sh
devmem 0x112c0060 32 0x1002   # SDA -> I2C0 (func2)
devmem 0x112c0064 32 0x1002   # SCL -> I2C0 (func2)
# read SC223A chip-id (reg 0x3107 -> 0xcb) over the hardware controller on bus 0:
i2ctransfer -y 0 w2@0x30 0x31 0x07 r1
```

If that returns `0xcb`, the whole `goke_sns_bbi2c.ko` collapses into a couple of pinmux writes in the sensor bring-up. The two non-i2c things the module also does — driving `GPIO8_7` (reset/enable) high and setting MCLK / holding the MIPI-RX `f4` clock stable — belong in the init script either way; and the per-transfer "re-assert clock + pads" dance that `f4` required is an artifact of the bit-bang path (with the hw controller nothing touches clocks per transfer, so that failure mode should disappear).

The hardware result stands — the camera streams — so this isn't disputing that it works, only whether a board-specific kernel module is the right vehicle when the pins already speak hardware i²c. I'll run the `i2cget`/`i2ctransfer` path on the unit and report back; keeping this as draft until then.

AKoo7:
@widgetii You're right on both counts — tested on the actual unit (CP Plus CP-UNC-TA21L2C, gk7205v200, SC223A `0xcb3e`). Thanks for the careful read.

**1. The pads are hardware I²C0 (func2), and the SC223A ACKs there.**

With GPIO8_7 (the sensor reset/enable line) driven high and MCLK `0x120100F0=0x19`, your pinmux works exactly as you said:

```sh
devmem 0x112c0060 32 0x1002    # SDA -> I2C0 (func2)
devmem 0x112c0064 32 0x1002    # SCL -> I2C0 (func2)
# also de-muxed the default routing 0x112c0030/0x34 from func1 -> GPIO (0x1000)
# so both pin-pairs aren't on I2C0 at once

i2cdetect  -y -r 0                       # -> 0x30 present (nothing on bus 1/2)
i2ctransfer -y 0 w2@0x30 0x31 0x07 r1    # -> 0xcb
i2ctransfer -y 0 w2@0x30 0x31 0x08 r1    # -> 0x3e   => 0xcb3e = SC223A
```

No bit-bang, no `/dev/i2c-0` remap. My earlier "HW I²C0 doesn't ACK" attempts were a false negative: they predated my finding that GPIO8_7 is the reset/enable line, so the sensor was simply held in reset and nothing answered on *any* bus — which I misread as "the controller must be wired to the wrong pins / it must be software i²c." With reset released, the hardware controller on func2 is fine.

**2. The whole module is removable — including the per-transfer clock handling.**

Removed `goke_sns_bbi2c` entirely and brought the sensor up over the stock `open_sensor_i2c` path on bus 0:

```sh
rmmod goke_sns_bbi2c
load_goke -a -sensor sc223a -board demo
# 0x60/0x64 -> func2, de-mux 0x30/0x34, GPIO8_7 high, MCLK 0x120100F0=0x19
SENSOR=sc223a majestic
```

Result: `MIPI DETECT 1920x1080`, error counters clean, RTSP up, correct live 1080p — **and `0x120100F4` stays at the streaming value `0x10340` the whole time, never clobbered.** That confirms your other point: the per-transfer "re-assert clock + pads" dance was purely an artifact of the bit-bang module re-running its setup on every transfer; with the hardware controller nothing touches the clock per transfer, so the MIPI-lock failure mode disappears on its own.

So I agree the kernel module is the wrong vehicle. I'll rework the PR to drop `goke_sns_bbi2c.ko` and instead:
- route I²C0 to the vi-raw pads (`0x60/0x64` func2) — i.e. your `i2c0_for_vi_raw_sensor_pin_mux()`;
- drive GPIO8_7 (sensor reset/enable) high + set MCLK;
- let the stock `open_sensor_i2c` drive the SC223A on bus 0.

**One question on placement:** what's the right home in OpenIPC for this board-specific pinmux + reset GPIO — a new `load_goke -board` profile, a sensor-config entry, or a small bring-up snippet? I'd rather put it where you'd want it than guess; happy to push the reworked branch once I know.

**Footnote on the stock firmware (not a dispute):** I'd traced the *stock* LiteOS sensor writes to a software bit-bang dispatch, which is what sent me down the GPIO path — but vendor SDKs do sometimes bit-bang pins that also expose a hardware controller, so that observation doesn't contradict the pinmux, and it's moot for OpenIPC since the HW path clearly works. I'll correct the PR description accordingly.

widgetii:
Right place for board-specific pinmux/drivers/etc is https://github.com/OpenIPC/builder repo

PR #2226, opened 2026-07-22
[no ci] Update openipc_donors with support information

flyrouter:
Added note about device support and seller contact.


flyrouter (commented):
Done

Issue #37, opened 2021-08-13
Sigmastar kernel sources


fifteenhex:
Hi guys,

I'm trying to archive as many of the sigmastar kernel releases as possible here:
https://github.com/linux-chenxing/linux-ssc325
I think I saw on your SDK archive that you have some versions I don't but it looks like it's non-public now. Would you mind putting the kernel source tarballs somewhere for me?

As there are no public datasheets with registers they are the only way to find things like the hardware irq numbers without lots of painful reverse engineering.

ZigFisher:
Dear friend

We have been seeing your work for a long time and are amazed by it. Thank you very much !
Of course, we are ready to contribute to your research. However, for faster communication, do you have the opportunity to use the Telegram messenger? If there is such a possibility (I really hope) then find me there as @FlyRouter.
Thank you very much and see you soon.

PR #2234, opened 2026-07-28
Gk7201v200 support

oskarirauta:
# Goke GK7201V200: make the existing board produce a working camera

`gk7201v200_lite` is already in the tree, but it does not yield a usable image: majestic is
not built and the module set it would run does not drive this die. This series makes the board
work end to end, and adds it to the CI matrix so its images actually reach the releases.

Hardware this was developed and verified on: a XiongMai GK7201V200 (vendor chip id
`xm72010200`) camera with an SC2336 sensor and an RTL8733BU radio.

## Why the existing configuration cannot work

The `goke-osdrv-gk7205v500` module set is built **Apr 3 2023** and matches the Apr-2023
`libxmedia_*.so` userspace beside it. On the GK7201V200 die it brings the SoC up as far as the
ISP and then stops: `VEDU CreateChn` fails with `F008FFFF` and the VI→VPSS online link never
starts, so there is no picture at all.

The camera's factory firmware carries a **Sep 23 2023** build of the same vendor MPP, for chip
id `xm72010200`, which drives this silicon correctly. Both declare
`vermagic=4.9.37 ARMv7 thumb2 p2v8`, so they load on the same kernel — they are two builds of
one vendor stack, five months apart. That five-month skew against the Apr-2023 userspace is
real, and bridging it is what the new `gk7201-fixup` package does.

## What works, verified on hardware

Measured on the built image, majestic started by its own init script, no environment overrides:

- **Three encoder channels at once**, in majestic's own `VI_ONLINE_VPSS_ONLINE` mode. Frames per
  15-second window: main 1080p **20.0 fps**, substream 640x360 **14.9 fps**, MJPEG 1080p
  **5.0 fps**, with `VENC Send == VPSS SendOk` on all three and zero `Timeout from venc`.
- **RTSP main stream and substream**, `/image.jpg` snapshots, MJPEG over RTSP (`stream=2`, after
  majestic's own `.jpeg.rtsp` opt-in).
- **Microphone audio in the stream.** The vendor audio path is loaded last, exactly as the vendor
  does it (`xm_aio`, `xm_ai`, `xm_ao`, `xm_aenc`, `xm_adec`, `xm_acodec`), and needs the nine
  `libvqe_*.so` this series installs for the board — `libxmedia_api.so` resolves them at runtime
  and the audio path does not come up without them.
- **The SD card slot works**: the card is detected, mounts at `/mnt/mmcblk0p1` and is read/write.
  It carried every image, log and test artefact through this whole port, so it is well exercised
  in general use. I have not driven majestic's own recording to it, so I am not claiming that
  specific path.
- **AE/exposure over the full range**, and IQ/CCM/AWB giving vendor-like colour.
- **Automatic day/night** off the sensor's analog gain.
- **Wifi (RTL8733BU)**: associates, DHCP + SLAAC, carries traffic, survives a cold boot.
- **Hardware watchdog**, fed by majestic, enable and disable both effective without a reboot.
- **`dmesg | grep -ci oops` = 0**, including across a majestic restart and a settings change with
  SIGHUP — both of which used to leave this camera dead until a power cycle.

Sizes: uImage 1961/2048 KB, rootfs 6052/8192 KB.

## The commits

1. `linux: describe the GK7201V200 watchdog as the SP805 it is` — the only kernel patch.
2. `goke-osdrv-gk7205v500: carry the GK7201V200 vendor MPP modules` — the Sep-2023 module set,
   with provenance.
3. `goke-osdrv-gk7205v500: add the libhi_* -> libxmedia facades` — plus the script that
   generates them.
4. `goke-osdrv-gk7205v500: bring the GK7201V200 up with load_xm` — the load sequence and the
   board-scoped install hook.
5. `gk7201-fixup: new package bridging the GK7201V200 API skew` — kernel module + LD_PRELOAD
   shim.
6. `majestic: let a board opt into a hard stop` — one opt-in variable, no change for anyone else.
7. `wireless: add the rtl8733bu-gk7201v200-xm profile`.
8. `gk7201v200: make the existing defconfig produce a working camera`.
9. `ci: build gk7201v200_lite`.

## Everything here that touches shared code, declared upfront

I would rather point these out than have a reviewer find them.

- **`goke-osdrv-gk7205v500.mk` is a purely additive change** (98 lines, zero deletions). All
  board-specific installs sit in a `POST_INSTALL_TARGET_HOOKS` block guarded by
  `ifeq ($(BR2_OPENIPC_SOC_MODEL),"gk7201v200")`, so the existing install block is byte-identical
  and `gk7205v500_lite`, `gk7205v500_ultimate` and `gk7205v510_lite` are unaffected. The vendor
  modules live in their own `files/kmod.gk7201v200/` directory; `files/kmod/` is untouched.
- **`load_goke`** gains one guarded `exec load_xm` when `ipcinfo` reports this chip. One branch
  rather than a second variant threaded through the whole script.
- **`majestic/files/S95majestic`** gains `MAJESTIC_HARD_STOP`. With it unset — every board that
  does not ship an `/etc/default/majestic`, which is all of them today — the script behaves
  exactly as before. See the commit for why this board needs it.
- **`xm72050200.dtsi` is shared with gk7205v200.** Nothing binds to the old watchdog node by
  device tree: the vendor `xm_wdt` uses DT only for `devm_ioremap_resource`, and
  openhisilicon's `kernel/wdt/wdt.c` (what `open_wdt` is built from) uses a hardcoded
  `#define WDT_BASE` with no compatible string at all. So no other board loses anything.
- **`files/sensor/iq/sc2336p.bin` physically lands in `hisilicon-osdrv-hi3516ev200`.** That is
  not a mistake: that directory is OpenIPC's shared sensor-IQ pool, and both
  `goke-osdrv-gk7205v200` and `goke-osdrv-gk7205v500` carry `files/sensor/iq` as a tracked
  symlink into it. Nothing else starts shipping the file — that package installs its IQ files by
  name, not by glob.
- **`files/sensor/config/sc2336_i2c_1080p.ini` does reach the other gk7205v500 boards**, because
  it is picked up by the package's existing `config/*.ini` glob. It is a sensor config file and I
  judged that harmless and idiomatic, but it is a real effect and I would rather say so.

## Binary blobs, and their provenance

The vendor kernel modules are binaries taken from the camera's own factory firmware. No source
exists in public. `files/kmod.gk7201v200/PROVENANCE.md` records, so the extraction can be
repeated and audited independently:

- the source device and method (a full 16 MiB NOR dump read off the camera, with its md5),
- the path inside that image (`/mnt/mtd/ipc/modules/`) and the vendor's own load invocation,
- a per-file md5 table for all 43 files,
- and the licensing situation stated plainly: every module declares `license=GPL` while the
  vendor publishes no source. That is the same situation as the module set this package already
  carries; these are redistributed on the same footing.

If you would rather not carry a second binary set in-tree, say so — a build-time fetch, or an
extraction script pointed at the user's own dump, is a reshape rather than a rewrite, and the
files are listed individually in that document precisely so the decision can be informed.

The six `libhi_*.so` facades are generated, not hand-written, and `gen_xmedia_shim.sh` is
committed with them. Each stub is a single tail branch (`b XMEDIA_API_VENC_CreateChn`), which
touches neither `lr`, `sp` nor any argument register — ABI-transparent without knowing a single
signature. OpenIPC already ships exactly this shape for the GK stack (`libhi_mpi.so` in front of
`libgk_api.so`); this builds the same six facades against the XMEDIA implementation instead.

## One kernel patch, and two I deliberately dropped

Only the SP805 device-tree fix is here. The block at `0x12030000` is a stock ARM PrimeCell
SP805 — the silicon answers PeriphID `0x00141805` and PCellID `0xB105F00D`, and that PeriphID is
exactly the `amba_id` `sp805_wdt.c` matches on. Describing it correctly lets the in-tree driver
bind, with no vendor blob involved. (The vendor `xm_wdt` cannot be used regardless: it hardcodes
`movs r0,#37` at all four of its IRQ call sites instead of reading its own DT node, so the dog is
never fed and the board resets every ~60 s. This tree's own `load_goke` already carries
`# modprobe xm_wdt # temporarily disabled` for the same reason.)

Two patches used during development are **not** included, and one of them guards a real hazard
that is worth knowing about:

- Resolving a module symbol inside a vendor `xm_*.ko` faults (NULL `strtab`), and because that
  lookup happens inside the Oops printer, one fault becomes a recursive fault storm that wedges
  the board instead of printing a trace. A local patch that prints raw addresses in
  `dump_backtrace_entry` breaks the recursion. It is not in this series because it would degrade
  Oops output for every ARM board in the tree, and because the hardware watchdog added here now
  covers the wedge case: majestic stops feeding the SP805 and the board resets. Anyone debugging
  a vendor-module Oops on this SoC should apply it locally.
- An ARM kprobes fix (`reset_current_kprobe()` on every path, not only when the pre-handler
  returns 0) is likewise omitted: it only matters for handlers that return non-zero, and every
  handler this series ships returns 0.

## Known limitations, stated plainly

- **`image.rotate` is not supported on this SoC, and setting it costs you the video.** The vendor
  kernel refuses it — `HiSi_HAL_SetVPSSRotate: ERR_VPSS_NOT_SUPPORT` — and majestic then aborts
  `init_chn` for *every* channel, so a user who sets rotation in the web UI gets a camera with no
  video until they set it back. Verified identically in both VI-VPSS modes. Nothing in this
  series can reach it: majestic swaps the geometry itself before the SDK call. 180° works via
  `mirror` + `flip`, which majestic applies at sensor level.
- **IR-cut ships but is not verifiable on my unit: it is not fitted.** No wires leave the lens
  holder at all. The pin mapping is in the tree (`nightMode.irCutPin1 13` / `irCutPin2 15`,
  derived from the vendor `ircut.ko`, which is a dumb H-bridge pin toggler with all pulse timing
  in userspace) and the pad mux is set by `load_xm`, but I cannot claim it works.
- Also not fitted on this unit, so likewise unverified: the light sensor / LSADC, the speaker,
  and the illuminator LEDs.
- **The wifi profile is validated on one unit.** The vendor's own `load_drv.sh` probes four
  different radios, which proves cameras with this SoC ship with different modules — hence a
  named profile (`rtl8733bu-gk7201v200-xm`) rather than anything automatic. GPIO 9 is the enable
  and it is active low; without pulling it the chip never appears on the USB bus at all.
- I have exactly one camera of this model, so nothing here is cross-checked against a second
  unit.

## Board GPIO map, read out of the vendor image

The factory firmware carries one small `extdrv/*.ko` per peripheral, each of which is a plain pin
toggler. Disassembling `init_module` gives the pin directly from its `gpio_request(N)`, so this
map is read out of the vendor's own drivers rather than guessed or probed. Nothing in this series
depends on it beyond the two pins marked as used; it is here because it is the piece of knowledge
a second person working on this SoC would otherwise have to re-derive.

| GPIO | vendor driver | function | how it is known |
|------|---------------|----------|-----------------|
| 4 | `rled.ko` | red status LED | from the driver. `load_drv.sh` runs `rledctrl 0` at boot when an analog light sensor is fitted — you switch a status LED off so it cannot pollute the sensor, which is not something you would do to an illuminator |
| 9 | `wifien.ko` | **wifi enable, active low** | from the driver, **confirmed on hardware** — without pulling it the radio never appears on the USB bus. Used by this series |
| 12 | `audioout.ko` | audio-out enable (speaker amplifier) | from the driver; not fitted on my unit |
| 13, 15 | `ircut.ko` | IR-cut H-bridge, both high = idle | from the driver; the mechanism is not fitted on my unit, so not verifiable here |
| 14 | `wled.ko` | white LED / floodlight | from the driver; not fitted on my unit |
| 20, 45 | `sinit.ko` | sensor init | from the driver; driven at boot by `load_xm` |
| 50 | `reset.ko` | **reset button, active low** | from the driver, **measured on hardware**: idle 1, pressed 0. See the next section |
| 55 | `rs485.ko` | RS485 direction | from the driver; not fitted on my unit |
| 69 | `relay.ko` | **relay switching the IR illuminator panel** | from the driver, and the panel wiring per the board owner. The name is accurate rather than misleading: a SoC pin cannot source the current an IR LED panel draws, so the GPIO drives a relay and the panel hangs off that. This means majestic's `backlightPin` is **69, not 4** |

Only three pads are muxed by the vendor at boot — `0x120C0000` (ADC), `0x120C0010` (audio out),
`0x120C0014` (IR-cut) — and `load_xm` writes exactly those three. Everything else is assumed to
be GPIO at reset; if a pin does not respond, suspect the mux first.

An empty connector on the PCB proves nothing about a given camera: this firmware and board serve
several models, and the vendor's `load_drv.sh` loads `relay`, `rs485`, `extalarm` and `motorpt`
unconditionally although this unit has none of them.

## The reset button

The button is on **GPIO 50 and is active low** — measured on hardware: idle reads 1, pressed
reads 0. It needs no configuration at all and is usable exactly as this image ships it. The pad
mux is already correct at reset (proven by the measurement itself: a pad muxed elsewhere would
read a constant, this one tracked the press), the PL061's `GPIODIR` resets to input and the
driver does not force a direction at probe, and nothing claims the pin because the vendor
`reset.ko` is never loaded. So:

```
echo 50 > /sys/class/gpio/export
echo in > /sys/class/gpio/gpio50/direction
cat /sys/class/gpio/gpio50/value
```

Do **not** use the `gpio` helper on this pin: its export path writes `out` to `direction`, and
driving an output into a switch that shorts to ground is exactly what you must not do here.

Nothing is shipped to act on the button, deliberately. The only button handler in the tree is
`general/scripts/legacy/sbin/resetd` — legacy, installed by no package, with its GPIO unset,
polling four times a second (its own comment estimates ~1% CPU) and running `firstboot` after a
~5 s hold. Its polarity assumption does match what I measured, so anyone who wants it can point
its `GPIO=` at 50 and enable it in one line. Reviving a deprecated polling daemon for one board
seemed the wrong trade; documenting the pin costs nothing and leaves the choice with the user.
A `gpio-keys` node would only *protect* the pin (claiming it makes the sysfs export return
`EBUSY`) and would need `INPUT_KEYBOARD` + `KEYBOARD_GPIO` + `INPUT_EVDEV` — evdev is off — plus
another patch to the shared DTS.

## Installation notes

`defib` installs the bootloader for this chip via the `gk7205v200` profile, which is correct —
no gk7201v200 u-boot exists — but it then writes `soc` from the profile instead of from its `-c`
argument. Three values have to be set by hand before the first boot, because `uknor16m` /
`urnor16m` fetch `uImage.${soc}` and `rootfs.squashfs.${soc}`:

```
setenv soc gk7201v200
setenv sensor sc2336
setenv osmem 32M
saveenv
```

Do not touch `bootargs`, `bootcmd`, `mtdparts`, `baseaddr`, `totalmem` or `ethaddr` — this
board's stock values are correct, and `bootcmd` expands `bootargs` through `setargs`, so nothing
needs hardcoding.

**Ethernet works, and it works at the same time as wifi** — `S40network` brings eth0 up on the
wireless branch too, so the wired port stays available as a recovery path with wifi running.

Wifi needs `wlanssid`, `wlanpass`, `netaddr_fallback` and finally `wlandev` — **in that order**,
and the reason is worth spelling out because it is easy to lock yourself out. `S40network` keys
everything off `wlandev`: with it unset the script simply runs `ifup eth0`, i.e. DHCP; with it
set it takes the wireless branch, which ends with

    ifconfig eth0 "${net:-192.168.2.10}" hw ether "${eth:-00:00:23:34:45:66}" up

So the moment `wlandev` is set, **eth0 stops using DHCP**. If `netaddr_fallback` is unset it
lands on the built-in `192.168.2.10`, and since `ifconfig` is given no netmask the kernel applies
the classful default `255.255.255.0` — which is very likely not your subnet, and the camera
becomes hard to reach. Set `netaddr_fallback` to an address on your own network first, and
`wlandev` last.

There is a second route that needs no u-boot access at all: with the camera on ethernet and
reachable, configure wifi from the web UI and **finish it before rebooting**. The settings are
persisted once `wlandev` is set, so the next boot comes up on wifi. What does not work is
configuring wifi purely through the web page and expecting it to survive on its own: nothing in
the tree ever sets `wlandev`, so the save succeeds, everything looks correct, and there is no
wifi after a reboot because the wireless branch is never taken. That is pre-existing behaviour,
not something this series changes.

The factory bootloader is password-protected. The password on this model's vendor image is
`HI2105CHIP`; without it there is no console access to a stock camera at all.

## A separate suggestion, not part of this PR

The rotation failure above is arguably a majestic robustness bug rather than a platform gap: one
`ERR_VPSS_NOT_SUPPORT` from an *optional* image transform currently aborts channel creation for
every channel, turning a cosmetic setting into a total video outage. A camera that ignores a
rotation it cannot perform is strictly better than a camera with no video. A per-board way to
declare a feature unsupported — so the setting is ignored and the web UI can grey the control
out — would be nicer still, but the non-fatal-failure change stands on its own. Happy to file
that separately if it is wanted.

## Testing

Every claim above was measured on the assembled camera, not inferred, and re-verified on the
final image — kernel and rootfs both flashed with `sysupgrade`, majestic started by its own init
script, no environment overrides.

Verified after that flash: all three VPSS channels with their counters climbing across repeated
samples; both RTSP streams, `/image.jpg` and MJPEG over RTSP in a player and a browser; audio;
night mode; wifi associating and carrying traffic; and the SD card mounted read/write.
`S95majestic stop`, `start` and `restart` all behave, with the watchdog node reading `inactive`
after a stop — the direct proof that the daemon's watchdog is released rather than left armed —
and `active` again once majestic is back. Feeding was confirmed with a 60 s timeout, where
`timeleft` bounced back up three times in 90 s; a window shorter than half the timeout cannot
show this and proves nothing either way.

`dmesg | grep -ciE 'oops|kernel panic'` stayed at 0 throughout, including across the stop/start
cycles and a settings change with SIGHUP — both of which used to leave this camera dead until a
power cycle.

The tree was also built from scratch (`make clean`) from the exact commits in this series, and
the resulting rootfs compared file by file against the incrementally built one, to be sure the
image being described is the image the tree produces.


widgetii:
I’m closing this AI-generated content. If you’d like to engage in genuine work, begin by addressing the actual vendor SDK.

oskarirauta:
Like PR said, vendor SDK is not available - modules are available from vendor image which is provided in my repository on branch goke7201v200-vendor - retrieved from camera through uboot. Vendor SDK is commercially available for manufacturers and there is now way for me to get that. I also requested SDK from OpenIPC through a issue some time ago to work on that - and gave up on that due to not getting any replies and OpenIPC did not share SDK with me either so I see no point why SDK would be required from me. It never existed and since I am incapable of complex disassembly or reading complex and long assembly language - I needed help from someone, and in this case, I chose AI. It was a big task and in the end it produced a fully working image (besides rotation support, but it didn't exist in the first place).

This allowed full support for device that isn't supported previously by OpenIPC, so I think this judgement is unfair. Also, I am not actually requesting merging of my PR, I am requesting support for this hardware - and PR that I here presented shows clearly how it is possible. Do it this way, or do it your way; instructions are here, and they are thorough, mine wouldn't have this much commentary or explanations ever, so in view of developer (or reviewer), this work should give a huge head start.

I do know that work that I have presented serves the audience and society; I would had been more than glad if I would just had a chance to "flash away" a fully working image without requiring to pay for AI assistance, or even do the tinkering; but for that to exist - someone has to do it. Now it's done. I hope it will carry and "real support" will be available eventually. Or is this project dead or locked to selected platforms?

Usually platform support begins from small steps here and there. This time you got your hands on to fully functional image generator on your own soil. I am not sure what more you would be requiring for member outside of community to share with you? Well - I've tried, gift without asking anything in return, wasn't accepted and instead got returned.

I can't hope more than that someone someday finds this useful - as I surely did. Not getting back at this anymore as this was closed. I will though still provide some images of hardware to here. And pre-built binaries. PR has all necessary information for flashing if some daredevil's appear who rather choose pre-built instead of self-built. Makes things a lot easier for those who look a fast way to deploy openipc to their cameras. Provided binaries are running on the camera as we speak as-is and were flashed from tftp server.

[gk7201v200-openipc-firmware.zip](https://github.com/user-attachments/files/30458048/gk7201v200-openipc-firmware.zip)

References:
 - https://github.com/oskarirauta/firmware/tree/gk7201v200-support
 - https://github.com/oskarirauta/firmware/tree/goke7201v200-vendor

IMG_6439
image

oskarirauta:
Oh, on more thing - in the image, serial connection is wired, RX and TX are side-by-side, first one if RX (green wire) and second is TX (blue wire), rest of connections are GND and VCC but I cannot remember in which order, I grounded it from circuits screw hole.

widgetii:
There’s no way to create a functional streamer of any kind using extracted blobs from vendor firmware and support it in the long run.

Your approach is only suitable for creating IPC camera mods, but it’s not the objective of this project.

oskarirauta:
This doesn't belong to this subject, but since I don't have other means to contact you..
I'll delete this comment after a while, to keep subject in precision.

I was wondering if you'd be interested in support for Ingenic t41(nq)?
I have a working uboot for it....... Though, it's thingino baked - but still.. t41nq platform requires some registry settings to enable devices such as sd, wifi and sensor.. My email address is listed in my GitHub profile if you think this information could be useful...

(I also have openipc running on t41nq, but so far it is just linux.. instead of _openipc_..)

widgetii:
We don't have this hardware in the lab now

flyrouter:
@oskarirauta 

Good afternoon. 
Thank you for your interest in our project. 

We value and respect the contributions of all community members, but your contribution contains components that we cannot accept immediately. The thing is, components collected for one board may not work at all or may not work correctly on another board, and we've experienced this before. 

So please be patient, and most likely the information, build components and descriptions you provided will be useful for analysis by our team when this processor is re-integrated according to our norms and standards. 

Please check out our OpenHiSilicon repository, which has been created according to the standards we strive for.

Regarding your inquiries about the SDK, as you probably guessed from our conversation, we currently don't have it in full. I might have replied to your messages on behalf of OpenIPC myself, so everything's fine.

And finally, just because a commit isn’t accepted doesn’t mean others can’t use it for tests, understanding the experimenter’s ideas, repeating the experiment, and coming up with truly proper and acceptable solutions. So, in any case, thanks to you, you did what you could, it’s valuable and better than doing nothing like others do.

Pay attention to your IR filter where the wires don’t go. If it has a coil for switching slides, sometimes the power is supplied not through wires but through mechanical contact with electrical conduction through the mounting points with screws. We’ve already come across such modifications.

We plan to support the Ingenic T41 processor in the near future as soon as we gather all the necessary information and get some sample equipment.

oskarirauta:
> @oskarirauta
> 
> Good afternoon. Thank you for your interest in our project.
> 
>  ....
> 
> We plan to support the Ingenic T41 processor in the near future as soon as we gather all the necessary information and get some sample equipment.

Good morning @flyrouter.

Oh - I already have t41 openipc firmware with u-boot and all (builds from openipc tree and also generates _full image_)...
https://github.com/oskarirauta/firmware/tree/t41nq-support

Your problem with t41 uboot is that you are using older uboot that doesn't support t41 easily, you need patched experimental 2026.04 version - with patches that set proper registries to wake up wifi and sd. I have a testing phase going on with 4 cameras running this image. Didn't just case to make a PR out of it since outcome quite surely is the same, majestic and libimpl has serious problems with t41 where thingino was able to outcome it, so it needed some assistance similar to what goke needed so surely it's a no go. But here's ready built images with all the bells and whistles, such as uboot and full flash image included, or you can self-build.

[openipc-t41.zip](https://github.com/user-attachments/files/30579448/openipc-t41.zip)

oskarirauta:
Oh, and here are development steps taken that were necessary to produce that uboot..

Thingino already had a uboot for newer ingenic; it wasn't tailored for t41 though - but allowed boot to thingino's firmware, nothing worked though - such as sensor, wifi, sd, etc..

So then begun the reverse engineering process from vendor's uboot and necessary registry changes needed to enable hardware were eventually found. And now, I incorporated that work to openipc and "tailored" a openipc version (includes uknor16m, etc by default, prompts openipc, etc..) - so it's a fully functional uboot and ready to go, for Ingenic's t41. With device specific registry changes.. for other boards probably as well.

Source available from my repo for free if interest in it is found. T41 boards are available pretty cheap, my cheapest camera running t41 probably cost me 28eur home delivered- and I mostly bought it for development purposes, didn't know doc in advance though. But since it was fairly cheap with metal casing....

PR #2242, opened 2026-08-07
add wireguard watchdog

Puzzak01:


flyrouter (changes requested):
Neurotard or a dumb fool.

PR #2244, opened 2026-08-09
majestic: re-read /etc/TZ on every start so restarts pick up TZ changes

phedoreanu:
## What

`S95majestic` re-reads `/etc/TZ` and exports `TZ` before starting the daemon, guarded by `[ -r /etc/TZ ]`.

## Why

`rcS` exports `TZ` once at boot, and every descendant inherits that snapshot. Restarting majestic through one of its own children — the web UI, or `sysupgrade` (which calls `S95majestic restart`) — hands the daemon the stale boot-time value, which shadows any later edit of `/etc/TZ`. The OSD clock then keeps rendering the old zone until a full reboot, or until a restart from a fresh SSH login (which re-sources `/etc/profile`).

Re-reading the file in the init script makes every start pick up the current `/etc/TZ`, so a timezone change takes effect on `S95majestic restart` without a reboot.

The `[ -r ]` guard matters on its own: on musl an *empty* `TZ` pins UTC and skips the `/etc/localtime` fallback that an *unset* `TZ` would use, so exporting unguarded when `/etc/TZ` is missing would force UTC. (`rcS:4` and `profile:6` currently have this latent footgun; happy to port the guard there as a follow-up.)

*The original description blamed `/etc/profile` not being sourced by init; @widgetii correctly pointed out that `rcS` already exports `TZ` on the boot path, and the real defect is the staleness above. Description and commit message rewritten accordingly.*

## Verification

On a Hi3518EV200 camera with TZ `CET-1CEST,M3.5.0,M10.5.0/3`: the OSD clock follows local time including DST, and `tr '\0' '\n' < /proc/$(pidof majestic)/environ` shows `TZ` set. Staleness-path checks from the review are reported in the PR comments.

### DIFF
--- a/general/package/majestic/files/S95majestic
+++ b/general/package/majestic/files/S95majestic
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+# rcS exports TZ once at boot, so a later edit to /etc/TZ is shadowed by the
+# stale value inherited through majestic's own children (web UI, sysupgrade).
+# Re-read it here so a restart picks up the new zone without a reboot.
+# musl never reads /etc/TZ itself; it uses $TZ, else /etc/localtime.
+[ -r /etc/TZ ] && export TZ=$(cat /etc/TZ)
+
 DAEMON="majestic"
 PIDFILE="/var/run/$DAEMON.pid"
 DAEMON_ARGS="-s"


flyrouter (commented):
Thanks. 
It needs further checking. 
No timing issues have been noticed on the OSD in the lab.

general/package/majestic/files/S95majestic
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# majestic is started by init, which does not source /etc/profile, so the TZ
+# export in there never reaches it and the OSD clock renders UTC. musl ignores
+# /etc/TZ on its own; this export is what makes the timezone take effect.
```
widgetii: This comment states a root cause that the tree contradicts, and it's the part I'd most like changed before merge.

`rcS` is what init actually runs, and it exports `TZ` on line 4 *before* looping over `/etc/init.d/S??*` and running each one with `$i start`. `S95majestic` is installed into `/etc/init.d` by `majestic.mk:27`, so it's a direct child of `rcS` and does inherit `TZ`. `/etc/profile` never enters the boot picture — it isn't the reason the OSD showed UTC.

The accurate cause is *staleness*: `rcS` reads `/etc/TZ` once at boot, and every descendant inherits that snapshot. Edit `/etc/TZ` later and restart majestic from the web UI or from `sysupgrade` (`sysupgrade:445-446`), and the boot-time value shadows the updated file — because majestic's built-in HTTP server makes those restarts majestic's own children.

Suggested rewrite:

```sh
# rcS exports TZ once at boot, so a later edit to /etc/TZ is shadowed by the
# stale value inherited through majestic's own children (web UI, sysupgrade).
# Re-read it here so a restart picks up the new zone without a reboot.
# musl never reads /etc/TZ itself; it uses $TZ, else /etc/localtime.
```

general/package/majestic/files/S95majestic
```diff
@@ -1,5 +1,10 @@
...
+[ -r /etc/TZ ] && export TZ=$(cat /etc/TZ)
```
widgetii, line 7: The guard is a genuine improvement over `rcS:4`/`profile:6`, and it's worth calling out explicitly because it's subtler than it looks.

In musl (`src/time/__tz.c`) an **unset** `TZ` falls back to `/etc/localtime`, but an **empty** `TZ` pins UTC and skips that fallback:

```c
s = getenv("TZ");
if (!s) s = "/etc/localtime";
if (!*s) s = __utc;
```

`rcS:4` and `profile:6` are unguarded, so a missing `/etc/TZ` makes them set `TZ=""` — forcing UTC rather than letting the `/etc/localtime` fallback run. This line avoids that. Worth porting the same `[ -r ... ] &&` guard back to `rcS:4` as a one-line follow-up.

Minor: "musl ignores `/etc/TZ`" is right, but `gm8135`, `gm8136` and `msc313e` are uClibc boards, and uClibc *does* read `/etc/TZ` natively. Harmless there — just slightly overstated as a blanket claim.

widgetii (commented):
Reviewed this against the tree. **The fix is worth taking, but the rationale attached to it is not** — the stated root cause is contradicted by `rcS`, and merging as-is would enshrine a misleading comment in the tree. Details inline; summary here.

## The stated root cause doesn't hold

The description says majestic "is launched by init, which never sources `/etc/profile`, so the `TZ=$(cat /etc/TZ)` export in there does not reach it."

The `/etc/profile` half is true, but that isn't the path that matters. [`general/overlay/etc/init.d/rcS:4`](https://github.com/OpenIPC/firmware/blob/master/general/overlay/etc/init.d/rcS#L4) already exports `TZ`, and then launches every init script as its own child:

```sh
export TZ=$(cat /etc/TZ)          # rcS:4

for i in /etc/init.d/S??*; do
	$i start                       # rcS:10 — S95majestic is a direct child
done
```

`majestic.mk:27` installs `S95majestic` into `/etc/init.d`, so it matches the `S??*` glob and inherits `TZ`. That line has been there since `rcS` was created (2cafb1a4f, Dec 2021, *"Update TimeZone settings"*). **On the boot path `TZ` already reaches majestic** — consistent with @flyrouter's "no timing issues have been noticed on the OSD in the lab."

## But there is a real bug here — a stale `TZ`, not a missing one

`rcS` exports `TZ` **once, at boot**, and that value is baked into every descendant's environment. Since majestic serves the web UI from its own built-in HTTP server, anything the UI triggers is a majestic child. So:

> Edit `/etc/TZ` on a running camera → restart majestic from the web UI, or via `sysupgrade` (which calls `S95majestic restart` at `general/overlay/usr/sbin/sysupgrade:445-446`) → **the stale boot-time `TZ` is inherited and shadows the updated file.** Only a reboot, or a restart from an interactive SSH shell (which re-sources `/etc/profile`), picks up the new zone.

Re-reading `/etc/TZ` on every `start` fixes exactly that. I suspect this is what was actually observed on the Hi3518EV200.

## Suggestions

1. **Rewrite the commit message / description** around the real cause. Suggested framing: *"re-read `/etc/TZ` on every start so a timezone change takes effect on `S95majestic restart` without a reboot; `rcS` exports it once at boot and that stale value shadows the file."*
2. **Port the `[ -r ]` guard back to `rcS:4`** (see inline note — it fixes a latent musl footgun), and consider `rcS` as the primary home since `syslogd` timestamps and `crond` schedules have the same staleness exposure. Additive, not either/or: the copy here is what buys the per-restart refresh.

## Checked and *not* problems

- **Word splitting on the unquoted `$(...)`** — tested rather than assumed: `dash -c 'export TZ=$(cat file_with_spaces)'` yields `TZ=[x y]`. `export VAR=word` parses as an assignment, so no field splitting. Matches existing usage in `rcS`/`profile`.
- **`&&` short-circuit / exit status** — no `set -e`, and the statement is at the top rather than the end, so a missing `/etc/TZ` can't leak a non-zero exit.
- **Placement before the `case`** — also exports on `stop`/`reload`. Harmless.
- **`reload`** — `killall -1` won't re-read `TZ` since SIGHUP doesn't re-exec. Inherent, not a defect in this patch.

## Risk

Minimal. Five lines in one shell script, no build-system or package-version impact, guarded against a missing file, and `/etc/TZ` ships as `GMT0` so out-of-the-box behaviour is unchanged.

*Leaving this as a comment rather than an approval since the requested change is to the rationale — happy to see it merged once the description is corrected.*

widgetii:
@phedoreanu — could you run a couple of checks on the Hi3518EV200 before this merges? I want to make sure my read of `rcS` is right and not just plausible-sounding, since my analysis was static (reading the tree) rather than a hardware repro.

### 1. The cheap decisive one

On the camera:

```sh
head -5 /etc/init.d/rcS
grep GITHUB_VERSION /etc/os-release
```

If line 4 of the running `rcS` is missing the `export TZ=$(cat /etc/TZ)` — e.g. the camera is on an older or non-stock build — that alone explains everything you saw, and my whole objection collapses. Worth ruling out first.

### 2. Cold boot on **stock** master (no patch)

```sh
echo 'CET-1CEST,M3.5.0,M10.5.0/3' > /etc/TZ
reboot
```

Then, after it comes back up:

```sh
tr '\0' '\n' < /proc/$(pidof majestic)/environ | grep ^TZ=
```

Reading `/proc/<pid>/environ` is safe here — it reports majestic's own environment, so your SSH login sourcing `/etc/profile` can't contaminate the result.

- **`TZ=CET-1CEST,...` and the OSD clock is local** → `rcS` does propagate `TZ` on the boot path, and the UTC clock you originally hit came from something other than the stated cause (most likely test 3 below).
- **`TZ` empty or absent, OSD in UTC** → I'm wrong, and there's a different root cause here worth chasing before we settle on a fix.

### 3. The staleness path — what I think this patch actually fixes

Starting from the good state at the end of test 2, still on **stock** master:

```sh
echo 'EST5EDT,M3.2.0,M11.1.0' > /etc/TZ
```

Now restart majestic **from the web UI**, not from SSH — that distinction is the whole point, because the web UI is served by majestic's own built-in HTTP server, so the restart is majestic's child and inherits its boot-time environment. Then:

```sh
tr '\0' '\n' < /proc/$(pidof majestic)/environ | grep ^TZ=
```

- On stock master I'd expect the **old** `CET-1CEST,...` to still be there — `rcS` snapshotted `/etc/TZ` once at boot and that value shadows your edit.
- With this PR applied, the same sequence should yield `EST5EDT,...`, because `S95majestic` re-reads the file on every start.

If it comes out that way, the patch is doing real work — it just isn't doing what the commit message says, and that's the only thing I'd want changed before merge.

(A related wrinkle if you test the SSH restart too: `/etc/profile` is sourced at *login*, so if you edit `/etc/TZ` after logging in, your shell is carrying a stale `TZ` as well and will pass it down. You'd need to re-login to get a fresh one. The web UI path is the cleaner demonstration.)

phedoreanu:
@widgetii ran all three checks on the Hi3518EV200. Short version: you are right on every count, and the PR has been rewritten around the staleness cause.

**Check 1 — what the unit runs.** `GITHUB_VERSION="local+build, 2026-08-08"` — a recent master-lineage local build. Its running `rcS` does carry `export TZ=$(cat /etc/TZ)` on line 4, so my original observation is not explained by a missing rcS export. Your objection stands.

**Check 2 — cold boot.** After a power-cycle with `/etc/TZ` = `CET-1CEST,M3.5.0,M10.5.0/3`:

```
# tr '\0' '\n' < /proc/$(pidof majestic)/environ | grep ^TZ=
TZ=CET-1CEST,M3.5.0,M10.5.0/3
```

`rcS` propagates `TZ` on the boot path, exactly as you read it. The original commit message's root cause was wrong.

**Check 3 — staleness.** One twist on your protocol: instead of restarting from the web UI I injected the stale environment explicitly (`env TZ=<stale> <script> restart`), which is the same inheritance mechanism a majestic-child restart exercises and avoids the login-shell contamination wrinkle you flagged. `/etc/TZ` stayed `CET-1CEST…` throughout.

Stock script (TZ block stripped):

```
# env TZ='EST5EDT,M3.2.0,M11.1.0' /tmp/S95stock restart
Stopping majestic: OK
Starting majestic: OK
# tr '\0' '\n' < /proc/$(pidof majestic)/environ | grep ^TZ=
TZ=EST5EDT,M3.2.0,M11.1.0        <-- stale caller value wins over the file
```

Patched script, same stale caller:

```
# env TZ='EST5EDT,M3.2.0,M11.1.0' /etc/init.d/S95majestic restart
Stopping majestic: OK
Starting majestic: OK
# tr '\0' '\n' < /proc/$(pidof majestic)/environ | grep ^TZ=
TZ=CET-1CEST,M3.5.0,M10.5.0/3    <-- file wins
```

The patch does real work, and it is exactly the work you described: a per-start re-read, not first-time delivery.

**Pushed as 4bde2ef:** commit message, PR title/body and the in-file comment rewritten around staleness — I took your suggested comment text nearly verbatim. The `[ -r ]` guard stays. Happy to port it to `rcS:4` and `profile:6` as a follow-up PR; per your uClibc note that follow-up would matter only on musl boards, since `gm8135`/`gm8136`/`msc313e` read `/etc/TZ` natively through uClibc.

@flyrouter this also squares with "no timing issues noticed in the lab": on a plain boot the OSD clock is correct. The stale clock only appears after editing `/etc/TZ` on a running camera and restarting majestic from the web UI or sysupgrade without a reboot.

general/package/majestic/files/S95majestic
phedoreanu: Adopted — the in-file comment, commit message and PR description are rewritten around the staleness cause in 4bde2ef, with your suggested comment text taken nearly verbatim. Hardware confirmation is in the PR-level comment: the stock script keeps an injected stale TZ across restart, the patched script picks up /etc/TZ.

general/package/majestic/files/S95majestic
phedoreanu: Guard kept, and good catch on the blanket claim — the rewritten comment scopes it to musl ("musl never reads /etc/TZ itself; it uses $TZ, else /etc/localtime"). Happy to port the same guard to rcS:4 and profile:6 as a follow-up PR; per your note that only changes anything on musl boards, since gm8135/gm8136/msc313e read /etc/TZ natively through uClibc.

widgetii (approved):
Approving. Thanks for running all three checks properly and rewriting the rationale rather than just re-pushing the fix — the hardware results settle it: `rcS` delivers `TZ` fine at boot, and what this patch actually buys is the per-start re-read. The comment, commit message and PR body now all describe that.

Two cosmetic notes, neither worth another round-trip:

- Your comment above says "Pushed as 4bde2ef", but the head is `4eca4982` — looks like one more amend landed after you wrote it. No action needed.
- The commit message describes the web UI and `sysupgrade` as "both served by majestic's built-in HTTP server". `sysupgrade` isn't served by it; it's a shell script that inherits the stale environment when it happens to be launched from the UI, and wouldn't be stale if run from a fresh SSH login. I'll tighten that one clause when squashing so the permanent record is precise — the substance is unchanged.

One thing worth stating plainly for anyone reading this later: check 3 used `env TZ=<stale> <script> restart` as a stand-in for a web UI restart. That isolates the inheritance mechanism cleanly and the mechanism itself isn't in doubt, but it means the web-UI-specific path is reasoned rather than observed — `majestic-webui` is a fetch-only package here, so the restart wiring isn't visible in this tree. The fix is correct regardless of which caller supplies the stale value.

Follow-up for the `[ -r ]` guard on `rcS:4` and `profile:6` coming separately — that one has the wider blast radius, since an unguarded empty `TZ` pins UTC on every musl board rather than just affecting majestic.

Issue #2243, opened 2026-08-09
GC2023 over MIPI on hi3516cv200: platform gaps, closed-lib behavior, and the recovered vendor init table


phedoreanu:
Findings from bringing up a VStarcam Hi3518EV200 camera (GC2023 wired to MIPI CSI-2, 2-lane RAW10) on OpenIPC, offered as raw material for proper GC2023 support — ideally an open-source driver like the recent OV2710 work (#2035, #2038, #2042). Everything below was measured on real hardware.

**1. No GC2023 config ships.** `hi3518ev200_ultimate` ships `libsns_gc2023.so`, but none of the 25 files in `/etc/sensors` is a gc2023 ini, so majestic dies with `No matches for 'gc2023_i2c' in /etc/sensors` → `Cannot start SDK` → segfault. (The only gc2023 ini anywhere in the org, glutinium's `gc2023_1080p_line.ini`, is a byte-for-byte copy of the OV9712 720p template — its values are not evidence about this sensor.)

**2. `load_hisilicon` breaks MIPI-wired boards.** The `gc2023` entry muxes the parallel VI pads (`0x200f007c..0x200f0094`), which disables the MIPI PHY. A diagnostic worth writing down: `/proc/umap/vi` showing `IntCnt=0` with `TmgErr=0` and `ccErrN=0` means *wrong bus*, not wrong timings — bad ini timings raise the error counters; a dead PHY gives zero of everything. A companion PR adds a `gc2023_mipi` identity following the OV2710 precedent, leaving the DVP entry untouched.

**3. The shipped closed lib produces no frames on this MIPI hardware.** Three libsns binaries tested on the same unit, same MIPI ini, same fixed pinmux:

| library | frames | writes digital gain `0xb1` |
|---|---|---|
| vendor MIPI `libsns_gc2023.so` | **yes** | no — leaves it 0 |
| vendor DVP `libsns_gc2023_dvp.so` | none (VENC timeout) | yes, 0x01 |
| OpenIPC `libsns_gc2023.so` | none (VENC timeout) | yes, 0x01 |

The OpenIPC lib drives the sensor over I2C fine (it writes `0xb1`), but appears to program it for DVP output.

[...]

PR #2248, opened 2026-08-09
hisilicon-osdrv-hi3516cv200: add GC2023 MIPI sensor variant

phedoreanu:
Some GC2023 boards wire the sensor to MIPI CSI-2, not DVP — e.g. VStarcam Hi3518EV200 pan/tilt cameras (2-lane RAW10). For those boards the existing `gc2023` entry in `load_hisilicon` is fatal: it muxes the parallel VI pads (`0x200f007c..0x200f0094`), which disables the MIPI PHY. The tell is `/proc/umap/vi` showing `IntCnt=0` with `TmgErr=0` and `ccErrN=0` — VI receives no VSYNC at all, while bad ini timings would at least raise the error counters.

Following the OV2710 precedent (#2035/#2038), this adds:

- a `gc2023_mipi` sensor identity that only muxes I2C0 and sets the 24 MHz sensor clock — the plain `gc2023` DVP entry is untouched;
- `gc2023_mipi_1080p.ini`, cloned from `ov2735_mipi_1080p.ini` (same SoC, same 2-lane RAW10 MIPI geometry) with the sensor identity swapped and `Isp_Bayer` corrected to 0 (RGGB — GC2023 is SRGGB10; the symptom of getting this wrong is red and blue swapping).

Verified on a VStarcam Hi3518EV200: with this identity and ini, VI `IntCnt` climbs and majestic delivers 1920×1080 at the sensor's native timing.

One honest caveat, detailed in #2243: the shipped closed-source `libsns_gc2023.so` did not produce frames on the tested unit (VENC timeout; it appears to program the sensor for DVP output) — the working setup used the vendor's MIPI libsns plus an init-table replay. The platform side fixed here is required either way, and #2243 carries the full recovered init table (123 writes, including the digital-gain `0xb1` trap) toward an open-source driver.

### DIFF
--- a/general/package/hisilicon-osdrv-hi3516cv200/files/script/load_hisilicon
+++ b/general/package/hisilicon-osdrv-hi3516cv200/files/script/load_hisilicon
@@ -193,9 +193,10 @@ insert_sns() {
                 # ov2735_i2c_1080p.ini. All share the same MIPI pinmux +
                 # clock — sensor differentiation happens in libsns / .ini.
                 # gc2023_mipi is for boards with the GC2023 wired to MIPI
-                # (e.g. VStarcam Hi3518EV200): the plain gc2023 entry below
-                # muxes the parallel VI pads, which disables the MIPI PHY —
-                # VI then sees no VSYNC at all (IntCnt=0 in /proc/umap/vi).
+                # (e.g. VStarcam Hi3518EV200): the plain gc2023 entry in the
+                # first arm above muxes the parallel VI pads, which disables
+                # the MIPI PHY — VI then sees no VSYNC at all (IntCnt=0 in
+                # /proc/umap/vi).
                 devmem 0x200f0040 32 0x2 # I2C0_SCL
                 devmem 0x200f0044 32 0x2 # I2C0_SDA
 
--- a/general/package/hisilicon-osdrv-hi3516cv200/files/script/load_hisilicon
+++ b/general/package/hisilicon-osdrv-hi3516cv200/files/script/load_hisilicon
@@ -187,11 +187,15 @@ insert_sns() {
 
                 devmem 0x2003002c 32 0x94001 # sensor unreset, clk 37.125MHz, VI 99MHz
                 ;;
-        ov2710_mipi | ov2735_i2c_1080p | ov2735)
+        ov2710_mipi | ov2735_i2c_1080p | ov2735 | gc2023_mipi)
                 # MIPI variant. OV2710 uses libsns_ov2710_mipi.so via
                 # ov2710_mipi_1080p.ini; OV2735 uses libsns_ov2735.so via
-                # ov2735_i2c_1080p.ini. Both share the same MIPI pinmux +
+                # ov2735_i2c_1080p.ini. All share the same MIPI pinmux +
                 # clock — sensor differentiation happens in libsns / .ini.
+                # gc2023_mipi is for boards with the GC2023 wired to MIPI
+                # (e.g. VStarcam Hi3518EV200): the plain gc2023 entry below
+                # muxes the parallel VI pads, which disables the MIPI PHY —
+                # VI then sees no VSYNC at all (IntCnt=0 in /proc/umap/vi).
                 devmem 0x200f0040 32 0x2 # I2C0_SCL
                 devmem 0x200f0044 32 0x2 # I2C0_SDA
 
--- a/general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
+++ b/general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
@@ -1,0 +1,284 @@
+; GC2023 1080p - MIPI (CSI-2) variant, 2-lane RAW10.
+; Select with: fw_setenv sensor gc2023_mipi
+;              cli -s .isp.sensorConfig /etc/sensors/gc2023_mipi_1080p.ini
+; Verified on a VStarcam Hi3518EV200 board with the GC2023 wired to MIPI.
+;
+[sensor]
+Sensor_type   =gc2023                   ;sensor name
+Mode          =0                        ;WDR_MODE_NONE = 0
+                                        ;WDR_MODE_BUILT_IN = 1
+                                        ;WDR_MODE_2To1_LINE = 2
+                                        ;WDR_MODE_2To1_FRAME = 3
+                                        ;WDR_MODE_2To1_FRAME_FULL_RATE =4 ...etc
+DllFile   =/usr/lib/sensors/libsns_gc2023.so    ;sensor lib path
+
+
+[mode]
+input_mode = 0                          ;INPUT_MODE_MIPI = 0
+                                        ;INPUT_MODE_SUBLVDS = 1
+                                        ;INPUT_MODE_LVDS = 2 ...etc
+
+dev_attr = 0                            ;mipi_dev_attr_t = 0
+                                        ;lvds_dev_attr_t = 1
+                                        ;NULL =2
+
+[mipi]
+;----------only for mipi_dev---------
+data_type = 2                           ;raw data type: 8/10/12/14 bit
+                                        ;RAW_DATA_8BIT = 1
+                                        ;RAW_DATA_10BIT = 2
+                                        ;RAW_DATA_12BIT = 3
+                                        ;RAW_DATA_14BIT = 4
+lane_id = 0|1|-1|-1|-1|-1|-1|-1|        ;lane_id: -1 - disable
+
+[lvds]
+;----------only for lvds_dev---------
+img_size_w = -1                         ;oringnal sensor input image size W
+img_size_h = -1                         ;oringnal sensor input image size H
+wdr_mode = -1                           ;HI_WDR_MODE_NONE =0
+                                        ;HI_WDR_MODE_2F = 1
+                                        ;HI_WDR_MODE_3F = 2
+                                        ;HI_WDR_MODE_4F =3
+sync_mode = -1                          ;LVDS_SYNC_MODE_SOL = 0
+                                        ;LVDS_SYNC_MODE_SAV = 1
+raw_data_type = -1                      ;RAW_DATA_8BIT = 0
+                                        ;RAW_DATA_10BIT = 1
+                                        ;RAW_DATA_12BIT = 2
+                                        ;RAW_DATA_14BIT = 3
+data_endian = -1                        ;LVDS_ENDIAN_LITTLE = 0
+                                        ;LVDS_ENDIAN_BIG = 1
+sync_code_endian =-1                    ;LVDS_ENDIAN_LITTLE = 0
+                                        ;LVDS_ENDIAN_BIG = 1
+lane_id = -1|-1|-1|-1|-1|-1|-1|-1|      ;lane_id: -1 - disable
+lvds_lane_num = -1                      ;LVDS_LANE_NUM
+wdr_vc_num = -1                         ;WDR_VC_NUM
+sync_code_num = -1                      ;SYNC_CODE_NUM
+sync_code_0 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_1 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_2 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_3 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_4 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_5 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_6 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_7 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+
+[isp_image]
+Isp_x      =0
+Isp_y      =0
+Isp_W      =1920
+Isp_H      =1080
+Isp_FrameRate=25
+Isp_Bayer  =0   ;BAYER_RGGB=0, BAYER_GRBG=1, BAYER_GBRG=2, BAYER_BGGR=3 - GC2023 is SRGGB10
+
+
+[vi_dev]
+Input_mod =5    ;VI_INPUT_MODE_BT656 = 0
+                ;VI_INPUT_MODE_BT601,
+                ;VI_INPUT_MODE_DIGITAL_CAMERA
+Work_mod =0     ;VI_WORK_MODE_1Multiplex = 0
+                ;VI_WORK_MODE_2Multiplex,
+                ;VI_WORK_MODE_4Multiplex
+Combine_mode =0 ;Y/C composite or separation mode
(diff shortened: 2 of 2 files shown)


flyrouter (commented):
Thanks. 
It looks like an interesting proposal for improvement.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
```diff
@@ -0,0 +1,284 @@
+; GC2023 1080p - MIPI (CSI-2) variant, 2-lane RAW10.
+; Select with: fw_setenv sensor gc2023_mipi
+;              cli -s .isp.sensorConfig /etc/sensors/gc2023_mipi_1080p.ini
+; Verified on a VStarcam Hi3518EV200 board with the GC2023 wired to MIPI.
```
widgetii: This is the one thing I would block on. As committed, this file plus its own `DllFile` does not stream — per your table in #2243, the OpenIPC-shipped `libsns_gc2023.so` gives a VENC timeout on MIPI-wired hardware. The verification was done with the vendor MIPI libsns, which is not what this file loads.

Please make that explicit, so nobody selects this preset and spends a day chasing IQ or optics:

```suggestion
; NOTE: this does NOT work with the libsns_gc2023.so that OpenIPC ships -
; that library appears to program the sensor for DVP output and yields no
; frames on a MIPI-wired board (VENC timeout). DllFile below must point at
; a GC2023 MIPI sensor library, which OpenIPC does not currently ship.
; The MIPI geometry here was validated on a VStarcam Hi3518EV200 using the
; vendor's MIPI libsns - see #2243.
```

widgetii (commented):
Thanks — this is unusually well-evidenced work, and the writeup in #2243 is genuinely useful material. I reviewed both halves separately, because they have very different risk and value profiles.

## The `load_hisilicon` half: correct, and I want it

No regression risk, verified rather than assumed:

* `case` matching is exact-glob and the DVP `gc2023` arm comes first, so no existing `sensor` value changes behaviour.
* Kernel side: `open_sys_config` for cv200 carries its own duplicate pinmux table (`kernel/sys_config/hi3516cv200/sys_config.c`) and falls through to `printk("sensor_type '%s' is error!!!")` for names it does not know — non-fatal, and the script's `devmem` writes run afterwards and win. `gc2023`, `ov2735` and several others already hit that same branch today.
* The vendor `hi3518e_sys.ko` takes `sensor` as a bare `charp` with no name table, so an unknown value is inert there too.
* Affected defconfigs are `hi3516cv200_lite`, `hi3518ev200_lite`, `hi3518ev200_ultimate`, `hi3518ev201_lite`. The new `.ini` is installed unconditionally by the `*.ini` glob in the `.mk`: ~10 KB raw, ~3 KB compressed. Irrelevant even on the tight lite boards.

One asymmetry worth flagging: the OV2710 precedent you follow (#2038) updated **both** `sys_config.c` in openhisilicon and `load_hisilicon`. This PR only does the latter. That is harmless, but it means dmesg now gains a `sensor_type 'gc2023_mipi' is error!!!` line, which looks alarming to anyone reading a boot log. Adding the arm to `sys_config.c` as well would keep the two tables in sync and keep the log clean.

## The `.ini` half: one real objection, and it is the header

From your own table in #2243, on the same unit with the same MIPI ini and the same fixed pinmux:

| library | frames |
|---|---|
| vendor MIPI `libsns_gc2023.so` | yes |
| vendor DVP `libsns_gc2023_dvp.so` | none (VENC timeout) |
| OpenIPC `libsns_gc2023.so` | none (VENC timeout) |

`DllFile` in this file points at row three. So anyone who follows this file's own instructions gets a VENC timeout. That is fine — it is a platform-side template, and it is useful to someone who supplies their own MIPI libsns and edits one line — but the header currently says *"Verified on a VStarcam Hi3518EV200 board with the GC2023 wired to MIPI"*, which is not true of this file with this `DllFile`. Verification was done with a library that is not the one named here. That comment will cost somebody a day.

**Before merge I would like just the header corrected** (suggestion inline). The frame-rate and Bayer notes are questions, not blockers.

## On "what is the point if the camera still does not work"

Answering it here since it will come up. The platform half is a prerequisite that no library can work without — without it the MIPI PHY is dead and the sensor is unreachable on these boards regardless of what userspace you throw at it — and it costs nothing. The `.ini` is scaffolding that the open driver your init table in #2243 makes possible will plug straight into. Both are worth having, as long as the file is honest about what does and does not work today. That is the whole of my ask.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
```diff
@@ -0,0 +1,284 @@
+; GC2023 1080p - MIPI (CSI-2) variant, 2-lane RAW10.
+; Select with: fw_setenv sensor gc2023_mipi
+;              cli -s .isp.sensorConfig /etc/sensors/gc2023_mipi_1080p.ini
```
widgetii, line 3: Naming nit, non-blocking, but worth a check on hardware while you still have the unit.

Every legacy file in this directory follows `<sensor>_<bus>[_extras]_<res>.ini` — `ar0130_i2c_720p.ini`, `imx323_spi_dc_1080p.ini`, `sc1235_i2c_dc_960p.ini`. The only exceptions are the three recent additions: `ov2710_mipi_1080p.ini`, `ov2710_dc_1080p.ini`, `ov2735_mipi_1080p.ini`.

That convention is not cosmetic — majestic globs `/etc/sensors` for a pattern derived from the sensor identity, and #2243 records it searching for exactly `gc2023_i2c`. If the rule really is `<sensor>_<bus>`, then with `sensor=gc2023_mipi` it will look for `gc2023_mipi_i2c` and never auto-select this file, making the explicit `cli -s` step mandatory forever. Naming it `gc2023_mipi_i2c_1080p.ini` would auto-resolve, and would still not collide with the DVP `gc2023` pattern.

Worth one experiment: set `sensor=gc2023_mipi`, clear `.isp.sensorConfig`, and see whether majestic logs `matched sensor config:` or `No matches for`.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
```diff
@@ -0,0 +1,284 @@
...
+; Verified on a VStarcam Hi3518EV200 board with the GC2023 wired to MIPI.
+;
+[sensor]
+Sensor_type   =gc2023                   ;sensor name
+Mode          =0                        ;WDR_MODE_NONE = 0
+                                        ;WDR_MODE_BUILT_IN = 1
+                                        ;WDR_MODE_2To1_LINE = 2
+                                        ;WDR_MODE_2To1_FRAME = 3
+                                        ;WDR_MODE_2To1_FRAME_FULL_RATE =4 ...etc
+DllFile   =/usr/lib/sensors/libsns_gc2023.so    ;sensor lib path
```
widgetii, line 19: This is the single field that was not verified — everything else in the file was exercised, but through a different `.so` than the one named here. Worth stating in the header (see my comment on the block above) rather than leaving the reader to infer it from the PR description.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
```diff
@@ -0,0 +1,284 @@
...
+sync_code_5 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_6 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+sync_code_7 = -1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|-1|
+
+[isp_image]
+Isp_x      =0
+Isp_y      =0
+Isp_W      =1920
+Isp_H      =1080
+Isp_FrameRate=25
```
widgetii: Inherited verbatim from the `ov2735_mipi` template, which is a 25 fps config. But the table you recovered in #2243 is `sensor_linear_1080p30_init`.

If the sensor is actually running at 30 and the ISP is told 25, the AE limits are computed against the wrong frame time — which would fit with the exposure oddities you documented (`MaxLine=2`, `Error=-196`). Is 25 deliberate here, or a leftover of the clone?

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
```diff
@@ -0,0 +1,284 @@
...
+Isp_Bayer  =0   ;BAYER_RGGB=0, BAYER_GRBG=1, BAYER_GBRG=2, BAYER_BGGR=3 - GC2023 is SRGGB10
```
widgetii, line 79: RGGB is plausible for GC2023 and I have no reason to doubt it — just noting for the record that this, like the rest of the file, was validated through the vendor libsns rather than the one `DllFile` names. If a future open driver applies a mirror or flip in the sensor init, the Bayer phase shifts and this will need revisiting.

general/package/hisilicon-osdrv-hi3516cv200/files/script/load_hisilicon
```diff
@@ -187,11 +187,15 @@ insert_sns() {
...
                 ;;
-        ov2710_mipi | ov2735_i2c_1080p | ov2735)
+        ov2710_mipi | ov2735_i2c_1080p | ov2735 | gc2023_mipi)
                 # MIPI variant. OV2710 uses libsns_ov2710_mipi.so via
                 # ov2710_mipi_1080p.ini; OV2735 uses libsns_ov2735.so via
-                # ov2735_i2c_1080p.ini. Both share the same MIPI pinmux +
+                # ov2735_i2c_1080p.ini. All share the same MIPI pinmux +
                 # clock — sensor differentiation happens in libsns / .ini.
+                # gc2023_mipi is for boards with the GC2023 wired to MIPI
+                # (e.g. VStarcam Hi3518EV200): the plain gc2023 entry below
```
widgetii: Small factual slip: the plain `gc2023` entry is *above* this arm, not below — it is in the first case arm (`9m034 | jxf23 | ... | gc2023`) near the top of `insert_sns`. Worth fixing so the comment does not send the next reader scrolling the wrong way.

The substance of the comment is right, and I confirmed it: that arm muxes `0x200f007c..0x200f0094` to VI_DATA/VI_VS/VI_HS, and leaving those pads alone is exactly what the merged `ov2710_mipi` arm does.

phedoreanu:
All asks addressed, and the naming experiment produced a more interesting answer than either of us guessed.

**Header (b168796):** your text taken essentially verbatim. The file now leads with the NOTE that the OpenIPC-shipped `libsns_gc2023.so` does not stream on a MIPI-wired board and that `DllFile` must point at a GC2023 MIPI library OpenIPC does not currently ship; verification credit goes to the vendor MIPI libsns, with a pointer to #2243. The same commit adds the `Isp_FrameRate` note and fixes the "below"/"above" slip in the `load_hisilicon` comment.

**The naming experiment, run on the unit:**

```
# fw_setenv sensor gc2023_mipi
# ipcinfo -s
gc2023
# majestic -s   (foreground)
DEBUG [sdk] try_sensor_config@5428: Autodetected sensor as 'gc2023_i2c'
DEBUG [sdk] find_sensor_config@5388: matched sensor config: gc2023_i2c_1080p.ini
```

The search key does not come from the `sensor` env at all: majestic autodetects by I2C probe and derives `gc2023_i2c` even with `sensor=gc2023_mipi` set. So `gc2023_mipi_i2c_1080p.ini` would not auto-resolve either — the only auto-matching name is a `gc2023_i2c*` prefix, and that key cannot distinguish a MIPI-wired board from a DVP-wired one (the suffix records how majestic talks to the sensor, not how video leaves it). Since this file needs a manually supplied MIPI libsns anyway, staying off the auto-match path and requiring the explicit `cli -s` seems right — the same situation `ov2710_mipi_1080p.ini` / `ov2710_dc_1080p.ini` are in against their probed `ov2710_i2c` key. Name kept.

**sys_config.c asymmetry:** fixed for real rather than flagged — OpenIPC/openhisilicon#207 adds the `gc2023_mipi` arm as a mirror of the `ov2710_mipi` one, so a `gc2023_mipi` boot loses the `sensor_type 'gc2023_mipi' is error!!!` line and the two pinmux tables stay in sync.

**Isp_FrameRate 25 vs 30:** a leftover of the clone — but a leftover the verification ran with. The working vendor-libsns session used this file with 25 in place, while the recovered init table is `sensor_linear_1080p30_init`. Your theory fits the AE oddities in #2243 (`MaxLine=2`, `Error=-196`), so I noted the tension next to the value rather than silently flipping to an untested 30; it should be revisited together with the AE limits once an open MIPI driver exists (#2243 carries the full init table for that).

And thank you for the closing paragraph — "prerequisite plus scaffolding, as long as the file is honest" is exactly the framing I will steal for the next one of these.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
phedoreanu: Taken essentially verbatim in b168796 — the header now leads with this NOTE and credits the vendor MIPI libsns, pointing at #2243.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
phedoreanu: Ran it — result in the PR-level comment. Short version: the key is probe-derived, not env-derived. majestic logs "Autodetected sensor as 'gc2023_i2c'" and matches gc2023_i2c_1080p.ini even with sensor=gc2023_mipi set, and ipcinfo -s still reports gc2023. So gc2023_mipi_i2c_1080p.ini would not auto-resolve either, and any name that does auto-resolve (a gc2023_i2c* prefix) cannot distinguish MIPI-wired from DVP-wired boards. Kept the name; explicit cli -s stays the selection mechanism, same as the ov2710 mipi/dc pair.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
phedoreanu: Now stated in the header itself (b168796): DllFile is the one field the verified session did not exercise, and the header says outright that the shipped library does not stream on MIPI-wired hardware.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
phedoreanu: Leftover of the clone — but a leftover the verification ran with: the working vendor-libsns session used this file with 25 in place, while the recovered table is sensor_linear_1080p30_init. That mismatch existing during a working session is itself a data point, and your theory fits the AE oddities in #2243. b168796 notes the tension next to the value; flipping to 30 stays untested until an open MIPI driver exists to test it with.

general/package/hisilicon-osdrv-hi3516cv200/files/sensor/config/gc2023_mipi_1080p.ini
phedoreanu: Agreed, for the record: RGGB was validated through the vendor libsns, and a future open driver applying mirror/flip in sensor init shifts the phase. The comment on Isp_Bayer already pins the CFA reasoning, so leaving it as is.

general/package/hisilicon-osdrv-hi3516cv200/files/script/load_hisilicon
phedoreanu: Fixed in b168796 — the comment now says "the plain gc2023 entry in the first arm above".

PR #2249, opened 2026-08-10
sysupgrade: don't reboot on a failure before the first flash write

widgetii:
`die()` called `reboot_system()` unconditionally, so every failure path — a download that timed out, `Check your network!`, a bad checksum, the wrong SoC — rebooted the camera. Before the first write nothing has changed, so that reboot fixes nothing and costs two things: it destroys `/tmp` along with the log that says what went wrong, and to anything watching the camera come back it is indistinguishable from a successful upgrade.

The WebUI drives this over `/ws/upgrade` and has no other signal. Its "did it really reboot?" check was added for exactly this symptom (OpenIPC/majestic-webui#123), on the assumption that a failed upgrade leaves the camera up — which `die()` made false. So a download that never finished was reported as an update that worked. That is **OpenIPC/majestic-webui#120**, still open against the build it was closed on.

### What changes

- **`die()` weighs whether flash was touched.** New `mark_flash_touched`, set before every write — mirroring `mark_live_flash_dirty` but broader, since it also covers the kernel partition (not mounted, so it never forces a reboot on the success path, but just as final once `flashcp` has started). Past that point the partition is erased whether or not the write finished, so rebooting stays the least-bad option.
- **The lock is released on the non-rebooting path.** The reboot used to double as lock cleanup, since `/tmp` went with it; without this the retry the user is being invited to make would be refused.
- **`-m 120` → `--speed-limit 1024 --speed-time 60`** (with `-m 1800` as an outer backstop). A total-elapsed cap is not a timeout, it is a disguised throughput floor: at ~7 MB an image, any link slower than ~60 KB/s could never finish, and curl cut the transfer mid-stream with nothing to say that slowness was the reason. This is what the reporter's screenshot shows — a progress bar frozen at 32.7%.

### Verification

Compared 1.0.56 and 1.0.57 on the lab hi3516av300 against an identical pre-flash failure, with `busybox reboot` shadowed via `PATH` so the comparison carried no risk:

```
OLD 1.0.56:  "File ... not found Aborting." -> "Unconditional reboot" -> REBOOTED, lock left behind
NEW 1.0.57:  "File ... not found Aborting." -> did NOT reboot, lock released
```

Also confirmed on-device that curl 7.76.0/mbedTLS accepts the new flags, and that `busybox sh -n` passes (local `sh -n` does not exercise ash).

4 new behaviour cases and 6 new invariants in `test_sysupgrade.sh`, all of which fail against 1.0.56; the full suite passes.

One case worth flagging for review: a failed `flashcp` is **not** fatal on the split path (`rc=0`, no `|| die`), so the post-write `die()` case is only reachable via the combined-image whole-blob layout, which is what the new test uses. That non-fatality looks like a separate bug, untouched here.

### DIFF
--- a/.github/scripts/test_sysupgrade.sh
+++ b/.github/scripts/test_sysupgrade.sh
@@ -74,6 +74,10 @@ grep -q '@SB@\|@OSRELEASE@' "$SB/sysupgrade" \
 for s in S99rc.local S60crond S49ntpd S02klogd S01syslogd; do
     printf '#!/bin/sh\nexit 0\n' > "$SB/etc/init.d/$s"; chmod +x "$SB/etc/init.d/$s"
 done
+# This one logs: whether majestic is put back is a behaviour worth asserting,
+# not just a service that had to exist for the script not to trip over it.
+printf '#!/bin/sh\necho "S95majestic $1" >> "$FLASH_LOG"\nexit 0\n' \
+    > "$SB/etc/init.d/S95majestic"; chmod +x "$SB/etc/init.d/S95majestic"
 
 set_mtd() { cat > "$SB/proc/mtd"; }
 
@@ -662,6 +666,17 @@ else
     bad "pre-write refusal left services down; the camera stays up but degraded"
 fi
 
+# majestic must be RESTARTED, not started. free_resources' `killall -3` is not a
+# terminate -- SIGQUIT makes majestic release the SDK and keep serving -- so it
+# is still running, just with no video pipeline and no way back in place. Under
+# --web it is worse: majestic latched itself into upgrade mode and never clears
+# the flag.
+if grep -q "S95majestic restart" "$SB/tmp/flash.log"; then
+    ok "pre-write refusal restarts majestic (a gutted daemon needs more than start)"
+else
+    bad "majestic left running without its SDK; video stays down until a power cycle"
+fi
+
 # A lock this run did not take belongs to somebody else. die() is reachable
 # before create_lock, so removing the lock unconditionally there would unlink a
 # CONCURRENT sysupgrade's lock and let two upgrades run at once. run() clears
@@ -746,6 +761,9 @@ awk '/^create_lock\(\)/,/^}/' "$SRC" | grep -q 'lock_owned=1' \
 awk '/^die\(\)/,/^}/' "$SRC" | grep -q 'restore_resources' \
     && ok "die() restarts the services free_resources stopped" \
     || bad "a pre-flash die() leaves syslog/klogd/ntpd/cron stopped on a camera that stays up"
+awk '/^restore_resources\(\)/,/^}/' "$SRC" | grep -q 'S95majestic restart' \
+    && ok "restore_resources restarts majestic rather than starting it" \
+    || bad "SIGQUIT leaves majestic running without an SDK; 'start' is a no-op, it needs 'restart'"
 grep -q 'mark_flash_touched' "$SRC" \
     && ok "the flash-touched marker exists" \
     || bad "mark_flash_touched is gone; die() cannot tell a pre-write failure apart"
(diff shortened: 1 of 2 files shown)


widgetii:
Both review findings were correct and are fixed in `e186a645`.

**Unowned lock file.** Confirmed reachable — `print_sysinfo` → `get_system_info` runs at line 739, well before `create_lock` at 768. `create_lock` now records ownership and `die()` only removes a lock this run created.

**Pre-flash failure stops services.** This one was a genuine regression, and the sharper version of it: the reboot was doing the cleanup, so removing the reboot exposed it. I'd actually tripped over the symptom while testing on the lab camera and restarted the services by hand without connecting it to the change.

I did *not* take the suggested "move `free_resources()` after download/verify" — that would defeat its purpose. Dropping the page cache there is how a small camera finds the RAM to unpack the image (see #2236's reasoning), so moving it past the download would break the low-RAM boards it exists for. Went with the alternative instead: an explicit `restore_resources()` on both paths that end without rebooting — `die()` before the first write, and the `-x` honoured path in `reboot_system`, which had the same hole already.

One thing the test surfaced: `free_resources`' `killall -q -3 majestic` was **ignored** on this hi3516av300 build — same PID before and after — so majestic is started back only when `pidof` says it is actually gone, rather than trusting the kill. That non-firing SIGQUIT looks worth a separate look, since freeing that memory is the point of the call.

Re-verified on the lab hi3516av300 with `busybox reboot` shadowed via `PATH`:

```
pre-flash failure:  did not reboot; printed restart notice; released its own lock
                    after: syslogd:up klogd:up ntpd:up crond:up majestic:up
foreign lock:       "Another sysupgrade process is already running!", lock preserved
```

2 new behaviour cases and 3 new invariants; the full suite passes.

widgetii:
**Correction to my previous comment**, and it turned up something more serious. Fixed in `b2bfa895`.

I said `free_resources`' `killall -q -3 majestic` "was ignored" on the hi3516av300. That was wrong. It works exactly as designed — **SIGQUIT is not a terminate for majestic**: it asks the daemon to release the video pipeline and keep serving, where SIGINT/SIGTERM ask it to exit.

Same pid before and after is therefore correct, and `/proc/<pid>/status` confirms the signal is caught rather than ignored.

**But what it leaves behind is the actual problem.** The daemon keeps answering HTTP with no video pipeline — `/image.jpg` returns **200 with zero bytes** — and nothing restores it in place: a SIGHUP reload will not revive it, and `start` is a no-op against a process that is still running, so my previous `pidof`-guarded `start` was dead code. Only a restart brings video back.

Under `--web` it is worse. There is no kill at all, but majestic put *itself* into upgrade mode before spawning us, and `g_upgrade_in_progress` has exactly one write site (`main.c:249`, set true) and is never cleared. So the camera would sit video-less and half-served indefinitely.

So: **restart majestic on any run that ends without rebooting.** Under `--web` that drops the `/ws/upgrade` socket streaming the log, which is fine — `Aborting.` has already gone out and the browser treats a close after an abort as final, not as a reboot. Letting majestic exit instead is not an option: it is unsupervised (`/etc/inittab` respawns only getty).

Verified on the lab hi3516av300:

```
baseline:            majestic pid 970,  /image.jpg http=200 bytes=0        <- already gutted
after pre-flash fail: majestic pid 2844, /image.jpg http=200 bytes=372332  <- video back
                      cgi 200, syslogd/klogd/ntpd/crond all up, no reboot
```

That zero-byte baseline was left by an *earlier* test of mine on the same camera — I had checked `pidof majestic` and called it healthy. Which is how directly this bites: **`pidof` reports green on a camera with no video.** Check `%{size_download}` on `/image.jpg` instead.

New assertions: `restore_resources` must use `restart`, plus a behavioural check that a pre-flash refusal actually invokes it.

Issue #2231, opened 2026-07-24
sysupgrade `--no_reboot` (-x) corrupts the running rootfs and locks out SSH after an on-device flash


widgetii:
## Summary

Running `sysupgrade … -x` (`--no_reboot`) on a NOR camera that flashes **its own live rootfs** completes the flash successfully but leaves the running system unusable: services are stopped **and** — because the mounted rootfs MTD partition is overwritten under the running kernel *after* the page cache was dropped — subsequent reads of on-disk files return the new image's bytes at stale offsets. The live filesystem is effectively corrupt: SSH key auth (and most commands) start failing. The board still answers ping but you can no longer log in to issue the "manual reboot" the tool advises. **Only a power cycle recovers it**, which defeats the entire purpose of `--no_reboot`.

## Environment

- OpenIPC System Updater **v1.0.55** (`general/overlay/usr/sbin/sysupgrade`)
- gk7205v200 (Goke), `fpv` variant, 8 MB NOR, kernel 4.9.37, squashfs rootfs on NOR (`flashcp` split path)
- Invocation: `sysupgrade --url=<builder …-nor.tgz> -z -x`

## Steps to reproduce

1. On a NOR camera, run: `sysupgrade --url=<…-nor.tgz> -z -x`
2. Let it finish — kernel + rootfs both flash and verify 100%.
3. Note the closing message: *"You asked me not to reboot … Please reboot the camera manually whenever possible."*
4. Try to SSH back in (public-key auth that worked moments earlier):
   `ssh root@<camera>` → `Permission denied (publickey,password).`
5. The camera still replies to ping but is unloginnable and non-functional. A power cycle is the only recovery.

## Actual output (trimmed)

```
Stop services, sync files, free up memory
…
Kernel updated to 13:55:21 2026-07-24
RootFS updated to <ver>, 2026-07-24
You asked me not to reboot, so I won't.
Although a reboot is required to apply the changes.
Please reboot the camera manually whenever possible.
```
then, from the operator's host:
```
$ ssh root@<camera>
root@<camera>: Permission denied (publickey,password).
```
(ping still succeeds — the box is up, just corrupted and locked out.)

## Root cause

All line numbers are from `v1.0.55`:

- `free_resources()` stops `majestic` + services, then `sync; echo 3 > /proc/sys/vm/drop_caches` (~L340–356) to free RAM for the download/extract.
[...]

PR #2251, opened 2026-08-10
sysupgrade: flash from a ramfs, so the reboot cannot be eaten by the flash

widgetii:
A WebUI upgrade could leave a camera needing a **physical power cycle**: the log froze, the box stopped answering, and after the power cycle it was running the *new* firmware. So the flash completed and the reboot did not.

Reported in OpenIPC/majestic-webui#120 (gk7205v300, 64 MB), and distinct from the false-success bug fixed in #2249.

### Mechanism

`flashcp` rewrites the MTD partition backing the mounted squashfs rootfs, and that write goes to the raw `/dev/mtdN` character device — behind the block layer, so the page cache is never invalidated. The running system therefore keeps working for exactly as long as the pages it needs stay **resident**; anything faulted back in afterwards reads the new image at a stale offset. `free_resources()` has already dropped the cache to make room for the download, so on a 64 MB camera eviction is likely rather than theoretical — and on a 754 MB one it never happens, which is why this reproduces for some users and never in the lab.

What that kills is the reboot: `busybox reboot` is a fork+exec of an 867 KB binary living on the partition just overwritten, so it fails with EIO and nothing reboots.

Same mechanism @widgetii documented for the `-x` path in #2231; #2236 forced a reboot there, which cured that symptom without removing the hazard.

**Reproduced deterministically with no MTD writes at all** — copy the rootfs squashfs to tmpfs, loop-mount it, and rewrite the backing file:

```
exec from loop-squashfs         -> OK
overwrite backing store         -> (what flashcp does to mtd3)
exec again, pages still cached  -> OK        <- why it looks fine on a big camera
drop_caches, exec again         -> Input/output error, rc=126
```

### Fix

Move the flashing phase into RAM, the way OpenWrt's sysupgrade does. `enter_ramfs` stages busybox, its loader and libraries, and the script into a tmpfs, moves `/dev`, `/proc` and `/tmp` across, `pivot_root`s and **re-execs**.

The re-exec is the part that matters: `pivot_root` changes what paths resolve to, but the running shell keeps its text mapped from the old squashfs inode — a pivot without one is theatre.

Failure is always backwards-compatible: any step that does not work returns, and the caller flashes in place exactly as before, after lining up the least fragile reboot available (sysrq needs no fork, no exec and no new mapping; a tmpfs copy of busybox is next). `--no_ramfs` forces that path for board bring-up.

Two things only the hardware found:
- busybox dispatches on `argv[0]`, so the first run pivoted perfectly and then could not find `od`, `cut`, `grep` or `flashcp`. Applet symlinks are not optional.
- a `die()` after the pivot must reboot whatever it was doing — the old system is behind `/mnt`, so exiting there leaves precisely the corpse this change exists to prevent.

### Also: drop the `--web` pre-flash `killall -q -3 majestic`

`--web` exists so majestic can keep streaming the log; the daemon has already released its video pipeline by that point; and SIGQUIT to a majestic already in upgrade mode is unsafe, and is fixed in the daemon. The log survives the pivot anyway — `tail`'s fd and the websocket are open file descriptors, and moving a mount does not disturb them.

### Verified on a lab hi3516av300

After the pivot: `/` is tmpfs, busybox is on tmpfs, the old root is at `/mnt`, `/dev/mtd3`, `/proc/mtd` and the staged image are all reachable, all 14 tools the flash phase uses resolve, and `od`/`grep` still work after an explicit `drop_caches`. Rebooted from RAM via sysrq and came back clean.

7 new behaviour cases and 12 new invariants; **83 checks pass**.

One thing reviewers should weigh: the real pivot cannot be exercised by the offline suite (it needs root and real mounts), so it is covered by source invariants plus the hardware run above. The fallback — the property that a camera which cannot build a ramfs still upgrades — *is* covered behaviourally, since the stub fails by default.

Refs OpenIPC/majestic-webui#120.

### DIFF
--- a/.github/scripts/test_sysupgrade.sh
+++ b/.github/scripts/test_sysupgrade.sh
@@ -64,6 +64,7 @@ sed -e 's|grep "GITHUB_VERSION" "$1/etc/os-release"|grep "GITHUB_VERSION" "${1:-
     -e "s|/proc/cmdline|@SB@/proc/cmdline|g" \
     -e "s|/proc/sys/vm/drop_caches|@SB@/tmp/drop_caches|g" \
     -e "s|/etc/init.d/|@SB@/etc/init.d/|g" \
+    -e "s|/bin/busybox|@SB@/bin/busybox|g" \
     -e "s|@OSRELEASE@|/etc/os-release|g" \
     -e "s|@SB@|$SB|g" \
     "$SRC" > "$SB/sysupgrade"
@@ -110,6 +111,10 @@ stub killall    'exit 0'
 stub ntpd       'exit 0'
 stub curl       'exit 0'
 stub umount     'exit 0'
+# Logs so ordering can be asserted, and fails by default: an unprivileged test
+# host cannot really pivot, and the fallback is the safety property that matters
+# most here (a camera that cannot build a ramfs must still upgrade).
+stub pivot_root 'echo "pivot_root $*" >> "$FLASH_LOG"; exit ${STUB_PIVOT_RC:-1}'
 stub losetup    'case "$1" in -f) echo /dev/loop0;; *) exit 0;; esac'
 
 # download_firmware runs `md5sum -s -c`. -s (silent) is a busybox extension; GNU
@@ -150,6 +155,15 @@ exit 0'
 stub mount '
 [ $# -eq 0 ] && exit 0
 target=${!#}
+# STUB_MOUNT models the VERIFY-mount (a squashfs image over a loop device). The
+# ramfs pivot mounts tmpfs and relocates existing mounts; those are a different
+# operation and must not inherit the fault being injected, or the hang case
+# wedges before the code under test is even reached.
+for a in "$@"; do
+    case "$a" in
+        move|tmpfs|--move|-M) exit 0 ;;
+    esac
+done
 case "${STUB_MOUNT:-ok}" in
     ok)
         mkdir -p "$target/etc"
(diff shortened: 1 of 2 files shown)


widgetii:
Validated on **gk7205v200 with 34 MB of RAM** — the class of camera this hang actually happens on. Every earlier run was on a 754 MB hi3516av300, which can never reproduce the bug in the first place.

### It works, and it is cheap

```
=== PHASE 2: inside the ramfs ===
  / is            : tmpfs
  busybox lives on: tmpfs
  old root at /mnt: overlay
  ramfs usage     : 1.1M of 8.0M
  free RAM        : 9M free, 19M avail     (10M / 21M before)
  --- tools the flash phase uses ---
  all present
  od    : 67adf0b5      xxd   : 67adf0b5
  losetup -f: /dev/loop0    get_device-style: mtd3
  --- drop caches, then keep working ---
  after drop_caches: od=67ad grep=6 sed=y OK
  --- reboot from RAM via sysrq ---
```

Came back clean: uptime 3 min, `/` is overlay again, majestic up, video serving.

The staged root measures **1.1 MB** and the whole run costs **1 MB of free RAM**, so the 16 MB cap I had was sized for the wrong machine — dropped to 8 MB in `147e279d`. The cap is not a reservation, and `/tmp` (holding the unpacked image) is a separate mount that does not draw on it.

### What this board exercised that the lab one could not

- **musl instead of glibc**, where the loader *is* libc — `ldd` reports `/lib/ld-musl-arm.so.1` for both entries. The library staging picks it up correctly; had it not, the re-exec would have died with the rootfs still mapped, which is the one failure mode worse than the bug.
- A squashfs rootfs on a 5 MB `mtd3` with `/dev` on devtmpfs, confirming the mount moves behave the same across libc and vendor.
- 17 MB of tmpfs rather than 377 MB, so the sizing is now measured rather than assumed.

Harness still green at 83 checks.

Remaining gap, stated plainly: this is now verified on two SoCs, two libcs and two RAM sizes, but the *combination* that has never been run end to end is a real `flashcp` writing a real rootfs from inside the ramfs — every test so far stops short of the actual write by design. The fallback path is unchanged and behaviourally covered, so the downside of a surprise there is a camera that flashes the old way, not one that bricks.

widgetii:
Superseded by #2252, which is the same fix without the detour.

This branch carried two mechanisms for one job: the ramfs pivot, plus an earlier half-measure I had started before it (staging a busybox copy in `/tmp`, threading a `$safe_bb` through `set_progress`, and a sysrq reboot path with its own state and announcement). Once the pivot landed, that was dead weight — and hardware confirms it is unnecessary: `busybox reboot -d 1 -f` works fine from inside the ramfs on both a 34 MB musl board and a 754 MB glibc one, because by then busybox is on tmpfs.

#2252 is a single commit off master with that removed, re-verified on both boards and green at 82 checks.

Issue #239, opened 2022-06-20
Отмонтировать корневую фс при обновлении


p0i5k:
Делать pivot_root и chroot при выполнении sysupgrade.
Это позволит обновлять устройства с ubifs, а также может помочь от единичных глюков при обновлении

widgetii:
#2252 implements this: `sysupgrade` now does `pivot_root` + `chroot` before it writes anything.

`enter_ramfs` stages busybox, its loader and libraries, and the script into a tmpfs, moves `/dev`, `/proc` and `/tmp` across, pivots, and **re-execs**. That last part turned out to be the crux: `pivot_root` changes what paths resolve to, but the already-running shell keeps its text mapped from the old squashfs inode, so a pivot without a re-exec doesn't actually get you off the old rootfs.

### On the "единичные глюки при обновлении" half

There was a concrete one behind that, and it is now diagnosed. `flashcp` rewrites the MTD partition backing the mounted squashfs, and the write goes to the raw `/dev/mtdN` character device — behind the block layer — so the page cache is never invalidated. The running system keeps working for exactly as long as its pages stay **resident**; anything faulted back in afterwards reads the new image at a stale offset. Since `free_resources()` drops the cache to make room for the download, on a small camera that eviction is likely.

What it broke was the reboot itself: `busybox reboot` is a fork+exec of a binary living on the partition just overwritten, so it failed with EIO and the camera never rebooted — it sat there with a corrupt live filesystem until someone power-cycled it. Reported from the field in OpenIPC/majestic-webui#120.

Reproducible with no MTD writes at all: copy the rootfs squashfs to tmpfs, loop-mount it, exec a binary from it (fine), overwrite the backing file (still fine — pages cached), then `drop_caches` and exec again → `Input/output error`. That cached-vs-evicted distinction is why it hits 64 MB cameras and never reproduces on a 754 MB one.

### On the ubifs half — **not** addressed

Being straight about this so nobody assumes it came for free: #2252 deliberately leaves the old root **still mounted at `/mnt`** after pivoting, because the MTD partitions are reached through `/dev` and nothing needed it unmounted. The flash path is still `flashcp` against an MTD partition.

So the groundwork this issue asks for is there — after the pivot nothing depends on the old rootfs, which is the hard prerequisite — but a ubifs flow would still need the extra step of actually unmounting `/mnt` (and a ubi-aware write path instead of `flashcp`). That is a smaller change on top of #2252 than it was before it.

[...]

PR #2252, opened 2026-08-10
sysupgrade: flash from a ramfs, so the reboot cannot be eaten by the flash

widgetii:
A WebUI upgrade could leave a camera needing a **physical power cycle**: the log froze, the box stopped answering, and after the power cycle it was running the *new* firmware. So the flash completed and the reboot did not.

Reported in OpenIPC/majestic-webui#120 (gk7205v300, 64 MB). Distinct from the false-success bug fixed in #2249.

### Mechanism

`flashcp` rewrites the MTD partition backing the mounted squashfs rootfs, and that write goes to the raw `/dev/mtdN` character device — behind the block layer, so the page cache is never invalidated. The running system keeps working for exactly as long as the pages it needs stay **resident**; anything faulted back in afterwards reads the new image at a stale offset. `free_resources()` has already dropped the cache to make room for the download, so on a 34 MB camera eviction is likely rather than theoretical — and on a 754 MB one it never happens, which is why this reproduces for some users and never in the lab.

What it kills is the reboot: `busybox reboot` is a fork+exec of a binary living on the partition just overwritten, so it fails with EIO and nothing reboots.

Same mechanism @widgetii documented for the `-x` path in #2231; #2236 forced a reboot there, curing that symptom without removing the hazard.

**Reproduced deterministically with no MTD writes** — copy the rootfs squashfs to tmpfs, loop-mount it, rewrite the backing file:

```
exec from loop-squashfs         -> OK
overwrite backing store         -> (what flashcp does to mtd3)
exec again, pages still cached  -> OK        <- why it looks fine on a big camera
drop_caches, exec again         -> Input/output error, rc=126
```

### Fix

Move the flashing phase into RAM, the way OpenWrt's sysupgrade does. `enter_ramfs` stages busybox, its loader and libraries and the script into a tmpfs, moves `/dev`, `/proc` and `/tmp` across, `pivot_root`s and **re-execs**.

The re-exec is the part that matters: `pivot_root` changes what paths resolve to, but the running shell keeps its text mapped from the old squashfs inode — a pivot without one is theatre.

Failure is backwards-compatible throughout: any step that does not work returns and the caller flashes in place exactly as before. `--no_ramfs` forces that path for board bring-up.

### Three things only hardware found

- busybox dispatches on `argv[0]`, so the first run **pivoted perfectly and then could not find `od`, `cut`, `grep` or `flashcp`**. Applet symlinks are not optional, and the pivot is now refused outright if `sh`/`flashcp`/`reboot` are missing from the staged root.
- A `die()` after the pivot must reboot whatever it was doing: the old system is behind `/mnt`, so exiting there leaves exactly the corpse this change exists to prevent.
- `_ramfs_phase` must be exported **only after the pivot succeeds**. Setting it earlier leaks into the fallback, where it tells `die()` to reboot on a failure that never touched flash — reintroducing the false-success #2249 had just removed. Caught by the offline suite.

### Also: drop the `--web` pre-flash `killall -q -3 majestic`

`--web` exists so majestic can keep streaming the log; the daemon has already released its video pipeline by that point; and SIGQUIT to a majestic already in upgrade mode is unsafe, and is fixed in the daemon. The log survives the pivot anyway — `tail`'s fd and the websocket are open file descriptors, and moving a mount does not disturb them.

### Verified on two boards

| | gk7205v200 | hi3516av300 |
|---|---|---|
| RAM / tmpfs | 34 MB / 17 MB | 754 MB / 377 MB |
| libc | **musl** (loader *is* libc) | glibc (loader + libc + libresolv) |
| staged root | 1.1 MB of 8 MB | 2.3 MB of 8 MB |

On both: after the pivot `/` is tmpfs, busybox is on tmpfs, the old root is at `/mnt`, `/dev/mtd3`, `/proc/mtd` and the staged image are reachable, every tool the flash phase uses resolves and runs, and dropping the page cache changes nothing. Both rebooted from RAM and came back clean with video serving.

6 new behaviour cases and 12 new invariants; **82 checks pass**.

### What is not covered

The real pivot cannot be exercised by the offline suite (it needs root and real mounts), so it rests on source invariants plus the hardware runs above. The **fallback** — a camera that cannot build a ramfs still upgrades — *is* covered behaviourally, since the stub fails by default.

No test has yet let a real `flashcp` write a real rootfs from inside the ramfs; every run stops short of the write by design. Closing that needs a board you are willing to reflash. If something does surprise us there, the fallback is unchanged, so the downside is a camera that flashes the old way rather than one that bricks.

Supersedes #2251, which carried an earlier half-measure (staging busybox in /tmp plus a sysrq reboot) alongside this one. That approach is removed: it was superseded by the pivot, and hardware confirms `busybox reboot` works fine from inside the ramfs on both boards.

### DIFF
--- a/.github/scripts/test_sysupgrade.sh
+++ b/.github/scripts/test_sysupgrade.sh
@@ -64,6 +64,7 @@ sed -e 's|grep "GITHUB_VERSION" "$1/etc/os-release"|grep "GITHUB_VERSION" "${1:-
     -e "s|/proc/cmdline|@SB@/proc/cmdline|g" \
     -e "s|/proc/sys/vm/drop_caches|@SB@/tmp/drop_caches|g" \
     -e "s|/etc/init.d/|@SB@/etc/init.d/|g" \
+    -e "s|/bin/busybox|@SB@/bin/busybox|g" \
     -e "s|@OSRELEASE@|/etc/os-release|g" \
     -e "s|@SB@|$SB|g" \
     "$SRC" > "$SB/sysupgrade"
@@ -110,6 +111,10 @@ stub killall    'exit 0'
 stub ntpd       'exit 0'
 stub curl       'exit 0'
 stub umount     'exit 0'
+# Logs so ordering can be asserted, and fails by default: an unprivileged test
+# host cannot really pivot, and the fallback is the safety property that matters
+# most here (a camera that cannot build a ramfs must still upgrade).
+stub pivot_root 'echo "pivot_root $*" >> "$FLASH_LOG"; exit ${STUB_PIVOT_RC:-1}'
 stub losetup    'case "$1" in -f) echo /dev/loop0;; *) exit 0;; esac'
 
 # download_firmware runs `md5sum -s -c`. -s (silent) is a busybox extension; GNU
@@ -150,6 +155,15 @@ exit 0'
 stub mount '
 [ $# -eq 0 ] && exit 0
 target=${!#}
+# STUB_MOUNT models the VERIFY-mount (a squashfs image over a loop device). The
+# ramfs pivot mounts tmpfs and relocates existing mounts; those are a different
+# operation and must not inherit the fault being injected, or the hang case
+# wedges before the code under test is even reached.
+for a in "$@"; do
+    case "$a" in
+        move|tmpfs|--move|-M) exit 0 ;;
+    esac
+done
 case "${STUB_MOUNT:-ok}" in
     ok)
         mkdir -p "$target/etc"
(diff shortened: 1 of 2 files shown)


widgetii:
Links to **#239** (open since 2022-06-20), which asked for exactly this mechanism:

> Делать pivot_root и chroot при выполнении sysupgrade. Это позволит обновлять устройства с ubifs, а также может помочь от единичных глюков при обновлении

Two halves to that request, and this PR only answers one and a half:

- **pivot_root + chroot in sysupgrade** — done here.
- **"единичные глюки при обновлении"** — one concrete instance is now diagnosed and fixed: the reboot being eaten by the flash, which is what this PR is about.
- **ubifs devices** — *not* addressed. This deliberately leaves the old root mounted at `/mnt` (the MTD partitions are reached through `/dev`, so nothing needed it unmounted), and the write path is still `flashcp` against an MTD partition. A ubifs flow would need the additional unmount plus a ubi-aware write — smaller on top of this than before it, but not free.

Flagging that explicitly so #239 isn't closed on the assumption this covers all of it.

PR #2254, opened 2026-08-11
ci: run scheduled workflows only on the canonical repo

liam1472:
### The problem

A clone pushed to a **brand-new repository** — a private mirror, an internal re-host — inherits `.github/workflows` along with their cron schedules. From that moment the nightly builds the full ~96-platform matrix every night on that owner's Actions bill, publishing nightlies nobody will ever read.

Forks are already covered: GitHub disables Actions on forks by default and disables scheduled workflows there specifically. A **fresh repository created from a clone is not covered**, and nothing warns the owner — the README doesn't mention CI, and the first sign is the usage page.

I hit this on a private mirror of this repo. One night, unattended:

- **97 jobs, 2,146 billed minutes**, from a single scheduled run
- `Publish releases` failed there, because the mirror had no `nightly` release to write to
- and that is the nasty part: `preflight` decides whether to build by comparing `HEAD` against the sha recorded in the published `nightly` release. With publishing broken, `PREV` stays empty, so `PREV != HEAD` holds forever and **the skip gate can never engage**. Left alone it would have rebuilt the entire matrix every single night — roughly 64k minutes a month — while never publishing anything.

Nothing here is a bug in the pipeline itself; it is a pipeline that belongs to the canonical repo running somewhere it was never meant to run.

### The change

Gate the unattended entry points on `github.repository`, which Actions sets per **running** repository (`owner/repo`), so the comparison is false in any mirror and true upstream:

```yaml
if: github.repository == 'OpenIPC/firmware' || github.event_name == 'workflow_dispatch'
```

Applied to the jobs that start automatically:

| Workflow | Trigger gated |
|---|---|
| `build.yml` | `preflight` — the whole matrix hangs off it via `needs:` |
| `cleanup.yml` | weekly prune cron |
| `gcc-compat.yml` | weekly compat cron |
| `image.yml` | `workflow_run` |
| `manifest.yml` | `workflow_run` |

`image` and `manifest` need it too: a build whose jobs all skip still *completes*, which fires their `workflow_run` trigger.

### What is preserved

- **Upstream behaviour is unchanged.** `github.repository` is `OpenIPC/firmware` for scheduled runs, and for `pull_request` events it is the base repo, so PR builds here still run.
- **`workflow_dispatch` keeps working everywhere**, including in mirrors — an operator who deliberately clicks Run is not blocked. `build-one.yml` is dispatch-only and untouched.
- No change to build logic, matrix contents, or publishing.

Happy to narrow this to `build.yml` alone if you would rather keep the diff minimal, or to switch the literal to a `vars.CANONICAL_REPO` variable if you prefer that over a hardcoded name.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,13 +10,17 @@ on:
 jobs:
   preflight:
     name: Preflight
-    # Only the canonical repo builds unattended. A clone pushed to a new
+    # Only the canonical repo builds on the cron. A clone pushed to a new
     # repository (a private mirror, an internal re-host) inherits this file
-    # along with its cron, and would run the full platform matrix nightly on
-    # someone else's Actions bill while publishing nightlies nobody reads.
+    # along with its schedule, and would run the full platform matrix nightly
+    # on someone else's Actions bill while publishing nightlies nobody reads.
     # Forks are already covered — GitHub disables scheduled workflows there —
-    # but a fresh repository is not. Manual dispatch still works everywhere.
-    if: github.repository == 'OpenIPC/firmware' || github.event_name == 'workflow_dispatch'
+    # but a fresh repository is not.
+    #
+    # Scoped to the schedule event alone: pull_request and workflow_dispatch
+    # are somebody deliberately asking for a build, and keep working here and
+    # in every mirror.
+    if: github.event_name != 'schedule' || github.repository == 'OpenIPC/firmware'
     runs-on: ubuntu-latest
     outputs:
       should_build: ${{ steps.gate.outputs.should_build }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,13 @@ on:
 jobs:
   preflight:
     name: Preflight
+    # Only the canonical repo builds unattended. A clone pushed to a new
+    # repository (a private mirror, an internal re-host) inherits this file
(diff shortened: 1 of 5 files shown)


liam1472:
Please read this PR carefully such a silly build process triggering, Dev clone and build on demand, doesn't need to inherit ur massive build process no guard provided to general dev when they push this up to their repo they immediately paid a massive amount of money for those artifact they not even use. This is correct for OpenIPC maintainer but general dev build for their target hardware only.

liam1472:
Good catch — that was a real over-block, fixed in 45e5c52e.

The first version admitted only the canonical repo or a manual dispatch, which also took out `pull_request` in any non-canonical repo. That contradicts the intent stated in the comments: a PR is somebody deliberately asking for a build, not an unattended cron run, and a fork owner who has gone out of their way to enable Actions should still get PR CI.

Now gated on the event itself:

```yaml
if: github.event_name != 'schedule' || github.repository == 'OpenIPC/firmware'
```

`image` and `manifest` key off `github.event.workflow_run.event` for the same reason, since their unattended path is a cron-originated build.

| event | repo | result |
|---|---|---|
| `schedule` | `OpenIPC/firmware` | runs |
| `schedule` | any mirror | **skips** |
| `pull_request` | `OpenIPC/firmware` | runs |
| `pull_request` | any mirror | runs |
| `workflow_dispatch` | anywhere | runs |

Only the cron is touched, and only outside this repo.

PR #2255, opened 2026-08-11
ci: keep inherited workflows from billing non-canonical repos

liam1472:
Supersedes #2254 (closed) — same goal, final logic.

### Problem

Push a clone of this repo to a new repository (private mirror / internal re-host) and `.github/workflows` comes along with its cron. The full ~96-platform matrix then runs **on that owner's Actions bill** — nightly, unattended, publishing nightlies nobody reads. Forks are only partially protected: GitHub disables *scheduled* workflows on forks, but a fresh repository gets no protection and no warning.

Measured on one private mirror, from a single night: **97 jobs, 2,146 billed minutes**. And it never converges: the mirror's `Publish releases` fails (no `nightly` release exists there), so `preflight`'s sha skip-gate never engages and the full matrix rebuilds **every** night — ~64k minutes/month for nothing.

### The rule this implements

**Unattended runs happen only on the canonical repo.** Anywhere else a workflow runs only if someone explicitly asked for it (`workflow_dispatch`), or if it is both free and attended (`pull_request` on a public repo).

```yaml
if: >-
  github.repository == 'OpenIPC/firmware' ||
  github.event_name == 'workflow_dispatch' ||
  (github.event_name == 'pull_request' && !github.event.repository.private)
```

This is deliberately stricter than "gate it where the minutes stop being free". A public re-host's minutes *are* free, and its cron is still skipped: a re-host quietly publishing `nightly` and `latest` releases under its own name is a worse outcome than wasted minutes, so the unattended path is cut there too. Only the canonical repo runs on a schedule.

| event | where | result |
|---|---|---|
| `schedule` | OpenIPC/firmware | runs (unchanged) |
| `pull_request` | OpenIPC/firmware | runs (unchanged — contributor PRs execute in this repo's context) |
| `push: master` | OpenIPC/firmware | runs (unchanged) |
| `workflow_dispatch` | anywhere | runs (an operator explicitly asked; `build-one.yml` untouched) |
| `pull_request` | public fork / re-host | runs (free and attended; forks of a public repo are public by construction) |
| `schedule` | public re-host | **skips** (deliberate — see above) |
| `schedule` | private mirror | **skips** |
| `pull_request` | private mirror | **skips** (the only place PR CI costs a third party money) |
| `push: master` | any mirror | **skips** |

### Coverage

`build.yml` and `gcc-compat.yml` carry the guard on their entry jobs. `image.yml` / `manifest.yml` gate their cron-originated `workflow_run` path the same way — a nightly whose jobs all skipped still completes and fires `workflow_run`. `cleanup.yml` carries it on its weekly cron. `shell-tests.yml` carries it too: it triggers on `pull_request` and `push: [master]`, which made it the last billed entry point a clone inherits. `toolchain`, `uboot` and `build-one` are dispatch-only and need nothing.

`ci-gate` carries the guard as well, keeping `always()`. Without that, skipping `preflight` did not skip the gate: it ran anyway, read `preflight=skipped` and exited 1, so a guarded run came out **red** rather than skipped — a red nightly every day and a red required check on every internal PR. Guarding the job rather than teaching the gate script to tolerate `preflight=skipped` keeps the gate strict upstream, which is the one place it has to be.

A guarded run now skips whole, start to finish, and reports nothing.

### Upstream impact

None. On `OpenIPC/firmware` the first disjunct is true, so every added condition is a tautology and each job's `if` reduces to what it was before this PR — `ci-gate`'s reduces to plain `always()`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -545,7 +545,14 @@ jobs:
   ci-gate:
     name: CI Gate
     needs: [preflight, buildroot, publish]
-    if: always()
+    # always() so the gate still reports when a needed job fails -- but not on a
+    # run whose preflight the guard above deliberately skipped, where it would
+    # read preflight=skipped and turn a saved build into a red run. Upstream the
+    # added disjunct is true, so this reduces to always().
+    if: >-
+      always() && (github.repository == 'OpenIPC/firmware' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && !github.event.repository.private))
     runs-on: ubuntu-latest
     steps:
       - name: Require preflight + firmware matrix to succeed
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,24 @@ on:
 jobs:
   preflight:
     name: Preflight
+    # Guard rail for clones of this repo. A clone pushed to a new repository
+    # (a private mirror, an internal re-host) inherits this file along with
+    # its cron, and the full platform matrix — ~2000 runner-minutes a pass —
+    # then runs on that owner's Actions bill: nightly via the schedule, and
+    # again on every internal PR they open. Forks are partially covered
+    # (GitHub disables scheduled workflows there), a fresh repository is not.
+    #
+    # The line is drawn where the minutes stop being free:
+    #   - canonical repo: everything runs, unchanged
+    #   - workflow_dispatch: an operator explicitly asked; runs anywhere
+    #   - pull_request on a PUBLIC repo: free minutes (contributor PRs here
+    #     run in this repo's context; within-fork PRs are public by
+    #     construction) — runs
+    #   - anything unattended or billed on a non-canonical repo — skips
+    if: >-
+      github.repository == 'OpenIPC/firmware' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && !github.event.repository.private)
     runs-on: ubuntu-latest
     outputs:
       should_build: ${{ steps.gate.outputs.should_build }}
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -15,6 +15,9 @@ concurrency:
 jobs:
   prune:
     name: Prune old nightly releases
+    # Canonical repo only on the weekly cron — see build.yml's preflight.
+    # Dispatch runs are unaffected.
+    if: github.event_name != 'schedule' || github.repository == 'OpenIPC/firmware'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/gcc-compat.yml
+++ b/.github/workflows/gcc-compat.yml
@@ -15,6 +15,12 @@ on:
 jobs:
   gcc-compat:
     name: GCC ${{matrix.gcc}}
+    # Same guard rail as build.yml's preflight: run where the minutes are
+    # free or explicitly asked for, skip unattended/billed mirror runs.
+    if: >-
+      github.repository == 'OpenIPC/firmware' ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'pull_request' && !github.event.repository.private)
     runs-on: ubuntu-latest
     container:
       image: gcc:${{matrix.gcc}}
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -24,11 +24,18 @@ jobs:
     # A flaky board doesn't deny image assembly for the platforms that did
     # upload a firmware tgz — firmware_url returns empty for missing ones and
     # create() skips them.
+    #
+    # First clause: on a mirror the nightly's jobs skip, but the run still
+    # completes and fires workflow_run, so gate the cron-originated path on
+    # the canonical repo too — see build.yml's preflight. Dispatch-driven
+    # reassembly is untouched anywhere.
     if: >-
-      github.event_name == 'workflow_dispatch' ||
-      ((github.event.workflow_run.event == 'schedule' ||
-        github.event.workflow_run.event == 'workflow_dispatch') &&
-       contains(fromJSON('["success", "failure"]'), github.event.workflow_run.conclusion))
+      (github.event.workflow_run.event != 'schedule' ||
+       github.repository == 'OpenIPC/firmware') &&
+      (github.event_name == 'workflow_dispatch' ||
+       ((github.event.workflow_run.event == 'schedule' ||
+         github.event.workflow_run.event == 'workflow_dispatch') &&
+        contains(fromJSON('["success", "failure"]'), github.event.workflow_run.conclusion)))
     env:
       SIGMASTAR: ssc30kd ssc30kq ssc325 ssc333 ssc335 ssc335de ssc337 ssc337de ssc338q ssc377 ssc377d ssc377de ssc377qe ssc378de ssc378qe
       INGENIC: t10 t10l t20 t20l t20x t21n t23n t30a t30a1 t30l t30n t30x t31a t31al t31l t31lc t31n t31x
--- a/.github/workflows/manifest.yml
+++ b/.github/workflows/manifest.yml
@@ -24,11 +24,18 @@ jobs:
     # platforms that did upload artifacts. enrich_manifest.py reads whatever
     # assets actually exist on the release, so a partial release just shows
     # up with fewer platforms in its `platforms` map.
+    #
+    # First clause: on a mirror the nightly's jobs skip, but the run still
+    # completes and fires workflow_run, so gate the cron-originated path on
+    # the canonical repo too — see build.yml's preflight. Dispatch-driven
+    # manifest regeneration is untouched anywhere.
     if: >-
-      github.event_name == 'workflow_dispatch' ||
-      ((github.event.workflow_run.event == 'schedule' ||
-        github.event.workflow_run.event == 'workflow_dispatch') &&
-       contains(fromJSON('["success", "failure"]'), github.event.workflow_run.conclusion))
+      (github.event.workflow_run.event != 'schedule' ||
+       github.repository == 'OpenIPC/firmware') &&
+      (github.event_name == 'workflow_dispatch' ||
+       ((github.event.workflow_run.event == 'schedule' ||
+         github.event.workflow_run.event == 'workflow_dispatch') &&
+        contains(fromJSON('["success", "failure"]'), github.event.workflow_run.conclusion)))
     runs-on: ubuntu-latest
     steps:
       - name: Checkout master (for the script)
--- a/.github/workflows/shell-tests.yml
+++ b/.github/workflows/shell-tests.yml
@@ -9,8 +9,16 @@ on:
   workflow_dispatch:
 
 jobs:
+  # Same guard build.yml and gcc-compat.yml carry. These two jobs are cheap, but
+  # they are the last billed entry point a clone inherits: on a private mirror
+  # every master sync push and every internal PR runs them on that owner's bill.
+  # Upstream the first disjunct is true, so the condition is a tautology here.
   load-hisilicon-parsing:
     name: load_hisilicon os_mem_size derivation
+    if: >-
+      github.repository == 'OpenIPC/firmware' ||
+      github.event_name == 'workflow_dispatch' ||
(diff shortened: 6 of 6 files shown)


liam1472:
Fair point on the duplication — here is the reasoning for keeping the literal, and two alternatives if the maintainers would rather not have it.

Job-level `if` cannot read `env`, so there is no way to define the name once per workflow; the realistic options are:

1. **Literal ×5** (current). All five hits live in one directory and one grep finds them. If the canonical location ever changes, these strings would be part of the same sweep as every badge, doc and release URL — and a stale guard fails *loud* upstream (nightly stops) rather than silently costing anyone money.
2. **`vars.CANONICAL_REPO`** — single source, but if the variable is ever unset/mistyped the upstream nightly silently stops running, which is a worse failure mode than the one this PR fixes. A fallback (`vars.CANONICAL_REPO || 'OpenIPC/firmware'`) avoids that but then the literal is back in all five files plus an extra layer of indirection.
3. **`github.repository_id == 357631421`** — the repo id survives both rename and transfer, so the scenario the review describes ("canonical location changes") requires zero edits, ever. Cost: a magic number that needs a comment carrying the human-readable name.

I kept (1) as the least surprising, but (3) is strictly more robust against exactly this concern — happy to switch the five guards to it if preferred.

liam1472:
@widgetii  sorry for the direct ping — flagging this one since it concerns the CI files you maintain: inherited cron in cloned repos silently bills the owner (~2,146 min/night measured). Small guard, upstream behaviour untouched. Happy to adjust if you'd rather draw the line differently.

widgetii (commented):
The change is sound in principle, and I verified the "upstream behaviour byte-for-byte unchanged" claim actually holds. One gap contradicts the PR's own table, though.

### Verified correct

- **`build.yml` / `gcc-compat.yml`** — the guard's first disjunct is true on `OpenIPC/firmware`, so the job condition is a tautology here.
- **`image.yml` / `manifest.yml`** — the new first clause is `wr.event != 'schedule' || repo == canonical`; the second disjunct is always true upstream, so the whole expression reduces to the pre-PR condition. Checked by case analysis over all five reachable trigger paths.
- Operator dispatch of `image`/`manifest` themselves still works: `github.event.workflow_run` is null, and `null != 'schedule'` is true in Actions expressions, so the new clause doesn't swallow it.
- Guarding the `github.event.repository.private` lookup behind `event_name == 'pull_request'` is the right defensive call — it avoids depending on whether the `schedule` payload carries `repository` at all. And on `pull_request`, `github.event.repository` is the *base* repo, the same repo as `github.repository`, so the two clauses are consistent (no head/base confusion).
- All five files parse.

---

### 1. Blocking — `ci-gate` still runs, and hard-fails, on every mirror run

`.github/workflows/build.yml` L545-551 is `if: always()`, so skipping `preflight` does not skip it:

```
preflight  -> skipped (new guard)
buildroot  -> skipped (needs preflight)
publish    -> skipped
ci-gate    -> RUNS (always()), reads needs.preflight.result == 'skipped'
              -> "::error::preflight did not succeed" -> exit 1
```

So on a private mirror the table is wrong in two rows: `schedule` and `pull_request` don't skip — they produce a **failed** workflow run. Every night at 22:30 UTC the mirror owner gets a red nightly plus a failure notification, and every internal PR gets a red `CI Gate`.

The billing goal still lands (~2,146 min -> ~1 min), so this isn't fatal to the premise. But "publishing nightlies nobody reads" becomes "failing nightlies nobody reads", which is most of the noise the PR set out to remove.

Fix keeps upstream identical (`always() && true` is `always()`):

```yaml
  ci-gate:
    name: CI Gate
    needs: [preflight, buildroot, publish]
    if: >-
      always() && (github.repository == 'OpenIPC/firmware' ||
      github.event_name == 'workflow_dispatch' ||
      (github.event_name == 'pull_request' && !github.event.repository.private))
```

Worth doing it this way rather than teaching the gate script to tolerate `preflight=skipped` — that would silently weaken the gate upstream, which is the one place it has to stay strict.

### 2. `shell-tests.yml` is the one entry point left unguarded

It triggers on `pull_request` **and** `push: [master]`, with no `if`. On a private mirror, every master sync push and every internal PR runs two jobs. It's ~1 minute, not the money — but it's the same class, and it's the only billed entry point left (`toolchain`, `uboot`, `build-one` are all dispatch-only). Either guard it, or add a line to the commit message explaining the exemption the way `build-one` is called out.

### 3. Follow-up, not this PR — `OpenIPC/builder` has identical exposure

`OpenIPC/builder`'s `.github/workflows/` carries the same preflight/cleanup/manifest design with its own daily cron. A clone of *builder* reproduces this exact problem. If this lands, the guard wants mirroring there with `OpenIPC/builder` as the literal — same pattern as the size-report mirror (builder#102).

### Minor

- **On the `repository_id` alternative** from the thread above: I confirmed `357631421` is correct for this repo. I'd still keep the literal — a rename would already require sweeping badges, docs and release URLs, and a stale guard fails loud upstream rather than quietly costing someone money. If you do switch, quote it (`== '357631421'`); `github.repository_id` is a string, and the unquoted form leans on Actions' implicit string-to-number coercion.
- **Body framing**: a *public* re-host is also skipped on cron, even though its minutes are free — the implementation is stricter than the stated "where the minutes stop being free" principle. The stricter behaviour is the right call (a re-host publishing `nightly`/`latest` releases is worse than wasted minutes), but the description should say that rather than state a principle the code doesn't follow.
- **Optional readability**, `image.yml`/`manifest.yml` are now three paren levels deep. This is equivalent, and reads as one branch per legitimate path:

```yaml
  if: >-
    github.event_name == 'workflow_dispatch' ||
    (contains(fromJSON('["success", "failure"]'), github.event.workflow_run.conclusion) &&
     (github.event.workflow_run.event == 'workflow_dispatch' ||
      (github.event.workflow_run.event == 'schedule' && github.repository == 'OpenIPC/firmware')))
```

---

Guard `ci-gate` and I'm happy with this. Everything else is optional.

liam1472:
Thanks — the `ci-gate` catch is right, and it is why the table in the description was wrong. `always()` kept the gate running on exactly the runs the new guard had just made cheap, so a private mirror got a red nightly and a red required check on every internal PR instead of a silent skip.

Fixed in bece24eb by giving `ci-gate` the same guard `preflight` carries, so the two can never disagree — there is no longer a run where `preflight` is skipped but the gate still evaluates it. `always()` is kept, so the gate still reports when a needed job genuinely fails. Upstream the added disjunct is true, so the condition reduces to `always()` and nothing changes on `OpenIPC/firmware`.

I took your reasoning for putting it in the `if` rather than teaching the gate script to tolerate `preflight=skipped`: upstream is the one place the gate has to stay strict. With this, the `schedule` and `pull_request` rows of the table are now accurate for a private mirror — the whole run is skipped, not failed.

On the rest: happy to fold in the `shell-tests.yml` guard and your flatter `image.yml`/`manifest.yml` expression (I checked it agrees with the current one on every trigger path, including `conclusion == 'cancelled'` and direct dispatch where `workflow_run` is null), and to reword the description so it states the stricter rule the code actually implements rather than the weaker "where the minutes stop being free" — a public re-host is skipped on cron too, deliberately. Say the word if you want any of those in this PR or left out of it.

Agreed on keeping the `OpenIPC/builder` mirror as a separate follow-up once this lands, and on keeping the repository literal over `repository_id`.

widgetii:
`bece24eb` is the right shape — verified the parenthesised body is character-for-character `preflight`'s guard, so the two conditions can't drift apart into disagreeing. Re-traced the matrix on the new head: private mirror skips whole on both `schedule` and `pull_request`, public re-host PR and dispatch-anywhere still run, and upstream the added disjunct is a tautology so it collapses back to plain `always()`. Blocking issue closed.

Also confirmed your equivalence check on the flatter `image`/`manifest` expression holds for `conclusion == 'cancelled'` and for direct dispatch where `workflow_run` is null — the dispatch disjunct short-circuits before the null deref.

On the three you offered:

- **`shell-tests.yml` — yes, please fold in.** It's the last billed entry point (`push: [master]` plus `pull_request`), and without it the commit message's coverage claim isn't quite true. Minutes are negligible; consistency is the point.
- **Description reword — yes.** State the rule the code actually implements, including that a public re-host is skipped on cron deliberately.
- **Flatter `image`/`manifest` expression — leave it out.** The current one is correct, just dense, and re-touching it buys another equivalence review for no behavioural change. Not worth the churn in a PR that's otherwise ready.

Agreed on `OpenIPC/builder` as a follow-up once this lands, and on keeping the literal.

liam1472:
Both folded in, and the flatter `image`/`manifest` expression left alone — agreed it buys another equivalence review for no behavioural change.

**`shell-tests.yml`** (6cdd400b) — guarded, same condition as `preflight` character for character, on both jobs. That closes the gap: `toolchain`, `uboot` and `build-one` are dispatch-only, so `push: [master]` and `pull_request` here were the last billed entry points a clone inherits, and the coverage claim now holds as written.

**Description** — rewritten around the rule the code actually implements: *unattended runs happen only on the canonical repo; anywhere else a workflow runs only if someone explicitly asked for it, or if it is both free and attended.* It now says outright that a public re-host is skipped on cron even though its minutes are free, and why — a re-host quietly publishing `nightly`/`latest` under its own name is a worse outcome than wasted minutes. The table gained the `push: master` and public-re-host-on-cron rows that were missing, and there is a short coverage section listing which workflows carry the guard and which are dispatch-only. The `ci-gate` reasoning is written up there too, since a reader of the merged commit will want to know why the gate is guarded rather than made tolerant of `preflight=skipped`.

Thanks for the re-trace on `bece24eb` and for confirming the `cancelled` / null-`workflow_run` cases independently — good to have that checked from both sides rather than taken on my word.

Ready from my side. I will open the `OpenIPC/builder` follow-up once this lands, with `OpenIPC/builder` as the literal.

widgetii (approved):
Approved. Verified on `6cdd400b8`: the guard is character-for-character identical across `preflight`, `ci-gate` (wrapped in `always() && (...)`), `gcc-compat` and both `shell-tests` jobs, so they cannot drift into disagreeing. Full-tree audit confirms every job with a non-dispatch trigger is now guarded, and the four unguarded jobs live in dispatch-only workflows.

CI run 31576719736 confirms the guards don't over-block: this is a fork PR, and `Preflight`, all 96 `Firmware (<board>)` jobs, `CI Gate`, both `shell-tests` jobs and all five `gcc-compat` jobs executed normally. `Publish releases` skipped as it always has on PRs. Upstream behaviour is unchanged, as claimed.

Worth stating plainly for the record: this run proves the guards don't block anything here, but it cannot exercise the skip path — on `OpenIPC/firmware` every added condition is a tautology, so the mirror behaviour is verified by reasoning only, from both sides independently.

Thanks for the thorough iteration on this one.

liam1472:
No walker mate

widgetii:
@liam1472 sorry — "No walker mate" lost me. Was that meant for this thread, or did it land here by mistake? Happy to pick it up either way, just want to make sure nothing's getting missed.

widgetii:
@liam1472 heads-up so you don't duplicate work: I went ahead and did the `OpenIPC/builder` mirror rather than leaving it with you — OpenIPC/builder#118, merged as `222db9e`. Apologies if you already had it in flight locally.

It is not a straight copy; builder's trigger set differs in two ways worth recording:

- **No `pull_request` trigger anywhere in that repo**, so the third clause of the guard is unreachable there today. I kept it deliberately and said so in the comment — it is the clause that lets contributor PRs run, and dropping it would mean that adding a `pull_request` trigger later silently skips CI on every fork. That is exactly the over-block caught on #2254, and it seemed cheaper to carry a correct-if-reached disjunct than to re-derive it later under pressure.
- **No `ci-gate` equivalent**, so the `always()` trap you fixed in `bece24eb` does not exist there. I documented it in the PR body anyway, so whoever adds a gate job to builder knows it needs `always() && (...)` rather than a bare `always()`.

Also: `manifest.yml` there had a *weaker* pre-existing condition than the one here — it never checked `workflow_run.event` at all, because builder's `Build` only fires on schedule/dispatch, so the check was redundant. The schedule gate is added without disturbing that.

Same caveat as here, and worth stating plainly: builder has no PR CI, so nothing validated that change either. On `OpenIPC/builder` every added condition is a tautology and the skip path is unreachable by construction. The real check is that its next nightly (03:00 UTC) still runs — if a guard were malformed, the failure mode is a silent skip and no released nightly, not a red run. I will be watching for that one.

Thanks again for the work on this — the `ci-gate` catch was the substantive one and it came from you acting on the review rather than defending the original.

PR #2256, opened 2026-08-11
gk7102: rtl8xxxu Wi-Fi + -mno-unaligned-access

shipa-2:
## Summary
- Add `rtl8xxxu-openipc` (RTL8188FTV / 0bda:f179) with `RCR_APPEND_FCS` so RX FCS stripping does not corrupt EAPOL
- Add `rtl8188fufw.bin` to `linux-firmware-openipc`
- Enable on `gk7102_lite` / `gk7102s_lite`: overlay `rtl8xxxu-gk7102` (GPIO9), WPA crypto options, `BR2_TARGET_OPTIMIZATION="-mno-unaligned-access"`
- Fix `mbedtls-openipc` configure on newer host CMake (`CMAKE_POLICY_VERSION_MINIMUM=3.5`)

## Toolchain note
`BR2_TARGET_OPTIMIZATION` alone is not enough while packages use the prebuilt external SDK: rebuild/publish `toolchain.goke-gk710x.tgz` with the same flag (`make BOARD=gk7102_lite toolchain`, or temporarily comment out `BR2_TOOLCHAIN_EXTERNAL*`), and drop musl `src/string/arm/memcpy.S` so libc does not use unaligned LDR/STR. With that SDK, Wi-Fi + SSH work without wpa/Dropbear alignment workarounds.

## Test plan
- [x] Flash `gk7102_lite` uImage + rootfs
- [x] `wlandev=rtl8xxxu-gk7102`, WPA2 connect, DHCP, outbound network
- [x] SSH login without Dropbear byte-copy workaround

Made with [Cursor](https://cursor.com)

PR #2257, opened 2026-08-11
mbedtls-openipc: CMAKE_POLICY_VERSION_MINIMUM=3.5

shipa-2:
## Summary
- Set `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` so `mbedtls-openipc` configures on newer host CMake.

## Test plan
- [ ] `make BOARD=gk7102_lite` (or any board using mbedtls-openipc) configures mbedtls without CMake policy errors

Part of splitting #2256.

Made with [Cursor](https://cursor.com)


widgetii (commented):
Thanks for splitting this out of #2256 — the diagnosis is right (mbedtls 2.25.0 really does die on CMake 4.x), but I believe this change is **redundant**: master already fixes CMake 4.x policy compat globally, for every package at once.

## Already fixed in `Makefile:7`

```makefile
export CMAKE_POLICY_VERSION_MINIMUM := 3.5
```

Added in 656094d5a — *"Fix build with modern host toolchains (GCC 15, CMake 4.x)"* (#1978). That line is present on `origin/master` and on this PR's own branch.

Three things I checked rather than assumed, with host cmake 4.3.3:

1. **The failure is real.** Configuring `mbedtls-openipc-2.25.0` (`cmake_minimum_required(VERSION 2.8.12)`) bare → `Compatibility with CMake < 3.5 has been removed from CMake.` Hard error, as you say.
2. **The env var alone is sufficient.** Same tree, `CMAKE_POLICY_VERSION_MINIMUM=3.5 cmake -S … -B …` with no `-D` → `Build files have been written to`. CMake honours it as an environment variable, not only as a `-D` cache entry.
3. **The export actually reaches cmake.** Buildroot's `pkg-cmake.mk` `CONFIGURE_CMDS` sets `PATH` and the `PKG_CONFIG_*` vars but never scrubs the environment (no `env -i`), so the top-level export propagates into every target *and* host cmake invocation.

So the `-D` here duplicates a value cmake already has. Harmless, but not load-bearing.

## It would also be incomplete on its own

Even without the global export, patching mbedtls alone wouldn't get a board to build. Packages using `cmake-package` whose sources declare min < 3.5:

| package | `cmake_minimum_required` | boards enabling |
|---|---|---|
| `mbedtls-openipc` 2.25.0 | 2.8.12 | **93** ← this PR |
| `libwebsockets-openipc` 4.2.2 | 2.8.12 | **18** ← would fail next |
| `mini` / `aura-httpd` / `usrsctp` | 2.8.12 / 2.8 / 3.0 | 0 (dormant) |

Clean: `aws-webrtc` 3.6.3, `libsrtp` 3.21, `baresip`/`libre` 3.14, `fdk-aac` 3.5.1, `yaml-cli` 3.5, `ipctool`/`jsonfilter` 3.10. `lzo` (3.0) and `libubox` (2.6) are sub-3.5 cmake packages but enabled on no board. `libzlib`/`libyaml`/`libcurl`/`libogg`/`opus` have low minimums but build via generic/autotools infra, so cmake never runs on them.

That gap is itself the argument for keeping this global rather than per-package.

## Minor

- The test plan checkbox is unticked and no checks have run on the branch yet — `gcc-compat.yml` triggers on `general/package/**/*.mk`, so it should fire once approved.
- Unrelated to this PR, but worth its own issue: mbedtls **2.25.0 is from Dec 2020 and long past EOL**, shipping TLS on 93 boards.

---

If you *are* hitting a configure failure on current master, could you paste the error and your `cmake --version`? That would point at something the global export isn't covering, which would be worth fixing properly.

general/package/mbedtls-openipc/mbedtls-openipc.mk
```diff
@@ -8,7 +8,8 @@ MBEDTLS_OPENIPC_VERSION = 2.25.0
 MBEDTLS_OPENIPC_SITE = $(call github,ARMmbed,mbedtls,v$(MBEDTLS_OPENIPC_VERSION))
 MBEDTLS_OPENIPC_CONF_OPTS = \
 	-DENABLE_PROGRAMS=$(if $(BR2_PACKAGE_MBEDTLS_OPENIPC_PROGRAMS),ON,OFF) \
-	-DENABLE_TESTING=OFF
+	-DENABLE_TESTING=OFF \
+	-DCMAKE_POLICY_VERSION_MINIMUM=3.5
```
widgetii, line 12: This is already covered globally, so this line is a no-op in practice.

`Makefile:7` on master has:

```makefile
export CMAKE_POLICY_VERSION_MINIMUM := 3.5
```

(from 656094d5a / #1978). CMake honours `CMAKE_POLICY_VERSION_MINIMUM` as an **environment variable** as well as a `-D` cache entry, and Buildroot's `pkg-cmake.mk` `CONFIGURE_CMDS` doesn't scrub the environment — so the export already reaches this package's cmake invocation.

Verified with host cmake 4.3.3 against the extracted `mbedtls-openipc-2.25.0` tree:

```console
$ cmake -S mbedtls-openipc-2.25.0 -B /tmp/a
CMake Error at CMakeLists.txt:23 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

$ CMAKE_POLICY_VERSION_MINIMUM=3.5 cmake -S mbedtls-openipc-2.25.0 -B /tmp/c
-- Configuring done (0.7s)
-- Generating done (0.2s)
-- Build files have been written to: /tmp/c
```

The env-var form (which master already sets) is enough on its own.

widgetii:
This thread is closed as no response has been received

PR #2258, opened 2026-08-11
gk710x: enable WPA crypto in kernel configs

shipa-2:
## Summary
- Enable `CONFIG_CRYPTO_CCM`, `CONFIG_CRYPTO_CTR`, `CONFIG_CRYPTO_SEQIV`, and `CONFIG_CRYPTO_SHA256` in `gk710x` / `gk710xs` generic kernel configs for WPA2.

## Test plan
- [ ] Rebuild `gk7102_lite` / `gk7102s_lite` kernel; confirm crypto options are set
- [ ] WPA2 association works with an appropriate Wi-Fi driver

Part of splitting #2256.

Made with [Cursor](https://cursor.com)


flyrouter (commented):
No major objections if it’s really necessary. 
However, as with other Goke cores, it might be worth setting these some options via =m

widgetii:
@flyrouter's suggestion was right, and it turns out to be unanimous rather than a
preference — pushed as `5291924`.

Every other Goke core in the tree ships these as modules, with only SHA256 built in:

| core | CCM | SEQIV | CTR | SHA256 |
| --- | --- | --- | --- | --- |
| gk7201v200 | `=m` | `=m` | `=m` | `=y` |
| gk7202v300 | `=m` | `=m` | `=m` | `=y` |
| gk7205v200 | `=m` | `=m` | `=m` | `=y` |
| gk7205v300 | `=m` | `=m` | `=m` | `=y` |
| gk7205v500 | `=m` | `=m` | `=m` | `=y` |
| gk7605v100 | `=m` | `=m` | `=m` | `=y` |

Six out of six. gk710x/gk710xs as submitted were the only Goke configs setting all
four `=y`, which puts three crypto implementations into the kernel image — on the
oldest and most flash-constrained parts of the family, where it is least
affordable. The kernel crypto API loads these on demand, which is why `=m` works
on the other six.

CCM, SEQIV and CTR are now `=m`; SHA256 stays `=y`. The four symbols match the
other cores exactly.

@shipa-2 — I pushed this to your branch rather than waiting, since the PR had been
idle since the review and the change is mechanical. Two other things for the
record:

- I rebased the branch onto master earlier (it was 40-odd commits behind and its
  queued CI predated the matrix-narrowing selector, so it was showing 98 red
  builds that were really cancellations). It now builds `gk7102_lite` and
  `gk7102s_lite` — the two boards this can actually affect.
- Your test plan is still unchecked: neither "confirm crypto options are set" nor
  "WPA2 association works" was ticked. CI proves the kernel builds with these
  symbols, which is not the same as proving WPA2 associates. If you have a gk710x
  board with Wi-Fi, that result would be worth having on the record here — merging
  on the strength of convention plus a green build, not on a hardware test.

PR #2263, opened 2026-08-12
sysupgrade: put the mounts back when the pivot does not happen

widgetii:
Follow-up to #2252, addressing two review findings on that PR that I did not act on before it merged. Both are live on master.

### 1. A failed pivot strands the camera (the serious one)

`enter_ramfs` moves `/dev`, `/proc` and `/tmp` into the ramfs and *then* attempts `pivot_root`. Every failure after that point returned without moving them back — so the "falls back to flashing in place" promise was empty. The fallback needs `/dev` for the MTD nodes, `/proc` for `get_device` and `/tmp` for the unpacked image.

And it is worse than just a broken fallback: it leaves the **running** camera with no device nodes at all, which is a bigger failure than the one #2252 fixes.

`ramfs_unwind()` restores them and drops the tmpfs; every path between the first move and the pivot now goes through it.

Two details the hardware insisted on:
- the `umount` needs a **lazy fallback** — `/sys` carries submounts (debugfs, configfs) and a move-back that does not take leaves the tmpfs busy;
- `enter_ramfs` unmounts `$RAM_ROOT` **before** mounting, so an attempt that could not fully unwind does not get a second tmpfs stacked on the first.

An `exec` that fails *after* a successful pivot is deliberately not unwound: the old root is behind `./mnt`, so returning would send the fallback off to flash against a root it can no longer see. Nothing has been written by then, so it reboots and comes back on the current firmware.

### 2. State lost across the re-exec

`remote_update` and `lock_owned` were not exported. `verify_rootfs` branches on the first for an unmountable rootfs; `die()` needs the second to know the lock is its to remove.

I audited the rest of the phase-2 call graph for the same mistake. `jffs2` looked like a third, but it is local to `do_wipe_overlay` and derived from the exported `flash_type`.

### Verification

On an ssc30kq, forcing `pivot_root` to fail:

```
before: /dev=devtmpfs /proc=proc /tmp=tmpfs /sys=sysfs
  Could not move into RAM; flashing in place instead.
enter_ramfs rc=1
after:  /dev=devtmpfs /proc=proc /tmp=tmpfs /sys=sysfs
PASS mounts restored
PASS /ram unmounted
fallback: mtd=2 dev=ok tmp=ok majestic=up
```

4 new invariants; **86 checks pass**.

Worth recording: my first two attempts at this hardware test were themselves wrong and I nearly reported a false PASS. The first ran `enter_ramfs` via `eval` in an interactive shell, where `$0` is not a file — so `cp "$0"` failed and it returned *before* any mount moved, making "before == after" vacuous. The second put the `pivot_root` stub in `/tmp`, which `enter_ramfs` moves away mid-function, so the real `pivot_root` ran and the camera genuinely pivoted. The passing run above uses a real script file and a stub outside `/tmp`.

### DIFF
--- a/.github/scripts/test_sysupgrade.sh
+++ b/.github/scripts/test_sysupgrade.sh
@@ -913,7 +913,7 @@ awk '/^die\(\)/,/^}/' "$SRC" | grep -q '_ramfs_phase' \
     || bad "a die() inside the ramfs must reboot; there is no system left to return to"
 # The WebUI keeps majestic alive on purpose: it is the server streaming the log,
 # and SIGQUIT to a majestic already in upgrade mode is a use-after-free
-# (widgetii/majestic#288). The only legitimate one left is free_resources'
+# (tracked daemon-side). The only legitimate one left is free_resources'
 # non---web kill, which frees video memory for the download.
 if [ "$(grep -c 'killall -q -3 majestic' "$SRC")" = "1" ] &&
     awk '/^free_resources\(\)/,/^}/' "$SRC" | grep -q 'killall -q -3 majestic'; then
--- a/.github/scripts/test_sysupgrade.sh
+++ b/.github/scripts/test_sysupgrade.sh
@@ -62,6 +62,7 @@ sed -e 's|grep "GITHUB_VERSION" "$1/etc/os-release"|grep "GITHUB_VERSION" "${1:-
     -e "s|/etc/os-release|@SB@/etc/os-release|g" \
     -e "s|/proc/mtd|@SB@/proc/mtd|g" \
     -e "s|/proc/cmdline|@SB@/proc/cmdline|g" \
+    -e "s|/proc/mounts|@SB@/proc/mounts|g" \
     -e "s|/proc/sys/vm/drop_caches|@SB@/tmp/drop_caches|g" \
     -e "s|/etc/init.d/|@SB@/etc/init.d/|g" \
     -e "s|/bin/busybox|@SB@/bin/busybox|g" \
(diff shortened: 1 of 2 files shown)


widgetii:
Both findings were right; fixed in `a00cca009`.

**1. Unwind restore not validated.** Correct — the function restored the mounts and returned regardless of whether it worked, so a failed move-back still handed the fallback an environment with no `/dev` or `/proc`. That is the condition this function exists to prevent, reached by a different route. It now checks the three the fallback cannot do without (`/dev/null` as a character device, `/proc/mounts` readable, `/tmp` present in it) and reboots rather than flashing blind if any is missing. Nothing has been written at that point, so the camera comes back on its current firmware.

Not via `reboot_system()` deliberately: `-x` must not be honoured in that state, and there is no state left worth tidying.

**2. Blind RAM_ROOT unmount.** Also correct, and the reasoning about the environment override is exactly right — the test harness sets `RAM_ROOT`, which is what makes it a real interface rather than a constant. It now only clears the mount if `/proc/mounts` shows a tmpfs of our own there.

Worth recording how the first fix surfaced a harness gap: the sandbox rewrites `/tmp` to `$SB/tmp` inside the grep **pattern**, while `/proc/mounts` stayed the host's — so the check could never match and every case took the emergency reboot path. **35 tests failed at once**, which is the right way for that to show up rather than passing quietly. The sandbox now provides a `/proc/mounts` describing its own layout.

Re-verified on an ssc30kq with `pivot_root` forced to fail: mounts come back identical, tmpfs gone, check passes, run falls back rather than rebooting.

```
before: /dev=devtmpfs /proc=proc /tmp=tmpfs /sys=sysfs
  Could not move into RAM; flashing in place instead.
enter_ramfs rc=1
after:  /dev=devtmpfs /proc=proc /tmp=tmpfs /sys=sysfs
PASS mounts restored / PASS /ram unmounted
fallback usable: dev=ok proc=ok tmp=ok mtd=2
```

3 new invariants; **89 checks pass**.

PR #2266, opened 2026-08-14
ci: parse-check every shell script that ships in the image

widgetii:
Follow-up to #2244 / #2265. While reviewing a one-line change to `S95majestic` it became clear that nothing in CI would have caught a syntax error in it — `shell-tests` parse-checks exactly one file, `sysupgrade`.

## The gap

Everything that lands in the rootfs as a shell script — the init scripts, `/etc/profile` that every login sources, `/init` on initramfs boards, the `usr/sbin` helpers, and the vendor `load_*` scripts that insert the SoC drivers — is copied in **verbatim** by the overlay and by the package install steps. Nothing compiles or lints them, so a green build says nothing about whether they parse. A typo's first symptom is a service, or a driver load, that silently doesn't happen on a camera.

## What this adds

A `shipped-shell-parse` job running `.github/scripts/test_shell_parse.sh`, which discovers scripts **by shebang** across the whole overlay and every package `files/` dir, and parses each under **busybox ash**. 128 files are covered.

**Discovery is by shebang, not by path.** The first draft of this PR hand-listed globs (`init.d`, `usr/sbin`, `files/S??`) and missed **80 of the 127** scripts in the tree — `/init`, `/etc/rc.local`, the mdev and udhcpc helpers, the network if-up hooks, and every `osdrv-*/files/script/load_*`. Symlinks are resolved and deduplicated so `extutils` isn't parsed once per `usr/sbin` alias.

**busybox ash is the only interpreter, deliberately.** It's what runs on the device. dash would be the *wrong* bar: the shipped busybox sets `CONFIG_ASH_BASH_COMPAT` (`general/package/busybox/busybox.config:1138`), so `function name()` is valid in the field — and four scripts that work on real hardware are rejected by dash:

```
hisilicon-osdrv-hi3516cv6xx/files/script/load_hisilicon   ash: OK   dash: FAIL
hisilicon-osdrv-hi3519dv500/files/script/load_hisilicon   ash: OK   dash: FAIL
legacy/sigmastar-osdrv-msc313e/files/script/auto_run.sh   ash: OK   dash: FAIL
sigmastar-osdrv-infinity6e/files/script/zoom.sh           ash: OK   dash: FAIL
```

A check stricter than the target turns working code red, which is worse than no check — it teaches people to ignore the job.

## One thing worth reviewer attention

**I deleted two lines from `sysupgrade-verify`.** It ran `sh -n` then `dash -n` on sysupgrade; on `ubuntu-latest` `/bin/sh` *is* dash, so that was the same check twice, and neither was the target shell. The new job covers that file under ash. Both jobs carry the identical repo guard, so they're skipped together and this can't leave sysupgrade unchecked. Happy to restore it if you'd rather this were purely additive.

## Verification

```
$ bash .github/scripts/test_shell_parse.sh
...
checked 128 shell script(s)
all parsed clean under busybox ash
```

A check that cannot fail is worthless, so the failure paths are verified too:

| scenario | expected | actual |
|---|---|---|
| syntax error injected into `general/overlay/init` | fail | `FAIL ash general/overlay/init`, exit 1 |
| `STRICT=1`, busybox absent | fail | exit 1, explicit message |
| busybox without `ASH_BASH_COMPAT` | refuse | exit 1, refuses to apply a stricter parser than the device |
| discovery matches < 100 files | fail | exit 1, "discovery has regressed" |
| clean tree | pass | exit 0 |

`test_sysupgrade.sh` still passes after the job edit.

Three guards exist specifically so this can't silently become a no-op: a missing busybox, a busybox with the wrong config, and a discovery regression each fail loudly rather than reporting green — the last one added because the first draft's blind spot looked identical to a clean run.

## Scope

Syntax only. `-n` parses without executing, so this says nothing about runtime behaviour, and a bashism that ash accepts is by definition not reported. A `checkbashisms` pass would be worth having but is a much larger change — `local` is used throughout — and belongs in its own PR.

One caveat on placement: `shell-tests` is **not** part of the `CI Gate` required status check (that's `build.yml`, gating on `preflight`/`buildroot`/`publish`). So this job is advisory and won't block a merge on its own. Adding it to the required contexts is a branch-protection change and felt like your call.

Verified with busybox 1.35.0; the tree builds 1.36.1. The `ASH_BASH_COMPAT` probe means a version or config mismatch fails loudly rather than silently changing the bar, but I checked against 1.35.0, not the exact shipped build.

### DIFF
--- a/.github/scripts/test_shell_parse.sh
+++ b/.github/scripts/test_shell_parse.sh
@@ -1,0 +1,144 @@
+#!/bin/bash
+# Parse-check every shell script that ships inside the firmware image.
+#
+# Before this, the only parse coverage in CI was two lines in shell-tests.yml
+# pinned to sysupgrade. Everything else that lands in the rootfs -- the init
+# scripts, /etc/profile sourced by every login, /init for initramfs boards, the
+# vendor load_* scripts that insert the SoC drivers -- shipped with nothing
+# confirming it even parses. Nothing else covers them: the overlay and each
+# package's install step copy these files in verbatim, so a green build says
+# nothing about their syntax. A typo's first symptom is a service, or a driver
+# load, that silently does not happen on a camera.
+#
+# DISCOVERY is by shebang over the whole overlay and every package files/ dir,
+# not by a list of paths. An earlier draft hand-listed globs (init.d, usr/sbin,
+# files/S??) and silently missed 80 of the 127 scripts -- /init, /etc/rc.local,
+# the mdev and udhcpc helpers, the network if-up hooks, and every
+# osdrv-*/files/script/load_* -- because they did not match the shapes guessed
+# at. Anything with a shell shebang is checked wherever it lives, so a script
+# added in a new shape is covered when it lands.
+#
+# INTERPRETER is busybox ash, because that is what runs on the device, and only
+# that. dash is deliberately NOT used: the shipped busybox is built with
+# CONFIG_ASH_BASH_COMPAT=y (general/package/busybox/busybox.config), so the
+# `function name()` keyword is valid in the field, and four scripts that work on
+# real hardware -- both hi3516cv6xx/hi3519dv500 load_hisilicon, msc313e
+# auto_run.sh, infinity6e zoom.sh -- are rejected by dash. Checking against a
+# stricter shell than the target turns working code red, which is worse than no
+# check at all: it trains people to ignore the job.
+#
+# SCOPE: syntax only. `-n` parses without executing, so a passing file may still
+# be wrong at runtime, and a bashism that ash accepts is by definition not
+# reported here. This is the cheap half of the problem, not a portability audit
+# -- see the note in shell-tests.yml about checkbashisms.
+
+set -u
+
+# CI sets STRICT=1: there, a missing busybox means the check silently did not
+# happen, which should fail loudly rather than pass a green run. Locally it just
+# reports what it could not do and carries on.
+STRICT=${STRICT:-0}
+
+fail=0
+ok()   { echo "ok   $*"; }
+bad()  { echo "FAIL $*"; fail=$((fail + 1)); }
+note() { echo "--   $*"; }
+
+BUSYBOX=$(command -v busybox 2>/dev/null || true)
+
+if [ -z "$BUSYBOX" ]; then
+	if [ "$STRICT" != 0 ]; then
+		echo "FAIL busybox not found and STRICT=$STRICT -- refusing to report a"
+		echo "     pass when nothing was actually parsed."
+		exit 1
+	fi
+	note "busybox not found: nothing to parse against, skipping."
+	note "install busybox-static to run this as the device would."
+	exit 0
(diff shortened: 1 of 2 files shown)


widgetii:
Qodo's finding was correct, and understated the problem. Fixed in d436175.

**Verified the claim rather than taking it on faith**, since a bot finding can itself be wrong. It wasn't — and the real number is worse. Hand-listed globs covered 47 entries (42 unique, since five `usr/sbin` symlinks re-parsed `extutils`). A shebang sweep of the tree finds **127**. So the job was claiming broad coverage while missing **80 scripts**, including:

- `general/overlay/init` — the initramfs init
- `general/overlay/etc/rc.local`, `lib/mdev/automount.sh`, `usr/share/udhcpc/default.script`
- `etc/network/if-pre-up.d/*`, `if-post-down.d/*`, `etc/wireless/*`
- **every** `osdrv-*/files/script/load_*` — the scripts that insert the SoC drivers
- `legacy/webui/files/init.d/S50httpd`, i.e. the nested `files/*/S??*` shape called out in the finding

A parse-check that silently skips most of its subject is worse than none, because the green tick is read as coverage. Discovery is now by shebang across the whole overlay and every package `files/` dir, so it no longer depends on guessing where scripts live, and symlinks are resolved and deduplicated.

**A second bug surfaced while fixing this, which the finding didn't mention.** Widening discovery made four scripts fail — all the bash `function name()` keyword in `#!/bin/sh` files:

```
hisilicon-osdrv-hi3516cv6xx/files/script/load_hisilicon
hisilicon-osdrv-hi3519dv500/files/script/load_hisilicon
legacy/sigmastar-osdrv-msc313e/files/script/auto_run.sh
sigmastar-osdrv-infinity6e/files/script/zoom.sh
```

The tempting read is "four real bugs, ship it". It was wrong. My test had run `dash -n && busybox ash -n`, so **ash never executed** — the failures came from dash alone. The shipped busybox sets `CONFIG_ASH_BASH_COMPAT` (`general/package/busybox/busybox.config:1138`), so `function` is valid on the device, and all four parse fine under ash. Had I trusted that first result I'd have "fixed" four scripts that were never broken.

dash is therefore the wrong bar and has been dropped: it is stricter than the target, and a check that reddens working code just teaches people to ignore the job. busybox ash only. A guard now probes for `ASH_BASH_COMPAT` at startup and refuses to run if absent, rather than quietly applying a stricter parser than the device.

Also added a discovery-regression guard (fail if fewer than 100 scripts are found) — the previous blind spot looked identical to a clean run, and that shouldn't be possible twice.

On the two alternatives raised:

- **ShellCheck/checkbashisms** — agreed, and agreed it's noisy given `local` is used throughout. Worth its own PR, not this one.
- **Buildroot-produced busybox** — sharper than it first looks, since `ASH_BASH_COMPAT` is a *config* option, so a distro busybox could differ from the shipped one and silently move the bar. That's what the probe defends against. Building busybox in CI is still the more faithful answer if the probe ever proves insufficient.

PR #2272, opened 2026-08-16
qodo-gate: re-evaluate when Qodo posts its review comment

widgetii:
## Problem

Found by Qodo's own review of #2269 — a correct finding against the fix in that PR.

#2269 taught the gate to accept an issue comment headed `Code Review by Qodo` as proof of review, since that is the only proof a zero-finding PR ever produces. It did not give the workflow a reason to *look again* when that comment arrives.

So the gap only moved:

1. PR opens → `qodo-gate` runs → Qodo has not answered yet → **fail** (correct)
2. Qodo posts `Code Review by Qodo` a minute or two later — as an issue comment
3. `issue_comment` was not a trigger, so **nothing re-evaluates the gate**
4. The required check stays red until an unrelated push

The irony is that #2269 could not have caught this itself: it went green because I pushed a second commit, which re-triggered the gate at a point where the comment already existed. A `synchronize` event masked a missing `issue_comment` event.

## Fix

Adds the `issue_comment` trigger, plus the plumbing that payload requires:

- **A job guard.** `issue_comment` also fires for plain issues, which have no PR to gate. Every other trigger here is PR-only, so `github.event_name != 'issue_comment' || github.event.issue.pull_request` is the whole condition.
- **PR number from `github.event.issue.number`.** `issue_comment` carries `issue`, not `pull_request`; for a comment on a PR the issue number *is* the PR number.
- **Head sha resolved at runtime in the sweep step.** It read `github.event.pull_request.head.sha`, which `issue_comment` does not carry. Now looked up from the PR when empty, and still best-effort — a failed lookup skips the sweep rather than turning a PASS into a FAIL.

The sweep matters most on exactly this trigger: it is the one that clears a clean PR's earlier red run off the head commit, which branch protection's rollup would otherwise keep counting.

## Evidence

Review configuration only; nothing enters an image.

- Root cause is reproducible by inspection: `on:` had no `issue_comment`, while the detection logic added in #2269 depends on a comment.
- Workflow parses; triggers are `pull_request`, `pull_request_review`, `pull_request_review_comment`, `issue_comment`; both steps resolve `PR` from either payload shape.
- **This PR cannot test the path it fixes, and its own gate will stay red.** I initially expected it to self-heal; it will not. GitHub runs `issue_comment`-triggered workflows from the **default branch**, not from the PR head — so the new trigger is inert until this merges. Observed directly here: Qodo posted `PR Summary by Qodo` (an issue comment) at `16:01:26Z` and no `qodo-gate` run appeared for it; the only run is the `pull_request` one that failed at open. The same is true of the `pull_request_review` triggers, which is why those runs sat queued forever on #2268 while only `pull_request` runs completed.

  So merging this needs the same manual nudge #2269 needed — a push, or the `skip-qodo-gate` label. That is the bootstrap problem in a sentence: **the fix for "the gate cannot self-heal" cannot itself self-heal.**

  **Real verification is the first clean PR opened after this merges:** its `qodo-gate` should go red at open and turn green on its own once Qodo comments, with no push. If it stays red, this fix is wrong.

Also confirmed while verifying #2269's merge: the `.pr_agent.toml` loader fix works. A `/review` on master's config produced a normal review with **no** `failed to apply 'local' repo settings` comment, so the custom standards are live for the first time.

## Scope

- [x] No kernel patches, no single-board files, no debugging tooling
- [x] Nothing under `general/overlay/`; no package sources changed
- [x] No code — CI workflow only, does not enter any image

### DIFF
--- a/.github/workflows/qodo-gate.yml
+++ b/.github/workflows/qodo-gate.yml
@@ -62,9 +62,28 @@ permissions:
 jobs:
   qodo-gate:
     runs-on: ubuntu-latest
-    # issue_comment fires for plain issues too, which have no PR to gate.
-    # Every other trigger here is PR-only, so this is the whole guard.
-    if: github.event_name != 'issue_comment' || github.event.issue.pull_request
+    # Narrow the issue_comment trigger to the single comment that can change
+    # this gate's verdict: Qodo's review. Without all three clauses the job is
+    # an API-heavy workflow holding `actions: write` that any commenter could
+    # run at will, on any PR, as often as they liked — CI noise, and needless
+    # rate-limit exposure on a REQUIRED check.
+    #
+    #   issue.pull_request  — issue_comment fires for plain issues too, and
+    #                         those have no PR to gate.
+    #   comment author      — only the bot's own comment is evidence of review.
+    #                         A human typing /review still works: Qodo answers,
+    #                         and that answer is what triggers this.
+    #   comment body        — the bot also posts "Qodo is busy working" and
+    #                         "PR Summary by Qodo", neither of which means the
+    #                         diff was reviewed. Matching the review header
+    #                         keeps this to one run per review.
+    #
+    # Every other trigger here is PR-only and needs no guard.
+    if: >-
+      github.event_name != 'issue_comment' ||
+      (github.event.issue.pull_request &&
+       startsWith(github.event.comment.user.login, 'qodo-free-for-open-source-projects') &&
+       contains(github.event.comment.body, 'Code Review by Qodo'))
     steps:
       - name: Require a Qodo review with all its threads resolved
         env:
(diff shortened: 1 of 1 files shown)


widgetii:
Fixed in the latest push — a fair catch, and I had left the guard doing only half its job.

The guard established that the commented-on thing was a PR, and stopped there. So `issue_comment` fired for any comment by anyone: a required check turned into an API-heavy workflow holding `actions: write`, runnable at will.

It now also requires the comment to be the bot's **and** to carry the `Code Review by Qodo` header:

- **author** — only Qodo's own comment is evidence that Qodo reviewed. A human typing `/review` still works, because Qodo answers and that answer is what triggers the gate.
- **body** — the bot also posts `Qodo is busy working` (written before it has read anything) and `PR Summary by Qodo` (`/describe` output). Neither means the diff was reviewed, and matching the header also cuts three runs per review down to one.

On your severity framing: I agree it is worth fixing, though the practical exposure was bounded — `pull_request` already re-runs this on every push, so a determined commenter had a cheaper lever already. The stronger argument is the one you make about a *required* check: extra failure surface there is a merge blocker, not just noise.

Issue #2074, opened 2026-05-08
GK7202V300 (GK-W7): no wlan0 — USB Wi-Fi 8065:6000; Lite NOR 8M ships mt7601u only


bneigher:
### Hardware
- SoC: **GOKE GK7202V300** (board **GK-W7** style, 8MB NOR **XMC XM25QH64** / **XM25QH64CHIQ**, JEDEC `0x20 0x40 0x17`).
- Wi-Fi: **USB** device **`8065:6000`** (appears on **USB bus 001** as single HS device). No separate PCIe/SPI Wi-Fi visible.
- Ethernet: **No PHY** on this board (`femac … connect to PHY failed` in dmesg — expected).

### Firmware
- **OpenIPC** “Lite” **NOR 8M** full image for GK7202V300 (kernel **Linux 4.9.37**, build **2026-04-22**, U-Boot **2016.11-gd5c37e8-dirty** Jan 2025).
- **Video:** `sensor=gc2053` (via `fw_setenv`), **Majestic** runs; sensor pipeline OK.

### Problem
- After boot: **`ip link`** only shows **`lo`** — **no `wlan0`** (or any wireless iface).
- **`lsusb`:** `Bus 001 Device 002: ID 8065:6000`
- Wireless modules under `/lib/modules/.../drivers/net/wireless/`: only **`mediatek/mt7601u/mt7601u.ko`** (+ **`mac80211`**). Firmware tree has **`mt7601u.bin`** only.
- **`modprobe mt7601u`** loads the driver (`usbcore: registered new interface driver mt7601u`) but **no USB bind / no netdev** — **8065:6000** is not MT7601U.

### Stock firmware note (for donor/driver hints)
- On vendor firmware, Wi-Fi was associated with **SSV6x5x** stack (**`ssv6x5x.ko`**, **`p2p0`**, **nl80211** / vendor config paths). Unclear if **`8065:6000`** is exactly SSV without full **`lsusb -v`** strings.

### Ask
1. Is there a **donor camera / board profile** or **different OpenIPC build** (still **NOR 8M**) that includes a driver/firmware set matching **USB `8065:6000`** for **GK7202V300**?
2. If not, what is the recommended path: **custom kernel module** from vendor tree, **Ultimate**-only package, or **issue + PR** to add **VID:PID 8065:6000** support — and which driver name should we target?

### Attachments / evidence
[...]

PR #2276, opened 2026-08-16
load_goke: opt-in sensor_dvp / sensor_mclk gate for DVP-wired boards

bneigher:
Adds an **opt-in** env-var gate so DVP-wired boards can select the DVP pad
routing, without changing behaviour for anything that does not ask for it.

`open_sys_config` picks MIPI or DVP pad routing from its `chip=` and
`g_cmos_yuv_flag=` arguments. On a board that routes the sensor's parallel data,
sync, PCLK and i2c to the SoC's DVP pads, the MIPI arguments mux the i2c
controller to pads the sensor is not connected to — every address NACKs, no chip
ID is ever read, and no video is possible.

Per @widgetii's review note on OpenIPC/firmware#2074, this is keyed off an env
var rather than `$CHIP_TYPE`:

> `chip=gk7205v200 g_cmos_yuv_flag=1` is right for your W7 wiring but wrong for
> MIPI-wired gk7202v300 boards, and there are some in the wild — a blanket change
> would break them.

So a profile opts in with `fw_setenv sensor_dvp 1`. Unset — every existing board —
takes exactly the path it takes today.

`sensor_mclk` is gated the same way. `open_sys_config` reads MCLK only from its
module parameters and defaults to 27 MHz at this `chip=`, so a sensor init table
tuned for 24 MHz (the GC2053 ForCar tables are) runs against the wrong clock.
Worth noting for anyone who hits this: the `.ini` MCLK key is **not** consulted on
this path, so sweeping it produces identical results at every value — which is
indistinguishable from having ruled the cause out. That cost me a while.

Companion to the `gk7202v300_lite_w7_8m` device profile in OpenIPC/builder,
which is where the board-specific bring-up lives.

Verified on three GK-W7 boards (GK7202V300, 8 MB NOR, GC2053 in DVP mode with
SID strapped high): with `sensor_dvp=1` and `sensor_mclk=24` the i2c bus comes
up, the sensor answers at 7-bit `0x3f` with chip ID `0x2053`, and the pipeline
delivers 1920x1080 H.264 over RTSP at 25 fps with `FrmErrCnt 0`. With the vars
unset the script is byte-for-byte equivalent in behaviour to before.

`sh -n` clean.

Refs: #2074

### DIFF
--- a/general/package/goke-osdrv-gk7205v200/files/script/load_goke
+++ b/general/package/goke-osdrv-gk7205v200/files/script/load_goke
@@ -32,8 +32,14 @@ YUV_TYPE0=0 # 0 -- raw, 1 --DC, 2 --bt1120, 3 --bt656
 # Keyed off an env var rather than the SoC name on purpose: both wirings exist
 # on gk7202v300, so testing $CHIP_TYPE here would fix DVP boards by breaking
 # every MIPI one. Profiles opt in with `fw_setenv sensor_dvp 1`.
+# SYSCFG_CHIP is the value handed to open_sys_config's chip= selector, which is
+# NOT the same thing as CHIP_TYPE (the detected SoC). They coincide by default;
+# a DVP board needs the gk7205v200 pad-routing tables while still BEING a
+# gk7202v300. Keeping them separate leaves CHIP_TYPE meaning exactly one thing,
+# so existing and future chip-specific branches are unaffected by this opt-in.
+SYSCFG_CHIP=$CHIP_TYPE
 if [ "$(fw_printenv -n sensor_dvp 2>/dev/null)" = "1" ]; then
-	CHIP_TYPE=gk7205v200
+	SYSCFG_CHIP=gk7205v200
 	YUV_TYPE0=1
 fi
 
@@ -107,7 +113,7 @@ insert_osal() {
 }
 
 insert_detect() {
-	modprobe open_sys_config chip=$CHIP_TYPE sensors=unknown g_cmos_yuv_flag=$YUV_TYPE0 board=$BOARD
+	modprobe open_sys_config chip=$SYSCFG_CHIP sensors=unknown g_cmos_yuv_flag=$YUV_TYPE0 board=$BOARD
 	insert_osal
 	insmod gk7205v200_base.ko
 	modprobe open_isp
@@ -180,13 +186,22 @@ insert_ko() {
 	if [ "$SENSOR" == "bt656" ] || [ "$SENSOR" == "jxf23_dc" ]; then
 		YUV_TYPE0=1
 	fi
-	modprobe open_sys_config chip=$CHIP_TYPE sensors=$SENSOR g_cmos_yuv_flag=$YUV_TYPE0 board=$BOARD
+	modprobe open_sys_config chip=$SYSCFG_CHIP sensors=$SENSOR g_cmos_yuv_flag=$YUV_TYPE0 board=$BOARD
 	# open_sys_config takes MCLK only from its module parameters and defaults to
 	# 27 MHz at this chip=. A sensor init table tuned for 24 MHz (the GC2053
 	# ForCar tables are) then runs against the wrong clock. The ini's MCLK key is
 	# not consulted on this path, so the register is written directly.
+	# A wrong MCLK does not fail loudly — it produces a corrupted or absent
+	# image while every log line still looks healthy. So say something when the
+	# override cannot be applied, rather than continuing silently misconfigured.
 	case "$(fw_printenv -n sensor_mclk 2>/dev/null)" in
-		24) devmem 0x120100F0 32 0x0000000D ;;
+		24)
+			if ! command -v devmem >/dev/null 2>&1; then
+				logger -p daemon.err -t load_goke "sensor_mclk=24 requested but devmem is missing; MCLK left at the chip default"
+			elif ! devmem 0x120100F0 32 0x0000000D; then
+				logger -p daemon.err -t load_goke "sensor_mclk=24: MCLK register write failed; sensor may not produce a usable image"
+			fi
+			;;
 	esac
 	insert_osal
 	insmod gk7205v200_base.ko
--- a/general/package/goke-osdrv-gk7205v200/files/script/load_goke
+++ b/general/package/goke-osdrv-gk7205v200/files/script/load_goke
@@ -21,6 +21,22 @@ os_mem_size=${os_mem_size:=32}
 BOARD=demo
 YUV_TYPE0=0 # 0 -- raw, 1 --DC, 2 --bt1120, 3 --bt656
 
+# DVP-wired sensor boards -- OPT-IN, set by the device profile.
+#
+# open_sys_config selects MIPI or DVP pad routing from its chip= and
+# g_cmos_yuv_flag= arguments. A board that wires the sensor to the DVP pads
+# needs the DVP path; with the MIPI arguments the i2c controller is muxed to
+# pads that are not connected to the sensor, so every address NACKs and no
+# sensor is ever detected.
+#
+# Keyed off an env var rather than the SoC name on purpose: both wirings exist
+# on gk7202v300, so testing $CHIP_TYPE here would fix DVP boards by breaking
+# every MIPI one. Profiles opt in with `fw_setenv sensor_dvp 1`.
+if [ "$(fw_printenv -n sensor_dvp 2>/dev/null)" = "1" ]; then
+	CHIP_TYPE=gk7205v200
+	YUV_TYPE0=1
+fi
+
 cd /lib/modules/$(uname -r)/goke/
 
 ##################################################################
@@ -165,6 +181,13 @@ insert_ko() {
 		YUV_TYPE0=1
 	fi
 	modprobe open_sys_config chip=$CHIP_TYPE sensors=$SENSOR g_cmos_yuv_flag=$YUV_TYPE0 board=$BOARD
+	# open_sys_config takes MCLK only from its module parameters and defaults to
+	# 27 MHz at this chip=. A sensor init table tuned for 24 MHz (the GC2053
+	# ForCar tables are) then runs against the wrong clock. The ini's MCLK key is
+	# not consulted on this path, so the register is written directly.
+	case "$(fw_printenv -n sensor_mclk 2>/dev/null)" in
+		24) devmem 0x120100F0 32 0x0000000D ;;
+	esac
 	insert_osal
 	insmod gk7205v200_base.ko
 	insmod gk7205v200_sys.ko
--- a/general/package/goke-osdrv-gk7205v200/files/script/load_goke
+++ b/general/package/goke-osdrv-gk7205v200/files/script/load_goke
@@ -39,8 +39,16 @@ YUV_TYPE0=0 # 0 -- raw, 1 --DC, 2 --bt1120, 3 --bt656
 # so existing and future chip-specific branches are unaffected by this opt-in.
 SYSCFG_CHIP=$CHIP_TYPE
 if [ "$(fw_printenv -n sensor_dvp 2>/dev/null)" = "1" ]; then
-	SYSCFG_CHIP=gk7205v200
 	YUV_TYPE0=1
+	# Only gk7202v300 needs to borrow another chip's tables. Its pinmux() branch
+	# is MIPI-only — it ignores g_cmos_yuv_flag entirely and always calls
+	# i2c0_for_mipi_sensor_pin_mux/vi_mipi_rx_mux — so there is no DVP path
+	# reachable for it at any flag value. gk7205v200 and gk7205v300 each have
+	# their own DVP path already; forcing a v300 onto the v200 VI pad table would
+	# be exactly this bug pointing the other way.
+	if [ "$CHIP_TYPE" = "gk7202v300" ]; then
+		SYSCFG_CHIP=gk7205v200
+	fi
 fi
 
 cd /lib/modules/$(uname -r)/goke/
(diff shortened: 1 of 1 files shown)


bneigher:
Thanks — both findings on this PR were fair and are now addressed in `f67e5c2`.

**1. `CHIP_TYPE` role conflation.** Agreed, and this one mattered more than it
looks. The opt-in was overwriting `CHIP_TYPE`, which is the *detected SoC*, with
the value wanted for `open_sys_config`'s `chip=` pad-routing selector. Those are
two different things that merely coincide by default: a DVP board needs the
`gk7205v200` routing tables while still **being** a gk7202v300. A dedicated
`SYSCFG_CHIP` now carries the selector, so `CHIP_TYPE` keeps meaning exactly one
thing and existing/future chip branches are untouched by this opt-in. That also
better serves @widgetii's original review note — the whole point is that boards
which do not set `sensor_dvp` are behaviourally identical to before.

**2. Silent MCLK write.** Also agreed, and this is precisely the failure mode
worth guarding: a wrong MCLK does not fail loudly, it produces a corrupted or
absent image while every log line still looks healthy. `devmem` presence and exit
status are both checked now, and either failure logs to `daemon.err`.

For context on why the MCLK hook exists at all: `open_sys_config` reads MCLK only
from its module parameters and defaults to 27 MHz at this `chip=`, while the
GC2053 ForCar PLL tables are 24 MHz values. The `.ini` MCLK key is never
consulted on this path — so sweeping it produces identical results at every
value, which is indistinguishable from having ruled the cause out. That cost me
a while, and it is the kind of thing worth having in the tree's history.

widgetii (dismissed):
Sorry for the delay on this one — you did the hard part and then waited, which is the wrong way round.

I checked this against the pinned SDK source (`openhisilicon` @ `2d637e35`, `kernel/sys_config/sys_config.c`) rather than just reading the diff. **The approach is right and your diagnosis is exactly correct.** Two changes I'd like before merge, then this goes in.

## What checks out

**The DVP lever is the only one that exists.** `pinmux()` dispatches on `chip=`, and per `kernel/compat/compat.h` in the goke build `C_HI3518EV300` is `"gk7202v300"`. That branch is MIPI-only and ignores `cmos_yuv_flag` entirely — it unconditionally calls `i2c0_for_mipi_sensor_pin_mux_gk7205v200()`, `sensor_cfg_for_mipi_sensor_mux()`, `vi_mipi_rx_mux()`. There is no DVP path reachable for `chip=gk7202v300` at *any* `g_cmos_yuv_flag`. Only the `gk7205v200` branch honours `cmos_yuv_flag == 1` → `i2c0_for_vi_raw_sensor_pin_mux()` + `vi_raw_mode_mux_gk7205v200()`. Your i2c-NACK symptom falls straight out of that. This isn't a hack, it's the only lever.

**The `SYSCFG_CHIP` separation is provably as contained as you say.** `chip_list` feeds only `pinmux()` and `clkcfg()` — and `clkcfg()`'s `C_HI3518EV300` and `C_HI3516EV200` branches write byte-identical values (`0x00A0=0x00240400`, `0x00F4=0x00110000`, `0x00FC=0x8`). So the swap moves pinmux and nothing else. Splitting it out of `CHIP_TYPE` was the right call.

**The MCLK story is confirmed.** `parse_sensor_clock()` returns `0x6` for `gc2053`, and the file's own table says `0x6: 27MHz`. MCLK is sourced *only* from that sensor-name table, so the `.ini` key genuinely is never consulted on this path. Your note that sweeping it gives identical results at every value — indistinguishable from having ruled it out — is accurate and worth having in the history. Thanks for writing it down.

## 1. The `devmem` write is a blind store to a read-modify-write register

```c
void sensor_clock_config(int index, unsigned int clock)
{
	reg_write32(clock << 2, 0xF << 2, reg_crg_base + 0x00F0);  /* RMW bits [5:2] */
}
```

`reg_crg_base = ioremap(0x12010000)`, so `0x120100F0` is the right address, and the sns0 select field is **bits [5:2] only**.

`devmem 0x120100F0 32 0x0000000D` stores the whole word. Decomposed: `0xD` = `0b1101` → bits[5:2] = `0b0011` = `0x3`, and the table reads `0x3: 24MHz` — so you are selecting the clock you intend, which is why it works on your three boards. But the same store also sets bit 0 and forces bits [31:6] to zero, and nothing in the driver ever writes that register wholesale.

Could you mask to the field instead? It also removes the magic constant and makes the var extensible to the other seven values:

```sh
# CRG PERI_CRG60 @ 0x120100F0, sns0 clock select = bits [5:2].
# 0x0:74.25 0x1:72 0x2:54 0x3:24 0x4:37.125 0x5:36 0x6:27 0x7:12 (MHz)
# open_sys_config's sensor_clock_config() read-modify-writes just this
# field -- match that rather than storing the whole word.
cur=$(devmem 0x120100F0)
devmem 0x120100F0 32 $(( (cur & ~0x3C) | (3 << 2) ))
```

While you're there — add a `*)` arm that logs a non-empty unsupported value. Right now `sensor_mclk=37` does nothing silently, which is the same failure class the rest of this PR is guarding against.

## 2. `sensor_dvp=1` shouldn't unconditionally force `gk7205v200`

`gk7205v300` has its own DVP path (`vi_raw_mode_mux_gk7205v300()` under `cmos_yuv_flag == 1`). A DVP-wired v300 that sets `sensor_dvp=1` would get pushed onto the **v200** VI pad table — the same bug this PR fixes, pointing the other way. Only `gk7202v300` lacks a DVP branch and needs to borrow:

```sh
if [ "$(fw_printenv -n sensor_dvp 2>/dev/null)" = "1" ]; then
	YUV_TYPE0=1
	# gk7202v300's pinmux() branch is MIPI-only -- it ignores g_cmos_yuv_flag.
	# Borrow gk7205v200's tables (same die family, and they do honour it).
	# gk7205v200/v300 already have their own DVP paths.
	if [ "$CHIP_TYPE" = "gk7202v300" ]; then
		SYSCFG_CHIP=gk7205v200
	fi
fi
```

## Minor

- `logger -p daemon.err -t load_goke` — the file's convention is `logger -s -p ... -t goke` (lines 356, 360, 366, 377). The `-s` matters here: a failed MCLK write is bring-up-fatal and should reach the console the way "SENSOR is not detected" does.
- The `command -v devmem` guard is harmless, but `CONFIG_DEVMEM=y` is set in `general/package/busybox/busybox.config`, so it is always present. Keep it if you prefer belt and braces.
- Asymmetry worth a comment rather than a change: `sensor_dvp` applies to both `insert_detect` and `insert_ko`, but `sensor_mclk` only to `insert_ko`, so first-boot autodetect probes at the default clock. Probably fine since i2c is master-clocked, but it will surprise someone eventually.

## Before this can merge

- **CI has never run on this branch** — fork PRs need workflow approval, so the umbrella gate has nothing to go on. I'll approve the run once you push.
- The branch is behind master; please rebase.

The firmware/builder split is right — generic mechanism here, board profile in OpenIPC/builder#119. I'll pick that one up next.

bneigher:
Thanks for going to the SDK source rather than the diff — that turned a "looks
plausible" review into a checkable one, and both findings were right. Pushed in
`1084143`.

## 1. The MCLK write

You're right, and the way it was wrong is worth stating: `0x0000000D` selects the
correct clock **by accident of the constant**, not by writing the field.
Decomposed, bits [5:2] are `0x3` = 24 MHz — but the same store also cleared bits
[31:6] and set bit 0, which `sensor_clock_config()` never does. It worked on
three boards, which is exactly how a blind store survives review.

Now masked to the field the driver actually touches:

```sh
mclk_cur="$(devmem 0x120100F0)"
devmem 0x120100F0 32 $(( (mclk_cur & ~0x3C) | (mclk_sel << 2) ))
```

I also took your suggestion to make it extensible and filled in the whole table
(74.25/72/54/24/37.125/36/27/12), with a `*)` arm that logs. You were right that
`sensor_mclk=37` silently doing nothing was the same failure class this PR is
supposed to be guarding against — it was inconsistent to guard the write and not
the value.

## 2. `sensor_dvp` forcing gk7205v200

This one I should have caught. I generalised a fix from the only board I had.
Only gk7202v300 lacks a DVP branch and needs to borrow; a DVP-wired gk7205v300
setting `sensor_dvp=1` would have been pushed onto the v200 VI pad table — the
same bug, mirrored, and harder to diagnose because the flag would look correct.
Now gated on `CHIP_TYPE = gk7202v300`.

## Minor

- `logger -s -p ... -t goke` throughout, per the file's convention. Agreed on
  `-s`: a failed MCLK write is bring-up fatal and should reach the console the
  way "SENSOR is not detected" does.
- Kept the `command -v devmem` guard. `CONFIG_DEVMEM=y` so it is redundant, but
  the read is now load-bearing too, and I would rather a missing tool say so than
  produce an empty `$mclk_cur` and a nonsense arithmetic expansion.
- Documented the `insert_detect` / `insert_ko` asymmetry in a comment rather than
  changing it, as you suggested — autodetect probes at the sensor table's default
  clock, which is fine because i2c is master-clocked, but now it is written down.

## One thing I have not verified

The masked write is correct by inspection and `sh -n` clean, but it is **not yet
tested on hardware** — my three boards are running the previous whole-word store.
I will flash and confirm the sensor still comes up before this merges, and report
back either way. Given the failure mode here is a silently corrupted image rather
than a loud error, I would rather not have "it compiles" standing in for "it
works".

widgetii:
Heads up: **I rebased this branch onto current master and force-pushed it** — `1084143` → `aa6c983`. Please `git fetch && git reset --hard origin/w7-dvp-sensor-gate` (or re-pull your fork) **before you flash**, so your hardware test runs against what will actually merge rather than a base that was 36 commits behind.

Nothing in your change was altered: the three commits replayed cleanly (master hadn't touched `load_goke` since your fork point), and the resulting file is byte-identical to what you pushed. Still `+79/-2` in one file.

I also approved the CI runs — that was the reason this PR had never had a check on it, and it was my hold-up, not yours.

## On the masked write

I checked the arithmetic rather than eyeballing it, and it holds. In both `dash` and busybox `ash`, `cur=0x00000861` with `mclk_sel=3` gives `0x0000084D` — bits [5:2] = `0x3`, and bits 0, 6 and 11 preserved.

Two ways that expression could have silently broken, both of which happen to be fine, but worth recording since they are invisible failure modes of exactly the kind this PR is about:

- busybox `devmem` prints `"0x%0*llX"` (`miscutils/devmem.c:126`), so the `0x` prefix is there and `$(( ))` parses it as hex. Had it printed bare digits, a reading like `00000741` would have been taken as **octal** and you would have written the wrong clock — silently, with a healthy-looking log.
- Its default read width is `8 * sizeof(int)` = 32 (`devmem.c:35`), matching your explicit `32` on the write, and the write value goes through `bb_strtoull(..., 0)`, so the decimal output of `$(( ))` is read back correctly.

Your `mclk_sel=""` sentinel is also right — it distinguishes "unsupported value" from `sel=0` (74.25 MHz), which a `-z`-on-zero check would have conflated.

Agreed on keeping the `command -v devmem` guard now that the read is load-bearing; an empty `$mclk_cur` feeding the arithmetic expansion is a worse failure than a clear message.

## Where this stands

The review is still marked changes-requested, and I am leaving it that way until you confirm on hardware — for the reason you gave yourself, which was the right instinct: the failure mode here is a silently corrupted image, so "it compiles" should not stand in for "it works". Report back either way and I will clear it and merge.

Thanks for the turnaround, and sorry again for the nine days of silence before it.

widgetii:
Rebased head is green across all six workflows, and the matrix picked exactly the right set — all seven Goke boards that use this `load_goke` built: `gk7202v300_lite/ultimate`, `gk7205v200_lite/ultimate`, `gk7205v300_lite/ultimate`, `gk7605v100_lite`.

Since a build cannot catch a wrong MCLK, I verified the two things a build can't, so you aren't the only evidence for this.

## 1. The register layout, confirmed on hardware

I read `0x120100F0` on a lab **hi3516ev300** — which is the same `sys_config.c` and the same CRG as this family, just with the HiSilicon names bound in `compat.h` instead of the Goke ones (`C_HI3516EV300` is `hi3516ev300` there and `gk7205v300` here):

```
0x120100F0 = 0x00000011      sensor: imx335
  bits [5:2] = 0x4
  bit0       = 1
  bits[31:6] = 0
```

`parse_sensor_clock("imx335")` returns `0x4` in the driver, and the table reads `0x4: 37.125MHz`. So the field really is bits [5:2] at that address, and the encoding is what we both read it to be. That is now measured rather than inferred.

I also ran your expression in that board's own busybox ash against the live register value, and exercised the write path with a value equal to what was already there — a true no-op, readback identical, `majestic` undisturbed:

```
sel=0 -> 0x00000001    sel=4 -> 0x00000011
sel=3 -> 0x0000000D    sel=6 -> 0x00000019
                       sel=7 -> 0x0000001D
```

**Worth knowing: `sel=3` computes to `0x0000000D`** — precisely the constant the first version hardcoded. On this silicon everything above bit 5 and bit 1 reads zero, so the blind whole-word store and the correct read-modify-write coincide. That is why your three boards worked and why nothing looked wrong. The masked form is still the right one, because it is correct by construction rather than by a property of the register that no one had checked; but it also means **your existing hardware validation very likely still stands** — the new code computes the same word the old code wrote, unless the W7's register has bits set above bit 5, in which case the masked version is the more correct of the two.

## 2. The no-op path, proven by execution trace

The gate's safety claim is that a board with neither var set behaves exactly as before. That is a claim about the sequence of commands the script runs, so I asserted it that way: stub every external onto `PATH`, log each argv, run the base and the candidate with identical inputs, diff the traces.

Byte-identical across **4 chips × 3 sensors**, plus the first-boot autodetect path (`insert_detect`/`remove_detect`) and the `-r` removal path. Then the opt-in routing and the MCLK masking on top — 39 checks.

The important part: I ran the same harness against `f67e5c2`, the version before the review, as a negative control. It fails 13 checks — catching both the unconditional `gk7205v200` force and the whole-word write. So the harness bites; it is not passing everything I point it at.

I would like to add this as `.github/scripts/test_load_goke.sh` alongside the existing `test_load_hisilicon.sh`, so the no-op guarantee is enforced in CI rather than resting on this thread. Happy to send it as a separate PR after this merges, or push it here if you would rather it land together — your call, it is your branch.

## Where that leaves things

Both review findings are addressed and independently verified. The one thing neither CI nor I can reach is the DVP pinmux actually bringing up the GC2053 on a DVP-wired gk7202v300 — the lab board is MIPI-wired with an imx335, so it is the twin of a gk7205v300, not of your W7.

Given the above I am clearing the changes-requested. If your flash comes back clean, say so and I will merge; if you would rather I merge now on the strength of the equivalence above, that is reasonable too, since the opt-in path is inert for every board that does not set the vars.

bneigher:
Hardware verification, as promised. Flashed to a GK7202V300 (OpenIPC nightly
base, kernel 4.9.37, U-Boot 2016.11-gd5c37e8). Sensor comes up: `i2c init` failures
0, ISP errors 0, 1920×1080 @ 25 fps.

The interesting part is that the boards had more to say than "it works".

**The reset value.** Read from the U-Boot prompt, before Linux touches anything:

```
OpenIPC # md.l 0x120100f0 1
120100f0: 00000000
```

(I dumped `0x12010000 16` first to confirm `md` was reading live registers and not
a dead window — that block is full of varied non-zero content, so the zero here is
real.)

**What the write actually sees.** I then instrumented `load_goke` to log the
register immediately before the write, and rebooted:

```
goke: NCAM-DEBUG pre-write reg=0x00000019
```

Decoded, that is:

| bits | value | meaning |
| --- | --- | --- |
| [5:2] | `0x6` | 27 MHz — the chip default |
| bit 0 | `1` | already set, by the driver, before `load_goke` runs |
| [31:6] | `0` | nothing above the field on this SoC |

Two things fall out of that.

First, it independently confirms the table in the patch: the hardware sits at
`sel=6` and the documented default for this chip is 27 MHz, so entry 6 → 27 MHz is
right. I had transcribed that table rather than measured it.

Second, and more to your original point: **the register is already
driver-configured by the time this code runs.** The old whole-word store of
`0x0000000D` was therefore not writing into a blank register — it was overwriting
live state with a constant. On this SoC the final value is identical either way
(`(0x19 & ~0x3C) | 0xC == 0xD`), but only because bit 0 happened to already be 1
and bits [31:6] happened to be 0. Neither is something the script knew or checked.
That is a sharper version of what you flagged, and it only showed up on hardware.

**What I did not verify.** My board is a gk7202v300, so it still takes the
borrow-gk7205v200-tables path — the `sensor_dvp` change is exercised in the sense
that the gate evaluates true, but a DVP-wired gk7205v200/v300 taking its own path
is untested by me. If anyone has one, that is the case worth a second pair of eyes.

widgetii:
Your `pre-write reg=0x00000019` is the most useful number in this thread, and it holds up independently of everything else below.

Set beside what I read on the lab board:

| board | sensor | register | bits [5:2] | `parse_sensor_clock()` | table |
| --- | --- | --- | --- | --- | --- |
| gk7202v300 (yours) | gc2053 | `0x00000019` | `0x6` | `0x6` | 27 MHz |
| hi3516ev300 (lab) | imx335 | `0x00000011` | `0x4` | `0x4` | 37.125 MHz |

Two different SoCs, two different sensors, two different table entries, both matching — with bit 0 set and bits [31:6] clear in both cases. Neither of us had that alone. Your U-Boot `md.l` read of `00000000` at reset adds the piece I couldn't get: bit 0 is set by the driver between reset and `load_goke`, not by the bootloader. That makes your point sharper than mine was — the old whole-word store wasn't writing into a blank register, it was overwriting live driver state with a constant, and the two only agreed because bit 0 happened to already be 1 and nothing lived above bit 5.

## One correction to the record

You posted a correction on OpenIPC/builder#119 sixteen minutes after this comment: your boards run a **forked `load_goke`** that hardcodes the DVP branch and the 24 MHz write and never reads the env. I don't think you'd traced that back to here, so — the `1920×1080 @ 25 fps`, `i2c init` failures 0 result above is evidence your fork works. It isn't evidence the merged script does.

The timing agrees with your correction rather than with the original: this merged at 17:11 and no nightly built from it existed by 17:35, and a board on stock pre-merge `load_goke` couldn't have brought up a DVP-wired sensor at all, because the gate didn't exist. So the script behind that 1080p25 was necessarily yours.

Flagging it because this is already merged and I'd rather the thread not read as end-to-end validation it isn't. It doesn't change the merge — that went in on the lab register measurement, the trace-equivalence harness with its negative control, and CI green across all seven Goke boards, none of which route through your bench. And it doesn't touch the register measurements above, which are reads of hardware and don't care which script logged them.

What's still open is the same gap you identified yourself: merged `load_goke`, DVP board, env set by the profile rather than by hand. Builder#119 now has CI running for the first time — I approved the runs — so the artifact for that test is about to exist.

## The case neither of us can reach

Agreed on the DVP-wired gk7205v200/v300 path. Your board is a gk7202v300, so it exercises the borrow branch; the case where a chip keeps its own `chip=` and takes its own DVP routing has no hardware behind it from either of us. The lab board is MIPI-wired, so it can't stand in.

It is covered by the trace harness at the level of "the right arguments reach `open_sys_config`" — that's what caught the original unconditional force — but not at the level of "those arguments produce video". If anyone reading has a DVP-wired v200 or v300, that's the one worth a report.

Thanks for the instrumented reboot, and for the correction. Both made the record more accurate than it would have been.

bneigher:
Correcting my "Hardware verification" comment above. @widgetii is right, and the
error is mine: I presented a result from a **forked** `load_goke` as verification
of the merged one.

Concretely, our fork hardcodes the DVP branch behind `[ "$CHIP_TYPE" = "gk7202v300" ]`
and writes MCLK behind an internal flag. It never calls `fw_printenv`. So the
merged script's actual gate — `[ "$(fw_printenv -n sensor_dvp)" = "1" ]` — did not
execute on that board at all.

That makes one line of my comment worse than merely unsupported. I wrote that the
`sensor_dvp` change "is exercised in the sense that the gate evaluates true". It
was not exercised in any sense. The gate was not present in the script that ran,
and I would have caught that by reading the file I flashed instead of the file I
had written.

**What does not survive:** `1920×1080 @ 25 fps`, `i2c init` failures 0, ISP errors
0, as evidence for this PR. That is evidence our fork works, which was never in
question.

**What does survive,** because it is a register read and does not care which script
emitted it:

- `md.l 0x120100f0` = `0x00000000` at the U-Boot prompt (reset state), with the
  surrounding CRG block dumped first to prove `md` was reading live registers.
- `pre-write reg=0x00000019` — bits [5:2] = `0x6`, taken after `open_sys_config`
  configured the register and before anything overwrote it.

Your lab board reading `0x00000011` with bits [5:2] = `0x4` for imx335 is a better
cross-check than mine, because it moves both the SoC and the sensor and still lands
exactly where `parse_sensor_clock()` says. Two boards, two sensors, two different
table entries, bit 0 set and bits [31:6] clear in both. I had one point and called
it confirmation; two points that disagree in the right way is confirmation.

The outstanding test is unchanged and now has a defined shape: stock upstream
`load_goke`, `sensor_dvp` and `sensor_mclk` cleared from U-Boot env, flashed from
the CI artifact on builder#119, so the profile is the only thing that sets them.
Our boards currently fail both preconditions — they run the fork, and I set those
vars by hand this afternoon. I will clear them and report on the builder PR.

Thank you for joining the two threads up. I had posted the correction on #119 and
not carried it back to here, which left the more flattering version standing in the
record of the PR that actually merged.

PR #2280, opened 2026-08-17
sysupgrade: stop the ramfs pivot leaving debris in the overlay

widgetii:
Reported by @usa- in OpenIPC/majestic-webui#120: cameras upgraded since the ramfs pivot landed come back carrying two entries nothing put there on purpose, on four SoCs at once.

```
/overlay/root/dev/null
/overlay/root/ram/
```

His guess in the report was right — *"something is writing to `/dev/null` (like `2>/dev/null`) while it has not yet been created or is no longer available."*

## `/dev/null`

A shell redirection **creates** its target. `enter_ramfs` moves `/dev` — the devtmpfs that supplied the node — out of the old root, and then keeps running there for four more commands. Every `2>/dev/null` in that window therefore discards nothing; it writes a regular file named `null` into a directory the shipped rootfs leaves empty. That root is an overlay, so the file lands in the upper layer and outlives the upgrade.

The mode is the fingerprint. Root's umask on the camera is `0077`, and the artifact is `0600` — exactly what `open(O_CREAT, 0666)` yields:

```
# ls -l /overlay/root/dev/null
-rw-------    1 root     root             0 Aug 16 09:59 /overlay/root/dev/null
```

`/rom/dev` ships no `null` node at all, so there is nothing for overlayfs to have copied up — the file is new.

Holding a descriptor open on the real node survives the mount move, the pivot and the exec, because a descriptor names the inode rather than the path. Reproduced **both ways on real hardware with no flash writes**, by moving `/dev` aside and putting it back:

| inside the window | result |
|---|---|
| `2>/dev/null` (old) | creates `-rw-------` regular file |
| `2>&3` (new) | creates nothing, still discards |

## `/ram`

The mount point itself. The rootfs ships no `/ram`, so `mkdir -p "$RAM_ROOT"` writes it into the upper layer, and nothing ever removed it — not after a successful pivot, not after a failed one, and not after the early bail-outs, which leave a tmpfs mounted on it as well.

Reclaimed on each path now. After the pivot the old root sits at `/mnt` with the tmpfs already detached, so the second phase takes it back before the first `flashcp` write destroys the filesystem it lives on; the single `enter_ramfs` call site covers every bail-out. `rmdir`, never `rm -rf` — if the tmpfs is somehow still attached, refusing is the right answer.

## Impact

Neither artifact is dangerous *today*: `/init` moves devtmpfs back over `/dev` early enough that nothing has been observed writing through the file. But it is a silent write to flash on every upgrade, and a trap for anyone who later adds a redirection to `/init` between its own `pivot_root` and its `mount -o move /rom/dev /dev`.

## Verification

End to end on the lab hi3516av300, with the patched script and the artifacts cleared first:

```
2.6.08.15 -> 2.6.08.16, clean reboot
/overlay/root/dev/   empty
/overlay/root/ram    No such file or directory
```

`test_sysupgrade.sh` gains one behavioural check (a failed pivot reclaims its mount point) and seven source invariants covering the redirect window, `ramfs_unwind`, where fd 3 is opened, and `rmdir`-not-`rm -rf`. Both halves fail the suite when reverted:

```
FAIL failed pivot left RAM_ROOT behind
FAIL enter_ramfs names /dev/null after moving /dev away -- that CREATES the file
```

`scr_version` is bumped to 1.0.58 so `self_update` actually delivers this — it compares versions, so reusing 1.0.57 would never install on a running camera.

### DIFF
--- a/.github/scripts/test_sysupgrade.sh
+++ b/.github/scripts/test_sysupgrade.sh
@@ -950,6 +950,13 @@ grep -qF 'rmdir "/mnt$RAM_ROOT"' "$SRC" \
 grep -qE '^case "\$RAM_ROOT" in /\*\)' "$SRC" \
     && ok "a relative RAM_ROOT override is normalised to an absolute path" \
     || bad "RAM_ROOT is used as \"/mnt\$RAM_ROOT\" and matched against /proc/mounts; it must be absolute"
+# That rmdir runs AFTER the pivot, where the applet symlinks are the only tools
+# there are. `busybox --list` is a build option, so on a camera without it the
+# hardcoded fallback list is the whole toolbox -- an applet missing from it is
+# simply "not found", and a cleanup that fails is one that silently did nothing.
+awk '/^enter_ramfs\(\)/,/^}/' "$SRC" | awk '/for applet in sh ash/,/done/' | grep -qw 'rmdir' \
+    && ok "the fallback applet list stages rmdir for the ramfs phase" \
+    || bad "the ramfs phase runs rmdir, but a busybox without --list would not have that applet"
 awk '/^enter_ramfs\(\)/,/^}/' "$SRC" | grep -q 'export remote_update' \
     && ok "remote_update survives the re-exec (verify_rootfs branches on it)" \
     || bad "remote_update is not exported; an unmountable rootfs behaves differently in phase 2"
--- a/.github/scripts/test_sysupgrade.sh
+++ b/.github/scripts/test_sysupgrade.sh
@@ -797,6 +797,24 @@ else
     bad "--no_ramfs -> expected no pivot but a normal flash, log='$(cat "$SB/tmp/flash.log")'"
 fi
 
+# A pivot that did not happen must still hand the old root back as it found it.
+# The shipped rootfs has no /ram, so the directory enter_ramfs makes lands in the
+# overlay's upper layer and stays there: every upgrade since the pivot landed
+# left a stray /overlay/root/ram behind (OpenIPC/majestic-webui#120).
+#
+# rmdir is stubbed rather than left real because this sandbox fakes `mount` --
+# nothing was ever mounted on $SB/ram, so it still holds the staged busybox tree
+# and a real rmdir would rightly refuse to remove it.
+stub rmdir 'echo "rmdir $*" >> "$FLASH_LOG"; exit 0'
+reset_env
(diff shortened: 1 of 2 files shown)


widgetii:
Good catch on both — fixed in `874036cad` and `e54aa6946`.

**`RAM_ROOT` absolute.** Normalised once at the definition rather than at each use site. Two things assumed the leading slash and both failed *quietly*: `/proc/mounts` records mount points absolutely, so `ramfs_discard`'s "is this tmpfs ours" check could never match a relative path, and `"/mnt$RAM_ROOT"` composes `ram` into `/mntram`.

**`rmdir` not staged.** This one was a genuine hole in the fix, and the worst possible shape for it — the reclaim runs *after* the pivot, where the applet symlinks are the only tools that exist, and `rmdir` was never on the hardcoded fallback list (`rm` is; nothing after the pivot had needed `rmdir` before). It would have failed as "not found", been swallowed by its own `2>/dev/null`, and left exactly the debris the PR removes. Indistinguishable from not having the fix.

Worth noting for scope: `busybox --list` populates the symlinks wherever it is available, which is every stock OpenIPC build — the lab av300 reports 282 applets with `rmdir` among them — so the hardware verification in the PR description exercised the `--list` path and would not have caught this. The fallback list is precisely the "build without `--list`" case.

Both now have test invariants, and both fail the suite when reverted:

```
FAIL the ramfs phase runs rmdir, but a busybox without --list would not have that applet
```

widgetii:
The remaining `Bugs (1) — Ramfs cleanup may fail` on the review summary is stale: it is the same `rmdir` finding, carried forward rather than re-derived. It was fixed in `e54aa6946`, which is the commit the summary now points at.

At the exact sha and the exact lines the finding cites (`sysupgrade[622-627]` @ `e54aa694`):

```sh
	# --list is a build option; fall back to what the flash phase actually uses.
	for applet in sh ash awk basename cat cut dd flash_eraseall flashcp grep \
		head ln losetup ls mkdir mount od printf reboot rm rmdir sed sleep \
		sync tail timeout umount xxd; do
```

`rmdir` is there, so the premise ("that set omits rmdir") no longer holds. The item also lost its `⭐ New` badge on the re-run, which is consistent with a carried-over entry rather than a fresh confirmation.

Guarded by a test either way, so a regression would be caught rather than argued about:

```
ok   the fallback applet list stages rmdir for the ramfs phase
```

Issue #2036, opened 2026-05-03
Nightly CI is flaky on master — package download 404s break ~30% of scheduled runs


widgetii:
## Problem

Over the past 30 days (2026-04-03 → 2026-05-03), `build.yml` failed in **10 of 35** scheduled runs on `master`. ~80% of failures share a single root cause: a transient HTTP 404 during package download aborts the whole job.

## Data

| Date (UTC) | Run | Failed boards | Root cause |
|---|---|---|---|
| 2026-05-02 | [25264150594](https://github.com/openipc/firmware/actions/runs/25264150594) | hi3516cv300_lite, hi3518ev200_ultimate | mbedtls-openipc 404 from `codeload.github.com` |
| 2026-05-01 | [25237141812](https://github.com/openipc/firmware/actions/runs/25237141812) | hi3516dv100_lite | end-of-build "Server Error" (GHA infra) |
| 2026-04-29 | [25139121700](https://github.com/openipc/firmware/actions/runs/25139121700) | t30_lite, gk7605v100_lite | ingenic-opensdk-HEAD 404, majestic 404 |
| 2026-04-28 | [25082907818](https://github.com/openipc/firmware/actions/runs/25082907818) | hi3516ev300_ultimate | majestic 404 + lame-3.100 from sourceforge 404 |
| 2026-04-27 | [25024579089](https://github.com/openipc/firmware/actions/runs/25024579089) | hi3516cv100_lite, hi3516dv200_lite, hi3516dv300_lite | motors-HEAD 404, jsonfilter@\<sha\> 404 (×2) |
| 2026-04-26 | [24969305329](https://github.com/openipc/firmware/actions/runs/24969305329) | gk7205v200_ultimate | toolchain-external from `miniupnp.free.fr` 404 |
| 2026-04-19 | [24641267971](https://github.com/openipc/firmware/actions/runs/24641267971) | gk7605v100_lite | end-of-build "Server Error" (GHA infra) |
| 2026-04-06 | [24055456241](https://github.com/openipc/firmware/actions/runs/24055456241) | hi3516ev300_neo | one-off kernel build error (drivers/net/mdio); already self-resolved |
| 2026-04-05 | [24012348732](https://github.com/openipc/firmware/actions/runs/24012348732) | hi3516ev200_lite, t20_lite | ipctool-HEAD 404, opus 404 from `downloads.xiph.org` |
[...]

PR #2297, opened 2026-08-22
ci: say which boards failed and why, not just that the nightly is red

### LINKED ISSUE #2271: Decide whether one failing matrix cell should fail the whole nightly run
Split out from #2036, which is now closed — the 404 flakiness that motivated it is fixed, but this follow-up was never picked up.

## Question

`build.yml` builds ~99 matrix cells. The matrix is `fail-fast: false`, so one failing cell does not cancel the others — but the **run** still reports `conclusion: failure`. One board out of 99 turns the nightly red.

Is that the signal we want?

## Why it matters

The failure classes are not equivalent, and today they are indistinguishable at a glance:

| Class | Example | Should it fail the run? |
|---|---|---|
| Real breakage, broad | 2026-08-12 — `majestic-webui` fixup broke all 99 boards | **Yes** |
| Real breakage, one board | 2026-04-06 — `hi3516ev300_neo` kernel `drivers/net/mdio` | Probably yes |
| Transient, survived retry | 404 absorbed by the 7-attempt backoff | Already invisible — correct |
| Transient, exhausted retry | >10min CDN 502 storm on one board | Arguably no |
| Runner infrastructure | GHA "Server Error" | No |

After the #2036 work the nightly sits at 1 failure in 30 days, so the red/green signal is currently meaningful and this is **not urgent**. It was urgent when 30% of runs were red and everyone had learned to ignore the colour. Worth settling before that recurs, not after.

## Things to weigh

- **Alerting on red is only useful if red is rare.** It is rare now. That argues for leaving it alone.
- **A single flaky board can still mask a real one.** With 99 cells, "the nightly is red" does not tell you whether one board or fifty broke. A summary in the job output (counts by failure class) may solve the real problem without touching pass/fail at all.
- **Per-board history is the more useful signal.** "hi3516ev300_neo has been red 3 nights running" is actionable; "the nightly is red" is not.
- Interaction with the CI umbrella gate (shipped 2026-06-07) and with branch protection — changing the workflow conclusion has consequences beyond the nightly.

## Possible outcomes

Closing this as "current behaviour is correct" is a legitimate result. The point is to make it a decision rather than an accident of the default.

- [ ] Decide whether transient-classified failures should report without failing the run.
- [ ] Consider a per-run summary: N passed / N failed, grouped by failure class, in the run summary rather than buried in 99 job logs.
- [ ] Consider per-board streak tracking so a persistently broken board is louder than a one-off.

openipc-ai:
Closes #2271.

## The question, and the answer

`build.yml` builds ~99 matrix cells with `fail-fast: false`. One failing cell does not cancel the
others, but the run still reports `conclusion: failure`. #2271 asked whether that is the signal we
want, and said explicitly that "current behaviour is correct" is a legitimate result — the point
was to make it a decision rather than an accident of the default.

**The polarity does not change. The legibility does.** The reasoning is now a comment in `ci-gate`
rather than an unstated default:

- The nightly's product is a **complete** set of images. A board that failed leaves `nightly` and
  `latest` pointing at yesterday's image for it, so whoever flashes that board silently gets stale
  firmware. Green would be a lie about what got published.
- The transient classes are already absorbed upstream of the gate. Anything that reaches it
  survived seven attempts across ~38 minutes of backoff (#2036); that is no longer a blip.
- `ci-gate` serves pull requests too. Relaxing the matrix rule to quieten the nightly would relax
  branch protection with it.

Measured over the 30 scheduled runs from 2026-07-23 to 2026-08-21: **1 failure**. Red is rare, so
the alert works. The problem was never the colour.

## The actual problem

On 2026-08-12 ([run 31649720825](https://github.com/OpenIPC/firmware/actions/runs/31649720825)) a
`majestic-webui` fixup did a bare `rm` on two files upstream had deleted, and 94 of 99 boards
failed in the same place. Establishing that took opening 94 job logs, because the run's entire
output was the word "failure". "The nightly is red" does not distinguish one broken board from
ninety-four.

## What this does

**1. Names the cause where the evidence is.** Buildroot says which package and which phase in
exactly one line and nowhere else:

```
make[1]: *** [package/pkg-generic.mk:374: .../majestic-webui-dist/.stamp_target_installed] Error 1
```

The retry loop already holds that log, so it now classifies out of it and folds the result into
the `::error::`/`::warning::` it already emitted. **The phase is the class**: `.stamp_downloaded`
is a mirror or a 404 (the #2036 family), anything after it is real code. No new steps, no new
artifacts.

```diff
-echo "::error::build failed after ${attempt} attempts"
+echo "::error::${{matrix.platform}}: build failed after ${attempt} attempts — ${cause}"
```

**2. A `Build summary` job** that aggregates it once, into the run summary:

- *N of M failed*, grouped by cause, so 94-boards-one-bug reads as one row.
- **Flakes the retry loop absorbed** — a board that went green on its third attempt is invisible
  today, and the same cause exhausting its budget is what turns the nightly red. This is the
  leading indicator, and nothing surfaced it.
- **Headroom warnings** rolled up. The dry run against last night's green nightly already found
  `gk7205v300_lite` at **0KB free of 5120KB**.
- **Last 14 scheduled runs** per board, so "`hi3516ev300_neo` has been red three nights running"
  is visible without digging through run history.

Nothing is stored: three read-only endpoints, ~130 requests, and the footer prints the count it
actually used (the token budget is 1000/hr and `publish` already spends ~400 on paced uploads).

**3. It cannot fail the run.** That would be the exact mistake #2271 decided not to make, so the
job is deliberately not in `ci-gate`'s `needs` and its step swallows a non-zero exit into a
warning. What keeps it honest instead is `build-summary.py --self-test` in `lint.yml`, a merge
gate that asserts the grammars still match what `build.yml` emits — a reworded `::error::` would
otherwise reduce the summariser to reporting nothing, and reporting nothing looks exactly like a
quiet night.

## Sample output

Dry-run against [run 32534401852](https://github.com/OpenIPC/firmware/actions/runs/32534401852)
(a green nightly), abridged:

```
## nightly-20260821-...

**All 99 boards built.**

### Headroom warnings

| Board                  | Image           | Free  | Cap     |
|------------------------|-----------------|------:|--------:|
| `gk7205v300_lite`      | rootfs.squashfs |   0KB | 5120KB  |
| `hi3516ev300_lite`     | uImage          |   1KB | 2048KB  |
| `hi3516ev300_ultimate` | uImage          |   1KB | 2048KB  |
| `gk7205v200_lite`      | rootfs.squashfs |  12KB | 5120KB  |
...

### Last 14 scheduled runs

Every board passed in every run in the window.

<sub>108 API request(s).</sub>
```

## Test plan

None of this is reachable from PR CI — the summary job only runs where there is a matrix, which is
the same gap that motivated `lint-workflow-shell.py` — so it was proven against real data before
this PR existed.

- [x] `python3 .github/scripts/build-summary.py --self-test` — 4 grammars, 7 annotations in
      `build.yml`, 99 fixture boards.
- [x] **The drift gate actually catches drift.** Reworded an annotation in `build.yml`:
      `grammar 'exhausted' no longer matches build.yml`. Added an unclassified one:
      `build.yml's matrix job emits 8 annotations, 7 are classified`. Both restored.
- [x] **`classify_failure` against the real 2026-08-12 log**, with the function extracted from the
      workflow YAML rather than retyped → `majestic-webui-dist/target_installed`. Synthetic
      download failure → `opus-1.4/downloaded`; repack failure → `Makefile:210: repack`; nothing
      recognisable → `unclassified`.
- [x] **End to end**: that cause, formatted into the exact annotation `build.yml` now emits, parses
      back to cause `majestic-webui-dist/target_installed` and `attempts: 7`.
- [x] **Dry-run against real runs.** 31649720825 → `92 of 96 boards failed` in 8s / 97 requests
      (it predates the new annotations, so it correctly degrades to `step: Build firmware` rather
      than crashing). 32534401852 → `All 99 boards built`, headroom table above, 15s.
- [x] `python3 .github/scripts/ci-matrix.py --self-test` — ok (99 boards, 132 packages, 52 cases).
- [x] `python3 .github/scripts/lint-workflow-shell.py [--self-test]` — 48 run blocks parse clean.
- [x] `bash .github/scripts/test_load_hisilicon.sh`, `bash .github/scripts/test_sysupgrade.sh` — pass.
- [ ] First nightly after merge: read the rendered summary and check the footer's request count
      against the budget. `--no-flakes` drops the bulk of it if it ever crowds `publish`.

## Notes for review

- `build-summary.py` is added to `ci-matrix.py`'s `NO_BUILD_SCRIPTS`. It cannot change a byte of an
  image, and unclassified it would widen every PR touching it to the full 99-board matrix.
  Touching `ci-matrix.py` to say so widens *this* PR to the full matrix — `ci-matrix.py` is
  deliberately unclassified, since a smoke set would build the boards that were already in
  `ALL_BOARDS` and never a newly added one. That is one full matrix now to stop every future PR
  touching the summariser from paying the same. The `build.yml` edit on its own would have taken
  the 13-board smoke set.
- The annotations endpoint is addressed as `/check-runs/{job_id}/annotations`, which relies on a
  job id being usable as a check run id. That is true today and is verified against both runs
  above, but it is not documented as a guarantee — a 404 there degrades one board to "no detail",
  never to a crash, and the counts stay exact.
- The `lint.yml` self-test is a separate job rather than a step in `workflow run blocks parse`, so
  no existing check context is renamed and branch protection is untouched.
- Stdlib only, so no install step anywhere.

### DIFF
--- a/.github/scripts/build-summary.py
+++ b/.github/scripts/build-summary.py
@@ -1,0 +1,683 @@
+#!/usr/bin/env python3
+"""Render the firmware matrix into one readable summary, per run and over time.
+
+WHY THIS EXISTS. build.yml builds ~99 boards with `fail-fast: false`, so one
+failing cell does not cancel the others -- but the run still reports
+`conclusion: failure`. Issue #2271 asked whether that is the signal we want.
+The answer, recorded in the ci-gate comment in build.yml, is yes: the nightly's
+product is a COMPLETE set of images, and a board that failed leaves `nightly`
+and `latest` holding yesterday's image for it, which is stale firmware for
+whoever flashes that board. The colour is right.
+
+What was wrong was that the colour was the only output. On 2026-08-12 (run
+31649720825) a majestic-webui fixup did a bare `rm` on two files upstream had
+deleted, and 94 of 99 boards failed in the same place -- a fact that took
+opening 94 job logs to establish, because the run said, in full, "failure".
+"The nightly is red" does not distinguish one broken board from ninety-four.
+This file is that distinction.
+
+It also reports the class the retry loop hides. Seven attempts with backoff to
+1200s absorb most transients (#2036), and a green run says nothing about how
+close it came. "Three boards needed retries tonight, all opus-1.4/downloaded"
+is the leading indicator that the nightly is about to go red again, and until
+now nothing surfaced it.
+
+WHERE THE DATA COMES FROM. Nothing is stored. Three read-only API endpoints:
+
+  * /actions/runs/{id}/jobs -- job names are `Firmware (<board>)`, so the board
+    parses out of the name, and steps[] says which step failed, which covers
+    the post-build steps for free.
+  * /check-runs/{job_id}/annotations -- the classes, attempt counts and
+    headroom warnings live here, because build.yml emits them as ::error:: and
+    ::warning::. This relies on a job id being usable as a check run id, which
+    is true today and is verified against real runs, but is not documented as a
+    guarantee. A 404 here degrades that board to "no detail", never to a crash.
+  * /actions/workflows/build.yml/runs?event=schedule -- the streak window.
+    Scheduled runs always build the full matrix (the selector only narrows
+    pull_request), so the window is comparable board-for-board.
+
+Only annotations matching the grammars below are read. That is also what filters
+the Node.js-deprecation warning every job carries: noise is excluded by
+construction rather than by a denylist that would need maintaining.
+
+COST. ~120 requests for a nightly: 2 for the jobs, one per board for the
+annotations, two per historical run. The repo's GITHUB_TOKEN budget is 1000
+requests/hour and the publish job already spends ~400 of it on paced asset
+uploads, so this is not free and the footer prints the count it actually used.
+--no-flakes drops the annotation fetch for boards that PASSED, which is the
+bulk of it, if that ever needs reclaiming.
+
+THIS MUST NEVER FAIL THE RUN. It reports; it does not judge. A bug here turning
+a good nightly red would be precisely the mistake #2271 decided not to make, so
+the workflow step swallows a non-zero exit into a warning and the summary job
+is deliberately not in ci-gate's `needs`. What keeps it honest instead is
+--self-test, which lint.yml runs as a merge gate: it checks that the annotation
+grammars below still match the strings build.yml emits, because a reworded
+annotation silently reduces this file to reporting nothing, and reporting
+nothing looks exactly like a clean nightly.
+
+Usage:
+    build-summary.py --run-id 31649720825 --event schedule
+    build-summary.py --run-id 31649720825 --out -      # stdout, for a dry run
+    build-summary.py --self-test                       # no network
+"""
+
+import argparse
+import json
+import os
+import re
+import sys
+import time
+import urllib.error
+import urllib.request
+from collections import Counter, defaultdict
+from concurrent.futures import ThreadPoolExecutor
+
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+WORKFLOW = os.path.join(REPO_ROOT, ".github", "workflows", "build.yml")
+
+API = "https://api.github.com"
+# Two weeks of nightlies. Long enough that "red three nights running" is
+# visible and a one-off is obviously a one-off, short enough that the window
+# costs ~28 requests.
+HISTORY_RUNS = 14
+# The matrix job's display name. Anything else in the run -- Select boards,
+# Preflight, Publish releases, CI Gate, this job -- is not a board.
+JOB_NAME = re.compile(r"^Firmware \((?P<board>[^)]+)\)$")
+
+# --------------------------------------------------------------------------
+# Annotation grammars
+#
+# Each entry is (name, regex, anchor). The anchor is a literal substring that
+# must still be present in build.yml; --self-test asserts it, so rewording an
+# annotation fails a merge gate instead of silently emptying this report.
+# --------------------------------------------------------------------------
+
+GRAMMARS = [
+    # echo "::error::<board>: build failed after <n> attempts — <cause>"
+    ("exhausted",
+     re.compile(r"^(?P<board>\S+): build failed after (?P<attempts>\d+) attempts — (?P<cause>.+)$"),
+     "build failed after ${attempt} attempts — "),
+    # echo "::warning::<board>: attempt <n> failed (<cause>), retrying after <s>s"
+    ("retry",
+     re.compile(r"^(?P<board>\S+): attempt (?P<attempt>\d+) failed \((?P<cause>.*)\), "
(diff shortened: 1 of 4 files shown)


openipc-ai:
Thanks — all three findings were real. Fixed in 0fce083, plus one wart of my own that the review made me look at.

**1. Racy API counter.** Correct, and it matters more than a lost integer usually would: the counter exists so the API budget is *measured rather than assumed*, and an under-report is precisely the wrong answer — it would say the job is cheaper than it is, right up until it crowds `publish`'s ~400 uploads. `self.calls += 1` is load-add-store across up to eight worker threads. Now under a `threading.Lock`.

**2. Unencoded query parameters.** Every value passed today is a bare word or an int, so nothing is broken now — but the helper reads as a general client and shouldn't be a trap for the next parameter. Replaced with `urllib.parse.urlencode`.

**3. `set(boards)` rebuilt per element.** Taking the suggestion one step further: the loop was grouping *board names* and then going back to `failed` to recover the attempt counts. Grouping the result objects instead drops the set entirely and is shorter:

```python
groups[result["cause"] or "unclassified"].append(result)
...
attempts = max(result["attempts"] for result in group)
boards = [result["board"] for result in group]
```

**4. Not from the review**, but adjacent and worth fixing in the same push: `preflight` computes `build_id` as `nightly-<date>-<sha>` for *every* event, so a pull request's summary was headed with the name of a nightly release that does not exist. Only `schedule` and `workflow_dispatch` use it now; everything else falls back to `Build <run_id>`. Pulled out as `report_title()` so `--self-test` covers it.

Re-verified: `--self-test` passes (4 grammars, 7 annotations, 99 fixture boards, and the new title cases). Dry runs still correct against real runs — 32534401852 as a PR now reads `## Build 32534401852` at **exactly 101 requests** (2 job pages + 99 annotations, which is the count being exact rather than approximately right), and 31649720825 as a schedule still reads `## nightly-20260812-1fa881` / `92 of 96 boards failed`.

openipc-ai:
CI is fully green, and the new job ran for real rather than just being parsed.

| | |
|---|---|
| `Firmware (…)` | **99/99 pass** |
| `CI Gate` | pass |
| `Build summary` | **pass** |
| `build summariser agrees with build.yml` | pass |
| `workflow run blocks parse`, `shipped shell scripts parse`, `sysupgrade rootfs verification`, `load_hisilicon os_mem_size derivation` | pass |
| `qodo-gate` | pass |
| `Publish releases` | skipped (no artifacts on a PR, as designed) |

From the `Build summary` job's own log ([run 32591745450](https://github.com/OpenIPC/firmware/actions/runs/32591745450)):

```
build-summary: 0 of 99 boards failed, 101 API request(s)
```

**101 is the number to look at.** It is exactly 2 job-list pages + 99 annotation fetches, and it matches the local dry run against a different run to the request — which is the point of putting the counter behind a lock. No traceback, no `::warning::build summary failed` fallback, and the job did not touch the gate.

The one test-plan box still open is the post-merge one: read the first nightly's rendered summary and check the request count with the 14-run streak window attached (~130 rather than 101) against the budget `publish` is already drawing on.

PR #2311, opened 2026-08-25
Makefile: pin the host C dialect so toolchains build on GCC 15

openipc-ai:
`make toolchain` (and `make BOARD=<board>` generally) fails on any distro shipping GCC 15.

GCC 15 defaults to `-std=gnu23`, where an empty parameter list means *"takes no arguments"* rather than *"unspecified"*. Several host packages Buildroot pins here predate that. gmp 6.3.0 is the first to hit it — its compiler probe fails with:

```
conftest.c:12:48: error: too many arguments to function 'g'; expected 0, have 6
```

which configure then reports as:

```
configure: error: could not find a working compiler, see config.log for details
```

That message sends you hunting for a missing toolchain rather than a dialect mismatch, and it stops the build before any target package is reached.

## The change

Pin the dialect for host C builds, rather than carrying a version bump for every affected host package. It sits next to the existing `CMAKE_POLICY_VERSION_MINIMUM` export, which solves the same class of problem for CMake's minimum-version policy.

- Host C only — target flags and C++ builds untouched
- Overridable from the environment (`HOST_CFLAGS=... make ...`)
- A no-op on hosts whose GCC still defaults to `gnu17`

## Why CI hasn't seen this

The runners are `ubuntu-22.04`, which ships GCC 11. This only bites people building locally on a current distro — which is now most of them, Ubuntu 25.10 and Fedora 42 included.

## Verification

On Ubuntu with GCC 15.2: `make BOARD=hi3516ev300_lite toolchain` fails at `host-gmp` without this and completes with it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,18 @@ BR_CONF = $(TARGET)/openipc_defconfig
 TARGET ?= $(PWD)/output
 export CMAKE_POLICY_VERSION_MINIMUM := 3.5
 
+# GCC 15 defaults to -std=gnu23, where an empty parameter list means "takes no
+# arguments" rather than "unspecified". Several host packages Buildroot pins
+# here predate that and their configure probes stop compiling: gmp 6.3.0 fails
+# its compiler test with "too many arguments to function 'g'", which surfaces as
+# the far less helpful "could not find a working compiler" and halts any
+# `make toolchain` on a current distro. Pin the dialect rather than carry a
+# version bump for every affected host package. Applies only to host C builds,
+# is overridable from the environment, and is a no-op on hosts whose GCC still
+# defaults to gnu17.
+HOST_CFLAGS ?= -O2 -std=gnu17
+export HOST_CFLAGS
+
 CONFIG = $(error variable BOARD not defined)
 TIMER := $(shell date +%s)
 


openipc-ai:
Not reproducible — the premise is that Buildroot assigns `HOST_CFLAGS` with an ordinary assignment, and it doesn't.

`package/Makefile.in` in 2024.02.10:

```make
250: HOST_CFLAGS   ?= -O2
251: HOST_CFLAGS   += $(HOST_CPPFLAGS)
```

`?=` is conditional assignment, and an imported environment variable counts as already defined, so line 250 is skipped and line 251 appends to our value. The rule about makefile assignments winning over the environment applies to `=` and `:=`, not `?=`.

Buildroot's own `printvars` resolves it, run against this repo's tree both ways:

```
$ make -C output/buildroot-2024.02.10 ... printvars VARS=HOST_CFLAGS
HOST_CFLAGS=-O2 -I/home/ai/git/firmware/output/host/include

$ HOST_CFLAGS="-O2 -std=gnu17" make -C output/buildroot-2024.02.10 ... printvars VARS=HOST_CFLAGS
HOST_CFLAGS=-O2 -std=gnu17 -I/home/ai/git/firmware/output/host/include
```

The second is what `export HOST_CFLAGS` in this PR produces for the recursive `$(MAKE) -C`.

And end to end, which is how the change was arrived at: `make BOARD=hi3516ev300_lite toolchain` fails at `host-gmp` on a GCC 15.2 host without it and completes with it, with `-std=gnu17` appearing in 3727 host compiles in the build log.

I did briefly mislead myself while checking this — my first `printvars` run invoked Buildroot's make directly, so this repo's `export` never happened and the flag was absent, which looks exactly like the reported bug. The A/B above is the corrected test.

No change made. Happy to switch to an explicit `HOST_CFLAGS=...` on the recursive `$(MAKE)` line if you'd prefer it not to depend on `?=` semantics — it would be equivalent here, just more literal.

openipc-ai:
### Dialect pin is overridden — refuted, with the line and a build to back it

The premise is right for a plain `=`, but Buildroot uses `?=`, and that is the
whole difference. `package/Makefile.in` in the pinned 2024.02.10:

```make
250:HOST_CFLAGS   ?= -O2
251:HOST_CFLAGS   += $(HOST_CPPFLAGS)
252:HOST_CXXFLAGS += $(HOST_CFLAGS)
```

`?=` is a no-op when the variable is already defined, and a variable imported
from the environment *is* defined. So line 250 does not fire, line 251 appends
Buildroot's include flags to the value this PR exported, and line 252 carries it
into host C++ too. Nothing is replaced.

That is also measurable rather than inferred. Today's `make BOARD=t31_lite
toolchain-asan` on a GCC 15 host — the exact failure this PR exists for — ran to
completion, and `-std=gnu17` appears **6321 times** in its log, on the real host
configure and compile lines:

```
(cd .../build/host-tar-1.34/ && rm -rf config.cache; ... CFLAGS="-O2 -std=gnu17 ...
cc1plus: warning: command-line option '-std=gnu17' is valid for C/ObjC but not for C++
```

The environment was the only source of that value, and the `cc1plus` warning in
particular could not appear at all if Buildroot's own assignment had won — it is
line 252 doing its job.

And the export itself, branch versus master, through a recipe's environment,
which is what the recursive `$(MAKE) -C` inherits:

```
host-gcc15-c23:  recipe env HOST_CFLAGS=[-O2 -std=gnu17]
master:          recipe env HOST_CFLAGS=[]
```

Worth saying plainly, because I got this wrong once on this PR already: my first
attempt to check it invoked Buildroot's `make` directly, which never runs this
repo's Makefile and so never runs the `export`. That test showed the flag
absent and looked like a confirmation of the finding. It was measuring nothing.
Anything checking an `export` has to enter through the top-level Makefile.

Merging as-is.

PR #2314, opened 2026-08-26
ci: publish an ASan toolchain for the boards where it is proven

openipc-ai:
Stacked on #2312 — retarget to `master` once that merges. Without those
libsanitizer patches a t31 ASan toolchain builds and then hangs before `main()`,
so the two are not independently useful.

## What this adds

`make BOARD=<board> toolchain-asan` has existed since #2003 and produces a gcc
with a sanitizer runtime. What has been missing is delivery: every agent wanting
to leak-check majestic on a musl board builds its own gcc first — eleven minutes
— and then has no way to hand the result to the next person.

- `.github/workflows/toolchain-asan.yml` — dispatch it, get
  `toolchain.<vendor>-<family>-asan.tgz` under a `toolchain-asan` release tag.
- `general/scripts/verify-asan-toolchain.sh` — the checks, as a script rather
  than a block of YAML, so whoever builds one by hand runs exactly what CI runs.

majestic's `build.sh` downloads the `-asan` variant on its own for any
non-`Release` build, so on a listed board the whole procedure becomes
`scripts/asan.sh smoke --host <camera>`.

The first toolchain is already published, built from this branch and verified on
the lab t31:
https://github.com/OpenIPC/firmware/releases/tag/toolchain-asan

## Three things that are deliberate

**The board list is a `choice`, not free text, and holds only `t31_lite`.** An
entry means someone ran majestic's `scripts/asan.sh smoke` on that hardware and
the sanitizer reported the planted faults. Building a toolchain proves nothing
about whether its sanitizer *reports*: five of the six mips32 defects fixed in
#2312 announce themselves, and the sixth hands you a clean run with exit 0.

**No "skip if the asset already exists".** `toolchain.yml` does that and across
thirty-odd boards it is right. Here the entire value of the artifact is the
libsanitizer patches it carries, and those change; a run that quietly declined
to rebuild would keep serving a toolchain missing the fix that was the reason
for dispatching it.

**It verifies before it uploads.** That the patches applied to the gcc version
Buildroot actually selected, that the packaged tarball carries a runtime, and
that its compiler links `-fsanitize=address`.

The first of those is the one that matters. The patches live in per-version
directories, so a Buildroot bump moves gcc out from under them, Buildroot
applies nothing, and the build still succeeds — producing a toolchain silently
unable to find a leak. Same failure mode as that sixth defect, by a different
route.

That is not a hypothetical, and the script demonstrates it. Run against
yesterday's t31 build tree with today's patch set:

```
FAIL 0004-libsanitizer-mips-ptrace-stack-pointer.patch ships for gcc 13.3.0 but was not applied
FAIL 0005-libsanitizer-no-unwinder-reentry.patch ships for gcc 13.3.0 but was not applied
...
--   6 of 9 shipped patches did not apply.
ok   runtime libasan.so.8 1460928 bytes
ok   mipsel-openipc-linux-musl-gcc links -fsanitize=address
```

**Both cheap checks pass on the broken toolchain.** Only the patch check
catches it.

## Verification

`general/scripts/verify-asan-toolchain.sh` on the published artifact:

```
ok   9 libsanitizer patches applied for gcc 13.3.0
ok   runtime libasan.so.8 1494724 bytes
ok   mipsel-openipc-linux-musl-gcc links -fsanitize=address
```

And on the lab t31 (mipsel o32 musl, 64 MB, `--low-memory`), the part no script
can answer — majestic's `scripts/asan.sh smoke`:

```
AddressSanitizer: heap-buffer-overflow ... READ of size 1  -> asan_probe_overflow
AddressSanitizer: heap-buffer-overflow ... WRITE of size 1 -> asan_probe_overflow
LeakSanitizer: detected memory leaks ... 81 byte(s)        -> asan_probe_leak_here
3 suspended thread(s) checked, every stack pointer inside its stack
```

`.github/scripts/lint-workflow-shell.py` passes; shellcheck clean under `-s sh`.

### DIFF
--- a/.github/workflows/toolchain-asan.yml
+++ b/.github/workflows/toolchain-asan.yml
@@ -1,0 +1,73 @@
+name: toolchain-asan
+on:
+  workflow_dispatch:
+    inputs:
+      platform:
+        description: 'Board to build the ASan toolchain for'
+        required: true
+        default: t31_lite
+        type: choice
+        options:
+          # Only boards where ASan has been proven end-to-end on real hardware
+          # with `scripts/asan.sh smoke` in the majestic repo. Building a
+          # toolchain is easy; knowing its sanitizer actually reports a leak on
+          # that SoC is not, and an unproven entry here would ship a green run
+          # that means nothing. Add a board once, and only once, smoke has
+          # reported the planted leak on it.
+          - t31_lite
+
+env:
+  TAG_NAME: toolchain-asan
+
+# Declared rather than inherited, the same way lint.yml argues for: the upload
+# step needs to create a release, and nothing else here needs any of it.
+permissions:
+  contents: write
+
+jobs:
+  toolchain-asan:
+    name: ASan toolchain (${{inputs.platform}})
+    runs-on: ubuntu-22.04
+    # A gcc built with libsanitizer runs a little over an hour on a runner. The
+    # default is six hours, which is long enough for a step that has hung rather
+    # than failed to hold a runner for most of a working day before saying so.
+    timeout-minutes: 150
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      # Deliberately no "skip if the release asset already exists" shortcut,
+      # which is what toolchain.yml does across its thirty-odd boards. The whole
+      # value of this artifact is the libsanitizer patches it carries, and those
+      # change; a run that quietly declined to rebuild would keep serving a
+      # toolchain missing the fix that was the reason for dispatching it.
+      - name: Build toolchain
+        run: |
+          GCC=$(make BOARD=${{inputs.platform}} toolname)-asan.tgz
+          make BOARD=${{inputs.platform}} toolchain-asan
+          SDK=$(find output/images -name '*_sdk-buildroot.tar.gz')
+          mv "${SDK}" "${GCC}"
+          echo "GCC=${GCC}" >> ${GITHUB_ENV}
+
+      # A toolchain that builds is not a toolchain that sanitizes. The checks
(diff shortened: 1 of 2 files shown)


openipc-ai:
Addressed in `c0a09233`.

### Workflow targets only `t31_lite`

Fair reading of the word, and now fixed — though the artefact was never
board-specific. `make toolname` yields `toolchain.<vendor>-<family>`, so
`t31_lite` and `t31_ultimate` produce the *same* toolchain; the defconfig was
only where its configuration is read from, and majestic maps t20, t21, t23 and
t31 to that one toolchain.

Rather than argue the point I made the file say what it means. The input is a
**family** now:

```yaml
      family:
        description: 'SoC family to build the ASan toolchain for'
        default: ingenic-t31
        type: choice
        options:
          - ingenic-t31
```

mapped to whichever defconfig on the family is convenient, with an assertion
that the mapping still yields the family it claims:

```sh
test "${NAME}" = "toolchain.${{inputs.family}}" || {
  echo "::error::${BOARD} yields ${NAME}, not toolchain.${{inputs.family}}"
  exit 1
}
```

A drifted mapping would otherwise publish an artefact under a name no consumer
asks for, which is a worse outcome than a build failure.

The list stays a closed `choice` for the reason it was closed before: an entry
means ASan has been proven on that family's hardware with majestic's
`scripts/asan.sh smoke`, not merely that gcc built. Five of the six mips32
defects in #2312 announce themselves; the sixth hands you a clean run with
exit 0. An unproven entry would publish a toolchain that reports nothing.

`.github/scripts/lint-workflow-shell.py` still passes.

PR #2315, opened 2026-08-26
xm530: autoload the SD card driver at boot

yatotoshka:
## Problem

On XM530 boards the vendor SD card driver (Arasan SDHC, kernel module
`sdio0_sd`) ships in the image but is never loaded: `general/overlay/etc/modules`
only lists `vfat` and `exfat`, so `S35modules` never modprobes it. With a card in
the slot the kernel shows no mmc controller and no `/dev/mmcblk0`; mdev's
automount helper (`/lib/mdev/automount.sh`) and majestic's record path
(`/mnt/mmcblk0p1/%F`) both expect the card at `/dev/mmcblk0p1`, so SD storage is
dead out of the box.

The module entry is appended to `/etc/modules` by a late post-build hook keyed on
`BR2_PACKAGE_XIONGMAI_OSDRV_XM530` (xm530 and xm550 defconfigs), because
`general/overlay/etc/modules` is shared by every SoC and must stay board-agnostic.
`check_target_modules.sh` additionally asserts `sdio0_sd.ko` lands in the image
whenever the package is selected.

## Hardware tested on

XM530AI (marking 30WX1), board IPC-RB-BLK530AI-0235P-AB0 V1.03, "anbiux-A8B-3mp",
128GB FAT32 (SDXC) card.
```
root@openipc-xm530:~# ipctool 
The ipctool installed as remote GitHub plugin
---
chip:
  vendor: Xiongmai
  model: XM530
board:
  vendor: Xiongmai
ethernet:
  mac: "f4:b1:9c:a8:ca:1f"
rom:
- type: nor
  block: 64K
  partitions:
    - name: boot
      size: 0x40000
      sha1: 69533542
      contains:
        - name: xmcrypto
          offset: 0x2fc00
        - name: uboot-env
          offset: 0x30000
    - name: env
      size: 0x10000
      sha1: 71a375d2
    - name: kernel
      size: 0x200000
      sha1: 0e79e563
    - name: rootfs
      size: 0x500000
      path: /,squashfs
      sha1: dc91c741
    - name: rootfs_data
      size: 0xb0000
      path: /overlay,jffs2,rw
  size: 8M
ram:
  total: 64M
  media: 29M
firmware:
  kernel: "3.10.103+ (SMP Tue Aug 18 23:09:10 UTC 2026)"
  toolchain: buildroot-gcc-13.3.0
  main-app: /tmp/majestic
sensors:
- vendor: SmartSens
  model: SC3335
  control:
    bus: 0
    type: i2c
    addr: 0x60
root@openipc-xm530:~# 

```
## Evidence

Before — card inserted, kernel has no idea it exists:

```
root@openipc-xm530:~# cat /proc/partitions
major minor  #blocks  name
  31        0        256 mtdblock0
  31        1         64 mtdblock1
  31        2       2048 mtdblock2
  31        3       5120 mtdblock3
  31        4        704 mtdblock4
root@openipc-xm530:~# dmesg | grep -ci mmc
0
```

After (this patch, verified across a reboot):

```
arasan_set_clock:
	new freq 400000
mmc0: driver initialized... IRQ: 73, Base addr 0xc4e00000
arasan_set_clock:
	new freq 50000000
mmc0: new high speed SDXC card at address b368
mmcblk0: mmc0:b368 TW000 116 GiB
 mmcblk0: p1

root@openipc-xm530:~# df -h /mnt/mmcblk0p1
Filesystem       Size  Used Avail Use% Mounted on
/dev/mmcblk0p1   116G  32K  116G   1% /mnt/mmcblk0p1
```

Note: the kernel only supports vfat for the card (exfat is not built in and has
no module in this tree), so the card must be FAT32; this patch does not change
that.

## Scope

### DIFF
--- a/.github/scripts/check_target_modules.sh
+++ b/.github/scripts/check_target_modules.sh
@@ -40,6 +40,7 @@ check_module() {
 }
 
 check_module BR2_PACKAGE_WIREGUARD_LINUX_COMPAT wireguard.ko
+check_module BR2_PACKAGE_XIONGMAI_OSDRV_XM530 sdio0_sd.ko
 
 if [ "$fail" -ne 0 ]; then
 	echo "check_target_modules: regression detected" >&2
--- a/.github/scripts/check_target_modules.sh
+++ b/.github/scripts/check_target_modules.sh
@@ -25,13 +25,13 @@ fi
 
 fail=0
 
-# config-var → expected .ko filename
+# config-var → expected .ko path (relative to lib/modules/<version>/)
 # Add new pairs here when a kernel-module package is introduced.
 check_module() {
 	var="$1"
 	ko="$2"
 	grep -q "^${var}=y" "$CONFIG" || return 0
-	if find "$TARGET_DIR/lib/modules" -name "$ko" | grep -q .; then
+	if find "$TARGET_DIR/lib/modules" -path "*/$ko" | grep -q .; then
 		echo "OK: $ko present (${var}=y)"
 	else
 		echo "MISSING: ${var}=y but $ko not found under $TARGET_DIR/lib/modules/" >&2
@@ -40,7 +40,7 @@ check_module() {
 }
 
 check_module BR2_PACKAGE_WIREGUARD_LINUX_COMPAT wireguard.ko
-check_module BR2_PACKAGE_XIONGMAI_OSDRV_XM530 sdio0_sd.ko
+check_module BR2_PACKAGE_XIONGMAI_OSDRV_XM530 xiongmai/sdio0_sd.ko
 
 if [ "$fail" -ne 0 ]; then
 	echo "check_target_modules: regression detected" >&2
--- a/general/package/xiongmai-osdrv-xm530/post-build-hook.sh
+++ b/general/package/xiongmai-osdrv-xm530/post-build-hook.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+# xiongmai-osdrv-xm530 post-build hook.
+#
+# The vendor SD card driver (Arasan SDHC, module name sdio0_sd) ships in the
+# image but nothing ever loads it: it is not listed in /etc/modules, so
+# S35modules never modprobes it and /dev/mmcblk0 does not appear even with a
+# card in the slot. mdev's automount helper and majestic's record path expect
+# the card at /dev/mmcblk0p1.
+#
+# general/overlay/etc/modules is shared by every SoC, so the entry is appended
+# here, only for the images that select this package. The hook must run after
+# the rootfs overlay is applied, hence late-post-build-hooks.list rather than
+# the package's INSTALL_TARGET_CMDS.
+set -eu
+
+TARGET_DIR="${1:?target dir required}"
+
+MODULES="${TARGET_DIR}/etc/modules"
+if [ ! -f "${MODULES}" ]; then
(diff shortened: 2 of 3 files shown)


openipc-ai (changes requested):
**The new module check can't fail, so it doesn't guard what it's meant to.**

`br-ext-chip-xiongmai/board/xm530/xm530.generic.config` carries `CONFIG_MMC_SD=m`, and OpenIPC/linux `xiongmai-xm530` has `obj-$(CONFIG_MMC_SD) += sdio0_sd.o` in `drivers/mmc/host/Makefile`. This PR's own `xm530_lite` run builds and installs it:

```
  LD [M]  drivers/mmc/host/sdio0_sd.ko
  INSTALL drivers/mmc/host/sdio0_sd.ko
```

So the image holds two modules called `sdio0_sd.ko` — that one under `kernel/drivers/mmc/host/`, and the vendor blob the package installs to `xiongmai/`. `check_module` does `find "$TARGET_DIR/lib/modules" -name "$ko"`, which matches the kernel-built copy regardless of whether `xiongmai-osdrv-xm530` shipped its own. If the package's `INSTALL_TARGET_CMDS` ever stopped installing it, the check would still print `OK` — exactly the #2032-class regression the script exists to catch.

Scoping the match by path fixes it without disturbing the wireguard entry:

```diff
-	if find "$TARGET_DIR/lib/modules" -name "$ko" | grep -q .; then
+	if find "$TARGET_DIR/lib/modules" -path "*/$ko" | grep -q .; then
…
-check_module BR2_PACKAGE_XIONGMAI_OSDRV_XM530 sdio0_sd.ko
+check_module BR2_PACKAGE_XIONGMAI_OSDRV_XM530 xiongmai/sdio0_sd.ko
```

`wireguard.ko` still matches as `*/wireguard.ko`.

Worth knowing which of the two actually loads, since it isn't the in-tree source: with both present, `depmod` (kmod 34.2) writes only `xiongmai/sdio0_sd.ko:` into `modules.dep` — the external directory wins over `kernel/`, it still wins when `modules.order` names the kernel path, and it isn't alphabetical. So `modprobe sdio0_sd` loads the vendor blob and the kernel-built copy is shadowed dead weight.

Two smaller things while you're pushing, since touching `late-post-build-hooks.list` re-runs the full board matrix each time:

- The description implies SD works out of the box, but the driver loads with `detect=0`, so `sdio0_powerup()` leaves GPIO49 (`sdio0_detect`) unmuxed and `need_detect`/`need_poll` are both 0. A card present at boot enumerates, as your evidence shows; one inserted afterwards has no detect path. Worth saying hot-plug is out of scope — `/etc/modules` does take arguments (`sdio0_sd detect=1`) if it shouldn't be.
- Have you been able to try this on a second XM530 board, ideally one without a card slot? `sdio0_powerup()` muxes GPIO43-48 and 51-52 to SDIO and drives GPIO50 output-low unconditionally, and this makes that a default on every xm530/xm550 image. If those pads are free on all XM530 designs that's fine, but it'd be good to have it stated.

Everything else looks right to me — the hook is idempotent, the `late-post-build-hooks.list` placement genuinely does run after the overlay copy, and `S35modules` → `S38mdev` orders correctly (mdev sets the hotplug handler before `mdev -s`, so the coldplug can't race the card).

openipc-ai (approved):

PR #2316, opened 2026-08-26
xm530: bring up the AltoBeam ATBM6032 USB WiFi chip

yatotoshka:
## Problem

Some XM530 boards (e.g. IPC-RB-BLK530AI-0235P-AB0 V1.03) carry an AltoBeam
ATBM6032 USB WiFi (007a:8888) with a power-down gpio. The stock xm530 image has
neither the driver nor a `/etc/wireless/usb` case for it, so WiFi is not usable
on these boards.

## What this change does

- Enables the existing `atbm60xx` package in `xm530_lite_defconfig` (603x model,
  USB interface) — the same three-line pattern as `t31_ultimate_defconfig` — and
  pins `ATBM60XX_VERSION` to a full 40-character SHA instead of moving `HEAD`.
- Adds the `atbm603x-xm530-usb` case to `/etc/wireless/usb`: `modprobe dwc_otg`,
  then `modprobe wifi_pdn value="$pdn"` where the PDN gpio is read from the
  per-board `wifipdn` env (set e.g. by a builder profile's customizer), then
  `modprobe atbm603x_wifi_usb`. `S40network` then applies the MAC and runs
  `ifup wlan0` exactly as for every other `wlandev`. The case is purely
  additive and follows the file's established per-board pattern.
- Re-runs `depmod` in `xiongmai-osdrv-xm530`'s target-finalize: the vendor
  kernel ships a `modules.dep` that lists only its own modules, leaving
  `dwc_otg`, `wifi_pdn` and the atbm module invisible to `modprobe` (same
  pattern as hisilicon-opensdk).
- Installs the vendor `cfg80211` rewrite as `cfg80211_xm711.ko` instead of over
  the in-tree `cfg80211` (`CONFIG_CFG80211=m`) at the canonical module path, and
  loads it from there in the `wifi xm711` helper. atbm60xx — and the
  kernel-built mac80211 — build against the in-tree module, so both WiFi stacks
  stay usable in one image.

Companion profile in [OpenIPC/builder#130](https://github.com/OpenIPC/builder/pull/130)
(`xm530_lite_anbiux-a8b-3mp`): on first boot sets `wlandev=atbm603x-xm530-usb`,
`wifipdn` to the board's PDN gpio and the upgrade url to this device's builder
release, so the camera works and keeps updating out of the box. This PR is the
fallback: with it merged, any xm530 with the chip works via `fw_setenv
wlandev=atbm603x-xm530-usb wifipdn=<gpio> wlanssid=... wlanpass=...
wlanmac=...` even without the builder profile.

## Status (2026-09-06)

- Rebased on the current upstream master (`317b443d`); the branch now sits on
  top of the merged SD-card autoload fix (#2315), so an image built from it also
  carries the `sdio0_sd` autoload.
- Re-verified on the camera from a factory-clean state (env wiped) on the
  latest nightly (`master+4c34a66`): the radio associates and takes a DHCP
  lease — see Evidence.

## Hardware tested on

XM530AI (marking 30WX1), board IPC-RB-BLK530AI-0235P-AB0 V1.03, SmartSens
SC3335, AltoBeam ATBM6032 USB WiFi (007a:8888), PDN on gpio 96

## Evidence

The chip, the PDN gpio and the exact load order used by the new case
(dwc_otg → wifi_pdn → atbm603x) are proven on this hardware — the camera is
currently associated:

Before (stock image, no driver for the chip at all):

```
# dmesg — no atbm driver registered, wlan0 never appears
ls /sys/class/net/
eth0  lo
```

After (radio driven, wlan0 up):

```
wlan0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP qlen 1000
    link/ether f4:b1:9c:a8:ca:20 brd ff:ff:ff:ff:ff:ff
    inet 192.168.1.243/24 brd 192.168.1.255 scope global wlan0
[atbm_log]:wlan0: authenticated
[atbm_log]:wlan0: associated
default via 192.168.1.1 dev wlan0
```

The Web UI is reachable both over the wlan0 address (192.168.1.243) and the
Ethernet address (192.168.1.242).

Honest status: the running camera currently loads the driver from a prebuilt
blob (built from the same atbm_60xx revision this PR pins) plus a local init
script, not from the clean-build path of this patch (compiled `atbm60xx` +
`wlandev` + `S40network` + the `wireless/usb` case). That path is still
unverified on-device; CI confirms atbm60xx compiles against the xm530 kernel,
and the size report says whether the 5M rootfs still fits (the image now also
carries the #2315 SD module). To close the gap: build an image from this branch
— possible today via builder#130's `build-one` with `firmware_repo` pointed at
this fork and `firmware_ref=xiongmai-atbm60xx-wifi`, or right after this PR is
merged — flash it and verify.

Known trade-off: enabling atbm60xx in the shared defconfig grows every xm530
image by the driver module. The cfg80211 rename leaves stock images carrying
the vendor rewrite as `cfg80211_xm711.ko` alongside the in-tree `cfg80211`
(previously it overwrote it), with the `wifi xm711` helper loading the renamed
copy, so xm711/lynx boards keep working at the cost of one extra module file.
The only xm530 profile proposed in OpenIPC/builder today ([builder#130]) is
this ATBM board; no xm711-equipped xm530 is registered or known to be
deployed. Both hunks stand alone and can be dropped or split if reviewers
disagree.

## Scope

Note on the overlay checkbox: the new `wireless/usb` case is purely additive
and matches the file's established per-board pattern (driver, SoC, board
bring-up — like the ~40 existing cases); it changes no existing board's
behaviour and only runs when `wlandev` is set to `atbm603x-xm530-usb`. The PDN
gpio is read from the per-board `wifipdn` env at run time — set by the builder
profile's customizer or the user, not hardcoded in the shared script.


flyrouter:
Good afternoon 
Thank you for your interest in our project and the PR you sent. 
Unfortunately, we won't be able to accept it because we have a separate repository for board customisation - Builder. Please take a look at it and create your model profile there, that would be the right approach. 
Thanks.

flyrouter:
By the way, I found a few patches in Builder that I haven't sent yet, and I'll try to upload them tomorrow or at the latest in the next few days. And maybe this will help us create a stable solution together. Thanks.

Issue #1953, opened 2026-03-14
SSC377QE - 30fps only, no cropping, majestic unstable


BlueMax:
Majestic is crashing occasionally (no errors), cropping is not working and fps are limited to 30fps. 

~~Defconfig reads `BR2_cortex_a35=y` but the SSC37X are Cortex-A53.~~

Any advice?

IM50Q01
NOR16
256MB
lite image

Edit: 
I've swapped `mi.ko` (chagall.bin, iqfile0.bin) to my vendor ones as they are newer (20240618<>20241230). ~~No majestic crashes so far but needs more testing.~~ _Still crashes..._

Regarding fps issues:
**majestic set to 60fps (or lower):** 
Sensor index 1: 2560x1920@60fps (Majestic log)
This mode is 4:3 and will stretch 2560x1920 to any resolution you put in. Any manual crop (like 0x0x1920x1080) results in null rtsp video stream. The fps is capped to 30 (or lower). 

**majestic set to 120fps:** 
Sensor index 3: 1920x1080@120fps
This one is 16:9 cropped 1920x1080@120fps. Looks ok. Not sure though if its really 1080p (shouldn't it be 720p@120?).

I guess `Sensor index 2` mode is missing (60fps mode?).

Is there an option to tweak this with majestic? Auto Focus (AF) Auto Exposure (AE) Auto White Balance (AWB).
`[MI_ISP_CUS3A_Enable] AE = 1, AWB = 1, AF = 1`

Edit2:
Here is a log from majestic crash. Looks like a `wget` attempt is the culprit here (randomly timed telemetry data?).
```
root@openipc-ssc377qe:~# majestic
19:28:21  <       majestic> [    main] main@237                      Majestic Lite for SigmaStar, version master+47984cb, built on 2026-03-11
19:28:21  <       majestic> [app_conf] parse_app_config@404          Main audio codec is opus (other options: aac, pcm, alaw, ulaw)
19:28:21  <       majestic> [watchdog] watchdog_start@94             Detected SSTAR Watchdog
19:28:21  <       majestic> [watchdog] watchdog_start@115            Watchdog timeout set to 10 seconds
19:28:21  <       majestic> [   httpd] start_web_server@373          Set server path: /var/www
19:28:21  <       majestic> [   httpd] new_http_server@359           HTTP server started on 0.0.0.0:80
[...]

PR #2343, opened 2026-08-31
sigmastar: load the infinity6c imx335 with four lanes

widgetii:
Companion to **OpenIPC/sensors#3**, which gives infinity6c its own imx335 driver instead of the symlink to the pudding one.

That driver reads the `lane_num` module parameter, and `drv_sensor_common.h` in `openipc/sensors` defaults it to **2**. At 2 every mode above 30 fps silently selects `pCus_init_mipi2lane_5m30fps_linear` — so the camera reports `2560x1920@60` in its mode table, answers 60 to anything that asks, and serves **30**. Nothing in the log says otherwise. The Sony datasheet is clear that all-pixel 60/50 fps is unavailable on two lanes at all.

Stated in the loader rather than by changing that default, because the default is shared: about twenty drivers across infinity6/6b0/6e read the same symbol and none of them can be tested here. **The other four families' copies of this script are deliberately untouched.**

## Safe in either order

This is inert against the driver in `openipc/sensors` today, which hardcodes `SENSOR_MIPI_LANE_NUM` to 4 and ignores the parameter. So it can land before or after the sensors change — neither ordering breaks anything, and the modes appear once both are in.

## Measured

SSC377QE + IMX335, **rebooting before every mode** (the driver assumes power-on register defaults, so hot-switching gives invalid numbers):

| mode | before | after |
| --- | --- | --- |
| 2592x1944@59 | *no such mode* | **59.2 fps** |
| 2560x1920@60 | 30.0 | **60.1** |
| 2208x1248@90 | ~3, no usable output | **89.5** |
| 1920x1080@120 | 114.6 | **116.1** |

No ISP or SoC overclocking, and no `dnrlevel` change, were needed.

Refs: #1953

### DIFF
--- a/general/package/sigmastar-osdrv-infinity6c/files/script/load_sigmastar
+++ b/general/package/sigmastar-osdrv-infinity6c/files/script/load_sigmastar
@@ -21,7 +21,29 @@ detect_sensor() {
 
 set_sensor() {
 	case $SENSOR in
-		gc4653|imx335|imx415|os04a10|sc4336p|sc401ai|sc501ai|sc830ai|sc850sl)
+		imx335)
+			# This driver reads lane_num, and drv_sensor_common.h defaults it
+			# to 2 -- at which point every mode above 30fps silently selects
+			# the 2-lane 30fps table and the camera serves 30fps while its
+			# mode list still advertises 60.
+			#
+			# Four is the default here because four is what the previous
+			# driver used unconditionally (it hardcoded the lane count and
+			# ignored this parameter), so every board that works today is
+			# already running four and keeps doing so. A board actually wired
+			# for two sets sensor_lanes=2 in U-Boot, as with sensor_mclk.
+			lanes="$(fw_printenv -n sensor_lanes 2>/dev/null)"
+			case "$lanes" in
+				2|4) ;;
+				"") lanes=4 ;;
+				*)
+					echo "sensor_lanes=$lanes is not 2 or 4; using 4" | logger -s -t OpenIPC
+					lanes=4
+					;;
+			esac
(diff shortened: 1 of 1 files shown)


widgetii:
Fair finding, and fixed in `b43ed97` rather than argued away — though the premise needs one correction.

**The correction:** a two-lane infinity6c IMX335 board cannot be working today either. The driver currently shipped for this family hardcodes `#define SENSOR_MIPI_LANE_NUM (4)` and *ignores* `lane_num` entirely, so every such board is already running four lanes. So this cannot cause a board to "lose video after upgrade" — there is no two-lane board in a working state to regress.

**But the point still stands for the future**, which is why it is worth fixing: once the companion change lands and the driver starts honouring `lane_num`, a genuine two-lane board would want 2, and hardcoding 4 in a shared script would deny it that. That is a real loss of capability, and rule 5 is right that a shared `load_<vendor>` script is the wrong place to assert one board's topology.

So the value is now a per-board U-Boot override with four as the fallback, following the `sensor_mclk` precedent already in `load_goke` — validated, and a bad value is logged and ignored rather than passed to `insmod`:

```sh
lanes="$(fw_printenv -n sensor_lanes 2>/dev/null)"
case "$lanes" in
    2|4) ;;
    "") lanes=4 ;;
    *) echo "sensor_lanes=$lanes is not 2 or 4; using 4" | logger -s -t OpenIPC; lanes=4 ;;
esac
insmod $MODULE/sensor_${SENSOR}_mipi.ko chmap=1 lane_num=$lanes
```

Four remains the default because four is what every working board already runs — the change states the number instead of implying it. Two-lane boards get a documented escape hatch they did not have before.

Verified: unset → 4, `2` → 2, `4` → 4, `junk` → warns and uses 4.

PR #2357, opened 2026-09-03
package/majestic-af: build the out-of-core autofocus plugin from source

widgetii:
## Problem

majestic's autofocus engine moved out of the core binary into the open-source
[OpenIPC/majestic-af](https://github.com/OpenIPC/majestic-af) plugin
(widgetii/majestic#553). Nothing in firmware builds or ships that plugin, so the
engine is unavailable. This adds a Buildroot package that **compiles it from
source** (like `motors`) and selects it for `hi3516ev300_lite`, so the 85H50AI
board — whose builder profile enables autofocus (OpenIPC/builder#142) — gets
working autofocus out of the box.

## Hardware tested on

`hi3516ev300` + XiongMai 85H50AI (motorized 2.8–12 mm lens) — the board the
engine was developed and proven on.

## Evidence

The `.so` from **this package** was built and installed into a real image by the
builder companion (OpenIPC/builder#145): its full `hi3516ev300_lite_xm-85h50ai`
device firmware build compiled `majestic-af.so` from source and installed it —
**passed (5m18s)**.

Runtime behaviour was validated on the target with the same sources + same
`hi3516ev200` toolchain during widgetii/majestic#553:

Before (no plugin `.so`):
```
GET /autofocus        -> unavailable        (majestic keeps running normally)
```
After (`/usr/lib/majestic-af.so` present, autofocus enabled):
```
GET /autofocus        -> done fv=9520 peak=9520 mag=-1.0 pos=20 path=2   # cold focus
GET /zoom?tele        -> zooming
GET /autofocus/status -> done fv=9716 peak=10104 mag=1.7 pos=1280        # zoom + refocus
GET /zoom             -> mag=1.7 age_ms=...                              # magnification -> OSD
kill -HUP <pid> mid-pass -> same pid alive, no crash, AF works after reload
```

## Scope

- [x] No files specific to a single retail camera model (the package is generic; the board selection is a one-line generic-SoC defconfig — the device profile lives in OpenIPC/builder)
- [x] Nothing under `general/overlay/` or a shared `load_<vendor>` hardcodes a board value
- [x] Package sources come from an OpenIPC repository (`OpenIPC/majestic-af`), pinned to a full 40-character SHA
- [x] No `LD_PRELOAD`, and no binaries that cannot be rebuilt from source (compiled from source here)
- [x] New code is selected by a defconfig (`hi3516ev300_lite`), so CI builds it


widgetii:
Thanks — responses to the review:

1. **Hardware evidence** — the engine was validated extensively on this exact board (hi3516ev300 + 85H50AI) as part of widgetii/majestic#553: cold focus, zoom-then-refocus, magnification to the OSD, and a mid-pass config-reload (SIGHUP) with clean thread teardown, no crash. The builder companion PR (OpenIPC/builder#145) runs the full device firmware build.

2. **No defconfig selection / 4. selectable without plugin support** — this is deliberate. `majestic-af` is a *device-specific* package: its `BR2_PACKAGE_MAJESTIC_AF=y` lives in the OpenIPC/builder device defconfig (`hi3516ev300_lite_xm-85h50ai`), which builds and CI-tests it there — hence `NOT_BUILT` here, like the other overlay/device-selected packages. There is no Kconfig symbol for "majestic built with plugin-symbol export": that's a property of the per-family S3 build (this board's `hi3516ev200`-family majestic is compiled `WITH_PLUGINS_SUPPORT`), so it can't be a `depends on`. If the seams aren't exported the plugin's `dlopen` simply fails and majestic runs without autofocus — it degrades, it doesn't break.

3. **Mutable HEAD** — intentional and consistent with `motors` (the package this is modeled on), which also tracks `HEAD` so testers get the latest. I'll pin to a tag once the plugin cuts its first release.

Issue #2319, opened 2026-08-27
Proposal: Make WireGuard startup resilient to temporary DNS unavailability


usa-:
I have identified the cause of an intermittent WireGuard startup failure. As I suspected, the problem occurs when the server address is specified as a hostname rather than an IP address, and DNS is unavailable during reboot, either because the network is not ready yet or because the DNS server itself is unreachable.

The error occurs after approximately 2.5 minutes in `/usr/sbin/wireguard`, at this line:

```sh
wg setconf wg0 /tmp/wireguard.conf || { echo "Error: Failed to apply wireguard configuration." >&2; exit 1; }
```

## How to reproduce

This can be reproduced in real time on a camera connected to a local network:

1. Disconnect the local network from the Internet (more precisely, block communication between the camera and the DNS server).
2. Configure WireGuard to use a hostname as the server address.
3. On the camera, execute:

```sh
chmod -x /etc/init.d/S98wireguard
```

4. Reboot the camera.
5. Run `wireguard` on the camera and observe the output:

```text
root@gk7205v300-imx335:~# wireguard
Try again: `${Endpoint}'. Trying again in 1.00 seconds...
Try again: `${Endpoint}'. Trying again in 1.20 seconds...
Try again: `${Endpoint}'. Trying again in 1.44 seconds...
Try again: `${Endpoint}'. Trying again in 1.73 seconds...
Try again: `${Endpoint}'. Trying again in 2.07 seconds...
Try again: `${Endpoint}'. Trying again in 2.49 seconds...
Try again: `${Endpoint}'. Trying again in 2.99 seconds...
Try again: `${Endpoint}'. Trying again in 3.58 seconds...
Try again: `${Endpoint}'. Trying again in 4.30 seconds...
Try again: `${Endpoint}'. Trying again in 5.16 seconds...
Try again: `${Endpoint}'. Trying again in 6.19 seconds...
Try again: `${Endpoint}'. Trying again in 7.43 seconds...
Try again: `${Endpoint}'. Trying again in 8.92 seconds...
Try again: `${Endpoint}'. Trying again in 10.70 seconds...
Try again: `${Endpoint}'. Trying again in 12.84 seconds...
Try again: `${Endpoint}'
Configuration parsing error
Error: Failed to apply wireguard configuration.
```

6. Reconnect the camera to the Internet and restore the normal startup:

```sh
chmod +x /etc/init.d/S98wireguard
```

## Possible solutions

[...]

PR #2377, opened 2026-09-06
Fix WireGuard startup error handling

usa-:
## Problem

WireGuard can fail to start after a reboot when the peer endpoint is specified as a hostname and DNS is temporarily unavailable. In that case, the camera can end up without a working WireGuard connection.

## Solution

The WireGuard script now waits for DNS to become available when the endpoint is specified as a hostname, while IPv4 and IPv6 literal endpoints skip the DNS wait. An empty `wg_endpoint` fails immediately with an error.

WireGuard is started asynchronously so that a DNS delay does not block the boot process. `rc.local` and other boot-time scripts should not assume that `wg0` is immediately available.

## Hardware tested on

- gk7205v300, G6S
- t31x, don't know the board name
- ssc337de, don't know the board name
- ssc378de, don't know the board name, multiple cameras

The fix has been tested for more than a week on these platforms.

## Evidence

See [OpenIPC/firmware#2319](https://github.com/OpenIPC/firmware/issues/2319) for the problem reproduction, testing results, and discussion.

## Scope


general/overlay/usr/sbin/wireguard
```diff
@@ -5,12 +5,21 @@ ip link add dev wg0 type wireguard || { echo "Error: Failed to create wg0 interf
...
 WG_PRIVKEY="$(fw_printenv -n wg_privkey)"
 WG_PRESHARED_KEY="$(fw_printenv -n wg_sharkey)"
+ENDPOINT=$(fw_printenv -n wg_endpoint)
+HOST="${ENDPOINT%:*}"
+
+if ! echo "$HOST" | grep -qE '^((^|\.)(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])){4}$'; then
+    while ! nslookup "$HOST" >/dev/null 2>&1; do
+        sleep 10
+    done
+fi
```
openipc-ai: **A bracketed IPv6 endpoint never starts.** `[2001:db8::1]:51820` gives `HOST='[2001:db8::1]'`, which fails the IPv4 test, and then `nslookup '[2001:db8::1]'` returns 1 forever, even with DNS up (measured on busybox 1.36.1). Today's script handles that endpoint, since `wg` parses the brackets itself. Skipping the wait when `HOST` starts with `[` is enough.

Two things folded into the suggestion: a why-comment citing #2319 (comments in this tree say why, and this one will not be obvious a year from now), and one `logger` line, so that from outside "DNS is down" and "WireGuard is broken" no longer look the same.

```suggestion
# A hostname endpoint makes wg setconf retry for ~2.5 min and then fail for good
# when DNS is not reachable yet at boot (#2319), so wait for the name here, out of
# the boot path. Literals skip the wait: nslookup on an IPv4 is a reverse lookup
# that needs DNS too, and wg parses a bracketed IPv6 itself.
case "$HOST" in
    \[*) ;;
    *)
        if ! echo "$HOST" | grep -qE '^((^|\.)(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])){4}$'; then
            logger -t wireguard "waiting for $HOST to resolve"
            while ! nslookup "$HOST" >/dev/null 2>&1; do
                sleep 10
            done
        fi
        ;;
esac
```

Minor, noting only: a name that resolves only through `/etc/hosts` never satisfies busybox's nslookup (it queries the server directly), though `wg` would resolve it. Rare on a camera.

general/overlay/usr/sbin/wireguard
```diff
@@ -5,12 +5,21 @@ ip link add dev wg0 type wireguard || { echo "Error: Failed to create wg0 interf
 
 WG_PRIVKEY="$(fw_printenv -n wg_privkey)"
 WG_PRESHARED_KEY="$(fw_printenv -n wg_sharkey)"
+ENDPOINT=$(fw_printenv -n wg_endpoint)
```
openipc-ai, line 8: An empty `wg_endpoint` used to fail at once on the console; now it sits in the loop until a DNS server answers (busybox `nslookup ''` returns 0 then) and fails in the background. Keeping the fast failure is one line:

```suggestion
ENDPOINT=$(fw_printenv -n wg_endpoint)
[ -n "$ENDPOINT" ] || { echo "Error: wg_endpoint environment variable is not set or empty." >&2; exit 1; }
```

general/overlay/etc/init.d/S98wireguard
```diff
@@ -4,7 +4,7 @@ case "$1" in
 	start)
 		wgc=$(fw_printenv -n wg_privkey)
 		if [ -n "$wgc" ]; then
-			wireguard
+			wireguard &
```
openipc-ai, line 7: Worth a sentence in the description: `wg0` now comes up asynchronously, so `rc.local` and anything else that runs at boot can no longer assume it exists. That is the point of the change, but anyone with a custom `rc.local` that touches `wg0` will notice.

openipc-ai (changes requested):
Thanks — the shape is right. A hostname endpoint costs `wg setconf` about 2.5 minutes of retries at boot and then fails for good, so waiting for the name outside the boot path is the correct fix, and the literal-IPv4 bypass is necessary: busybox `nslookup 1.2.3.4` is a reverse lookup and returns 1 even with DNS working. I checked the regex and the `nslookup` exit codes under the same busybox 1.36.1/musl the image ships. The shell parse and strip tests pass on this head.

One thing to fix before merge (inline): a bracketed IPv6 endpoint never starts. The rest is take-or-leave.

Two points that are not code:

- Title: the tree uses `area: lowercase imperative summary`, and maintainers squash with the PR title. `wireguard: wait for the endpoint to resolve instead of failing at boot` says what changed; "error handling" does not.
- Evidence: the before-log in #2319 is good. A pasted `wg show wg0` from a camera that booted with DNS unreachable and then recovered would close the loop.

On Qodo's review: the digit-prefixed-hostname half of its first finding was fixed by your second commit (the regex sends `3cam.example.com` to the resolver correctly). A DNS relapse between the probe and `wg setconf` is covered by wg's own 15 retries. `stop` doing nothing, and a second `start` failing at `ip link add`, were both true before this change.

For the record, `WG_ENDPOINT_RESOLUTION_RETRIES=infinity` on `wg setconf` (the shipped wireguard-tools supports it) was considered: it gives up immediately on NXDOMAIN and REFUSED, which a router's resolver can return while its WAN is down, so your loop is the more robust shape.

CI has not run yet because the fork's workflow runs are waiting for maintainer approval.

openipc-ai:
Thanks for applying the fast-fail on an empty `wg_endpoint`.

Two items from the review are still open on 0622cd35:

1. **IPv6 endpoints** ([thread](https://github.com/OpenIPC/firmware/pull/2377#discussion_r3946594041)) — this is the one blocking approval. A bracketed endpoint such as `[2001:db8::1]:51820` fails the IPv4 test, and then `nslookup '[2001:db8::1]'` returns 1 forever, so a tunnel that comes up today never comes up with this change. Skipping the wait when `$HOST` starts with `[` is enough; the suggestion in that thread does it and can be applied from the web UI like the last one.

2. **PR description** ([thread](https://github.com/OpenIPC/firmware/pull/2377#discussion_r3946594060)) — a sentence noting that `wg0` now comes up asynchronously, so `rc.local` and anything else that runs at boot can no longer assume it exists.

Once the first lands I will re-review.

general/overlay/etc/init.d/S98wireguard
```diff
@@ -1,18 +1,21 @@
...
 		;;
 
 	stop)
+		start-stop-daemon -K -q -p "$PIDFILE"
+		rm -f "$PIDFILE"
 		;;
 
 	*)
-		echo "Usage: $0 {start}"
+		echo "Usage: $0 {start|stop}"
```
openipc-ai: `/usr/sbin/wireguard` is a one-shot configurator, not a daemon: on the normal path it exits a few seconds after boot, as soon as `wg setconf` has run. That makes this `start-stop-daemon` form unsafe, and I would ask you to go back to `wireguard &` with the empty `stop`. Measured with busybox 1.36.1, the version in this tree:

1. **`stop` can kill a bystander.** The pidfile is stale for the camera's whole uptime, and `rcK` runs `S98wireguard stop` at every reboot. `-K -p` signals whatever pid the file holds: with the file pointing at an unrelated process, `start-stop-daemon -K -q -p /var/run/wireguard.pid` returned 0 and killed it. After months of uptime pids have wrapped, so this is a real path. It is the defect #2326 removed from nine init scripts.
2. **`-x /usr/sbin/wireguard` cannot guard a shell script.** `/proc/PID/exe` is busybox and argv[0] is `/bin/sh`, so `-x` never matches: a second `start` spawned a duplicate (rc=0, two instances), and `-K -p … -x /usr/sbin/wireguard` killed nothing (rc=1). That is why `S02fakehwclock`, the tree's only other shell-script service, does not use `start-stop-daemon` at all; the comment in that file explains it.
3. **`-b` sends the script's stdout and stderr to `/dev/null`**, so every `Error: …` line this PR's first commit added no longer reaches the console.

The first review treated the missing `stop` as pre-existing and not something this PR needs to solve. If a cancel for a pending DNS wait is wanted later, the `S02fakehwclock` pattern (write `$!` to the pidfile, re-check `/proc/PID/cmdline` before `kill`) is the safe shape, and it would also need `ip link del wg0`, since a `stop` during the wait leaves `wg0` created but unconfigured and the next `start` then fails at `ip link add`. That is a separate change.

```suggestion

case "$1" in
	start)
		wgc=$(fw_printenv -n wg_privkey)
		if [ -n "$wgc" ]; then
			wireguard &
		fi
		;;

	stop)
		;;

	*)
		echo "Usage: $0 {start}"
```

openipc-ai (changes requested):
Thanks, the IPv6 case is fixed. Measured with the new pattern under musl `grep -E`: `[2001:db8::1]` and `[::1]` skip the wait, `1.2.3.4` skips, `vpn.example.com` and `1abc.example.com` wait. Both scripts still parse under busybox `sh -n`, before and after comment stripping.

The `start-stop-daemon` change in 0c8e7fed is the new blocker; details and a suggestion to revert it are inline on `S98wireguard`.

Still open from the first review: a sentence in the description that `wg0` now comes up asynchronously, so `rc.local` and anything else at boot can no longer assume it exists.

CI on this fork is held for maintainer approval; it will be run once the init script is back to its previous form.

openipc-ai (approved):
Everything from both reviews is in: the IPv6 guard, the empty-endpoint fast-fail, `S98wireguard` back to `wireguard &` with the empty `stop`, and the description now says `wg0` comes up asynchronously. Both scripts parse under the busybox `sh` this tree builds, before and after comment stripping.

I brought the branch up to date with master (the protected branch requires it) and released the fork's CI runs. This merges once the board matrix is green.

usa-:
I don't think `wg setconf` should be retried in a loop. It can fail for reasons unrelated to DNS, and retrying it indefinitely would not fix those errors. The DNS wait is only meant to handle the case where the endpoint hostname is not resolvable yet.

PR #2380, opened 2026-09-07
rv1106: enable CONFIG_MTD_UBI_BLOCK, so the UBI squashfs rootfs mounts

widgetii:
## Problem

The rv1106 NAND image boots a read-only squashfs root from the UBI volume
`rootfs` via `/dev/ubiblock0_1`, with `rootfs_data` as a writable ubifs overlay
(`general/scripts/ubifs/ubinize_rockchip.cfg`; the zboot bootargs are
`root=/dev/ubiblock0_1 rootfstype=squashfs ... ubifs`). But the rv1106 kernel
config shipped `# CONFIG_MTD_UBI_BLOCK is not set`, so `/dev/ubiblock0_1` never
appears, root cannot mount, and the image does not boot.

The writable overlay this unlocks is also what lets per-device state persist:
dropbear host keys, the claim state in `/etc/shadow`, and the fw_env config the
patched u-boot-tools autodetects — without it, none of that survives a reboot
and the wired NIC falls back to a bogus fixed MAC.

## Hardware tested on

rv1106 (RV1106G3), Luckfox Pico Max, 256 MB SPI-NAND, sc3336 sensor.

## Evidence

Before (`# CONFIG_MTD_UBI_BLOCK is not set`): no `/dev/ubiblock0_1`, so
`root=/dev/ubiblock0_1 rootfstype=squashfs` has no device to mount.

After (`CONFIG_MTD_UBI_BLOCK=y`), serial boot on the Luckfox Pico Max:

```
block ubiblock0_1: created from ubi0:1(rootfs)
VFS: Mounted root (squashfs filesystem) readonly on device 254:0.
UBIFS (ubi0:2): UBIFS: mounted UBI device 0, volume 2, name "rootfs_data"
...
Welcome to OpenIPC
openipc-rv1106 login:
```

The image now boots to userspace; the writable `rootfs_data` overlay mounts, so
the dropbear host key, the claim, and the autodetected fw_env config persist
across reboots.

Local checks (no camera needed):

```
ci-matrix.py --self-test         PASS
ci-matrix.py --stdin             -> rv1103_lite, rv1106_lite
test_kconfig_graph.py            PASS
test_shell_parse.sh (STRICT=1)   PASS
built kernel .config             CONFIG_MTD_UBI_BLOCK=y
```

## Scope


widgetii:
Superseding with a single rv1106 PR that carries all the board's firmware changes together, to avoid multiple build-matrix runs.

PR #2393, opened 2026-09-11
gpio-motors: spin only while the rounding error is material

phedoreanu:
Follow-up to #2247, picking up both points from the merge comment there.

**Gate the spin on the error being material.** `delay_us()` spun for any sub-tick delay, so its CPU cost rose exactly as its benefit fell: at delay 9 on an HZ=100 kernel that is ~14s of pinned core per 200-step move to shave an 11% timing error, with every `clock_gettime` in the spin a real syscall on gk7205v500 (`CONFIG_ARM_ARCH_TIMER_VCT_ACCESS` is not set, so no vDSO fast path). The sleep threshold is now a quarter tick instead of a whole one:

- below a quarter tick (delay 1-2 at HZ=100), sleeping would at least quadruple the step period, so the spin stays — this is the range the spin was added for;
- from a quarter tick up (delay 3+), `usleep()` and accept the rounding.

Against the table in the merge comment, that keeps the 900%/400% rows spinning and puts the 100% and 11% rows to sleep, including the 14.4s worst case.

**What the threshold costs.** The merge-comment table listed 1, 2, 5 and 9 ms; the two rows it skipped are the ones this change gives back to the tick, so here they are on the record (200 steps = 1600 micro-steps, HZ=100):

| delay | before (#2247) | after (this PR) |
| --- | --- | --- |
| 3 ms | spins, 4.8 s move, core pinned | sleeps at 10 ms, 16 s move |
| 4 ms | spins, 6.4 s move, core pinned | sleeps at 10 ms, 16 s move |

Delay 4 is the number #2247 opened with ("still 18s at delay 4" on the Hi3518EV200), so this hands that measurement back in exchange for not pinning the core at 4.8-6.4 s per move. Anyone who wants the faster move at 3-4 ms can ask for 2 and get it at 3.2 s with the spin; that is the trade the threshold makes.

**Check `clock_gettime()` inside the spin loop.** An unchecked failure would leave `now` stale and the loop would never terminate. On failure it now falls back to `usleep()` for the full delay, same as the existing pre-loop failure path — over-waiting is the safe direction for a stepper.

Compiles clean with `-Wall -Wextra`.

**Measured on a Hi3518EV200** (same kernel class as gk7205v200/v500: HZ=100, no high-resolution timers, no preempt; `clock_getres` = 10 ms), with majestic streaming: `usleep()` lands on 10 ms at every delay; the merged code spins on every row at 72-76% CPU, with the 9 ms row burning 11.8 s of CPU to finish 0.7 s sooner than sleeping; this PR keeps the 1 and 2 ms rows identical to the merged code and sends 3, 4 and 9 ms to the tick at 1.7% CPU. Full table, raw output and harness source: https://github.com/OpenIPC/firmware/pull/2393#issuecomment-5635746495

The motor half of the evidence needs a gk7205v500/v510 owner: `BR2_PACKAGE_GPIO_MOTORS=y` only exists in those defconfigs, and nobody in the #2247 thread has one. Whether the new pacing at 3-4 ms slips or stalls a coil is the part the primitive test cannot answer.


general/package/gpio-motors/src/gpio-motors.c
```diff
@@ -166,19 +166,23 @@ void gpio_config() {
...
+ * CLOCK_MONOTONIC only below a quarter tick, where sleeping would at least
+ * quadruple the step period, and sleep everywhere else.
  */
 void delay_us(long us) {
 	if (us <= 0) {
 		return;
 	}
 
-	if (CLOCK_RES_NS <= 1000000 || us >= CLOCK_RES_NS / 1000) {
+	if (CLOCK_RES_NS <= 1000000 || us >= CLOCK_RES_NS / 4000) {
```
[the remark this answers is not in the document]
phedoreanu: Acknowledged, and the PR description says the same: I don't have a gk7205v500-family board, so this is not measured there. The threshold isn't a guess though — it is read straight off the measured table in the #2247 merge comment (per-delay spin cost vs. rounding error, taken on the affected kernel class), and the gating logic is identical to what was already merged, just with a smaller cutoff. If a maintainer with a V500/V510 board wants to re-measure before merging, the PR body names that family as the target for exactly that reason.

openipc-ai, line 185: The table I wrote in the merge comment listed 1, 2, 5 and 9 ms. It skipped 3 and 4 - and those are the rows this line costs:

| delay | before | after |
| --- | --- | --- |
| 3 ms | spins, 4.8 s move | sleeps, 16 s move |
| 4 ms | spins, 6.4 s move | sleeps, 16 s move |

(200 steps = 1600 micro-steps.) Delay 4 is the number #2247 opened with - "still 18s at delay 4" on your Hi3518EV200 - so this hands back the measurement that motivated the original PR.

I still want the threshold where you put it; that is the trade I asked for. But put those two rows in the PR body so it is on the record, rather than implied by the four rows that happened to be in my table.

general/package/gpio-motors/src/gpio-motors.c
```diff
@@ -193,7 +197,10 @@ void delay_us(long us) {
 	 * time, which a preemption in the middle of the spin can reach */
 	long long target = (long long)us * 1000;
 	for (;;) {
-		clock_gettime(CLOCK_MONOTONIC, &now);
+		if (clock_gettime(CLOCK_MONOTONIC, &now) != 0) {
```
openipc-ai, line 200: Fine as a fallback. Worth noting in passing that it sleeps the full `us` on top of the time already spun, so on a coarse kernel the recovery costs a whole tick - over-waiting, which is the right direction for a stepper.

openipc-ai (changes requested):
Both items from the merge comment are here, and the arithmetic checks out: `CLOCK_RES_NS / 4000` is 2500 us at HZ=100, so the spin survives exactly where I asked it to and the 14.4 s tail is gone.

Two notes inline: the new threshold quietly costs the two rows my table skipped, and the `clock_gettime` fallback over-waits by a tick.

**Qodo's rule violation is correct, and it is still open** - you answered it at 09:18, but answering is not the same as clearing it. This changes motor pacing in three shipped images with no before/after from a board that runs the code. `best_practices.md` 5.3 says flag and close, and I am not going to pretend otherwise because I asked for the change.

That said, I created this situation - my own scope note on #2247 conceded nobody in that thread has a gk7205v500-family board, and the lab here does not have one either. So let me split the ask rather than block on the impossible half:

- **The delay primitive is measurable without a PTZ board.** I measured `usleep()` quantisation on the lab gk7205v200 in #2247 - same HZ=100, same `# CONFIG_HIGH_RES_TIMERS is not set`, same `# CONFIG_ARM_ARCH_TIMER_VCT_ACCESS is not set`. A short harness calling the new `delay_us()` at 1, 2, 3, 4 and 9 ms, reporting wall-clock and CPU time per call, would settle the threshold behaviour on the same hardware class. That is a real before/after, not a paste from a board that was not exercising the change.
- **Motor behaviour needs a v500/v510 owner.** Whether the new pacing at delay 3-4 slips or stalls a coil is not something the primitive test can answer.

Your point that the threshold is read off the measured table rather than guessed is fair, and I am not asking you to re-derive it - only to show the primitive behaving as the table predicts on a board of that kernel class.

Please mark this draft until the first half is attached. Separately: CI has not run on this branch at all - no checks reported - and it is behind master. Rebase and let the matrix go green before this is mergeable either way.

general/package/gpio-motors/src/gpio-motors.c
phedoreanu: Both rows are in the PR body now, under "What the threshold costs", with the 4.8 s / 6.4 s before and 16 s after, and a note that delay 4 is the number #2247 opened with. Threshold left where it is.

general/package/gpio-motors/src/gpio-motors.c
phedoreanu: Agreed — the recovery over-waits by up to a tick on a coarse kernel. Sleeping only the remainder would still round up to the same tick there, so the full `us` costs nothing extra on the boards this matters for, and it keeps the fallback identical to the pre-loop one.

phedoreanu:
Measured, as asked — the primitive half, on a board of the same kernel class.

**Board:** Hi3518EV200 (ARM926EJ-S), OpenIPC 4.9.37, `CONFIG_HZ=100`, `# CONFIG_HIGH_RES_TIMERS is not set`, `# CONFIG_PREEMPT is not set`, `# CONFIG_SMP is not set`. The harness reports `clock_getres(CLOCK_MONOTONIC) = 10000000 ns`, so it takes the coarse path exactly like gk7205v200/v500. majestic was streaming throughout (load 0.6-1.4), so this is a loaded core, not an idle one.

**Harness:** one static binary with the two `delay_us()` bodies copied verbatim — `before` is #2247 as merged (sleep at or above one tick), `after` is this PR (sleep at or above a quarter tick) — plus plain `usleep()` as the baseline. 1600 calls per row, i.e. one 200-step move. Wall time from `CLOCK_MONOTONIC`, CPU from `getrusage()` (user + sys, so the sampling is tick-granular: read the CPU column as ±10 ms).

| delay | usleep | before (#2247) | after (this PR) |
| --- | --- | --- | --- |
| 1 ms | 9.996 ms/call, 1.7% CPU | 1.294 ms/call, 76% CPU | 1.300 ms/call, 76% CPU |
| 2 ms | 9.999 ms/call, 1.7% CPU | 2.504 ms/call, 76% CPU | 2.689 ms/call, 70% CPU |
| 3 ms | 10.013 ms/call, 1.7% CPU | 3.836 ms/call, 72% CPU | 10.013 ms/call, 1.7% CPU |
| 4 ms | 10.006 ms/call, 1.7% CPU | 4.915 ms/call, 73% CPU | 10.010 ms/call, 1.7% CPU |
| 9 ms | 10.400 ms/call, 1.6% CPU | 9.980 ms/call, 74% CPU | 10.426 ms/call, 1.7% CPU |

Reading it:

- `usleep()` lands on 10 ms at every delay — the tick floor from #2247, reproduced on the same class.
- `before` spins on every row. The 9 ms row is the tail from your table: 15.97 s wall against 16.64 s for sleeping, so 0.7 s saved for 11.8 s of CPU.
- `after` is identical to `before` at 1 and 2 ms (within noise), and sends 3, 4 and 9 ms to the tick at 1.7% CPU. The 3 and 4 ms rows are the cost the PR body now lists — 6.1 s and 7.9 s moves become 16 s.
- The spin rows overrun the requested delay (1.29 ms for a 1 ms ask, 4.9 for 4) because the encoder takes its tick slices; a spinning process is still preempted. That is the contrast with the kernel-side busy-wait in #2394, which is why that one carries a `cond_resched()`.

<details>
<summary>Raw output</summary>

```
clock_getres(CLOCK_MONOTONIC) = 10000000 ns; 1600 calls per row (200 steps x 8 micro-steps)
usleep   1 ms x1600  wall  15993.8 ms ( 9.996 ms/call)  cpu    270.0 ms (  1.7%)
before   1 ms x1600  wall   2070.9 ms ( 1.294 ms/call)  cpu   1580.0 ms ( 76.3%)
after    1 ms x1600  wall   2080.2 ms ( 1.300 ms/call)  cpu   1590.0 ms ( 76.4%)

usleep   2 ms x1600  wall  15998.9 ms ( 9.999 ms/call)  cpu    270.0 ms (  1.7%)
before   2 ms x1600  wall   4006.6 ms ( 2.504 ms/call)  cpu   3030.0 ms ( 75.6%)
after    2 ms x1600  wall   4302.0 ms ( 2.689 ms/call)  cpu   3010.0 ms ( 70.0%)

usleep   3 ms x1600  wall  16021.2 ms (10.013 ms/call)  cpu    270.0 ms (  1.7%)
before   3 ms x1600  wall   6137.7 ms ( 3.836 ms/call)  cpu   4390.0 ms ( 71.5%)
after    3 ms x1600  wall  16021.5 ms (10.013 ms/call)  cpu    270.0 ms (  1.7%)

usleep   4 ms x1600  wall  16010.1 ms (10.006 ms/call)  cpu    270.0 ms (  1.7%)
before   4 ms x1600  wall   7864.1 ms ( 4.915 ms/call)  cpu   5710.0 ms ( 72.6%)
after    4 ms x1600  wall  16015.9 ms (10.010 ms/call)  cpu    270.0 ms (  1.7%)

usleep   9 ms x1600  wall  16640.1 ms (10.400 ms/call)  cpu    270.0 ms (  1.6%)
before   9 ms x1600  wall  15967.8 ms ( 9.980 ms/call)  cpu  11800.0 ms ( 73.9%)
after    9 ms x1600  wall  16682.2 ms (10.426 ms/call)  cpu    280.0 ms (  1.7%)
```
</details>

<details>
<summary>Harness source (delay-bench.c, built with arm-openipc-linux-musleabi-gcc -O2 -static)</summary>

```c
/*
 * delay-bench: wall-clock and CPU cost of the gpio-motors delay_us()
 * primitive, before (#2247: spin below one tick) and after (#2393: spin
 * below a quarter tick), plus plain usleep() as the baseline. Both
 * delay_us() bodies are copied verbatim from the respective commits.
 */
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>
#include <unistd.h>
#include <sys/resource.h>

long CLOCK_RES_NS = 0;

static void spin_to(struct timespec start, long us) {
	struct timespec now;
	long long target = (long long)us * 1000;
	for (;;) {
		if (clock_gettime(CLOCK_MONOTONIC, &now) != 0) {
			usleep(us);
			return;
		}
		long long elapsed = (long long)(now.tv_sec - start.tv_sec) * 1000000000LL + (now.tv_nsec - start.tv_nsec);
		if (elapsed >= target)
			return;
	}
}

/* #2247 as merged: sleep at or above one tick, spin below */
static void delay_before(long us) {
	if (us <= 0) return;
	if (CLOCK_RES_NS <= 1000000 || us >= CLOCK_RES_NS / 1000) { usleep(us); return; }
	struct timespec start;
	if (clock_gettime(CLOCK_MONOTONIC, &start) != 0) { usleep(us); return; }
	spin_to(start, us);
}

/* #2393: sleep at or above a quarter tick, spin below */
static void delay_after(long us) {
	if (us <= 0) return;
	if (CLOCK_RES_NS <= 1000000 || us >= CLOCK_RES_NS / 4000) { usleep(us); return; }
	struct timespec start;
	if (clock_gettime(CLOCK_MONOTONIC, &start) != 0) { usleep(us); return; }
	spin_to(start, us);
}

static void delay_usleep(long us) { usleep(us); }

static double ts_diff(struct timespec a, struct timespec b) {
	return (b.tv_sec - a.tv_sec) * 1e3 + (b.tv_nsec - a.tv_nsec) / 1e6;
}

static double ru_ms(struct rusage r) {
	return (r.ru_utime.tv_sec + r.ru_stime.tv_sec) * 1e3 + (r.ru_utime.tv_usec + r.ru_stime.tv_usec) / 1e3;
}

static void bench(const char *name, void (*fn)(long), long us, int calls) {
	struct timespec t0, t1;
	struct rusage r0, r1;
	getrusage(RUSAGE_SELF, &r0);
	clock_gettime(CLOCK_MONOTONIC, &t0);
	for (int i = 0; i < calls; i++)
		fn(us);
	clock_gettime(CLOCK_MONOTONIC, &t1);
	getrusage(RUSAGE_SELF, &r1);
	double wall = ts_diff(t0, t1), cpu = ru_ms(r1) - ru_ms(r0);
	printf("%-7s %2ld ms x%4d  wall %8.1f ms (%6.3f ms/call)  cpu %8.1f ms (%5.1f%%)\n",
	       name, us / 1000, calls, wall, wall / calls, cpu, 100.0 * cpu / wall);
}

int main(int argc, char *argv[]) {
	int calls = argc > 1 ? atoi(argv[1]) : 1600;
	struct timespec res;
	if (clock_getres(CLOCK_MONOTONIC, &res) == 0)
		CLOCK_RES_NS = res.tv_sec ? 1000000000L : res.tv_nsec;
	printf("clock_getres(CLOCK_MONOTONIC) = %ld ns; %d calls per row (200 steps x 8 micro-steps)\n", CLOCK_RES_NS, calls);
	long delays[] = {1, 2, 3, 4, 9};
	for (unsigned i = 0; i < sizeof delays / sizeof *delays; i++) {
		long us = delays[i] * 1000;
		bench("usleep", delay_usleep, us, calls);
		bench("before", delay_before, us, calls);
		bench("after", delay_after, us, calls);
		printf("\n");
	}
	return 0;
}
```
</details>

The motor half stays open for a v500/v510 owner, as agreed. Marking ready for review with the primitive measurement attached; branch is rebased onto current master.

On CI: the runs for this head exist — build, lint, shell-tests, gcc-compat and qodo-gate all fired on the rebase — but every one is sitting at `action_required`, which is the fork-PR approval gate. They need a maintainer's "Approve and run" before anything reports; I can't clear that from my side.

general/package/gpio-motors/src/gpio-motors.c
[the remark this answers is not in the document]
phedoreanu: Evidence attached now: https://github.com/OpenIPC/firmware/pull/2393#issuecomment-5635746495 — the primitive measured before/after on a Hi3518EV200 (same HZ=100, no-hrtimer, no-preempt class), harness source included. The motor half remains for a gk7205v500/v510 owner, per the split in the review.

openipc-ai (approved):
Both items from the merge comment are in, the cost rows are on the record in the body, and the primitive half of the evidence is measured.

The Hi3518EV200 numbers are exactly the split I asked for: `usleep()` pinned at ~10 ms across every delay, `before` spinning on every row (9 ms costing 11.8 s of CPU to save 0.7 s of wall), `after` matching `before` at 1-2 ms within noise and dropping 3/4/9 ms to the tick at 1.7% CPU. Same kernel class as the boards that ship the package — HZ=100, no hrtimers, no preempt, no SMP — and measured on a loaded core with majestic streaming, which is the honest condition.

Your point that a spinning process is still preempted (1.29 ms for a 1 ms ask) is a useful thing to have on the record, and it is the right contrast with the kernel-side path in #2394.

Approving. The motor-behaviour half stays open for a gk7205v500/v510 owner, as agreed - not a condition on this.

PR #2395, opened 2026-09-11
fix: msposd OSD lost on Star6E boot (missing -d, waybeam respawn race)

moro161:
## Summary

Two related bugs found tracking down VTX-side OSD loss on a
`ssc338q_fpv_openipc-urllc-aio-nor` (RunCam WiFiLink) nightly build,
both in `start_telemetry()`:

- **OSD never appears at all:** the `msposd` invocation is missing
  `-d`/`--osd`. Without it, `msposd` runs and correctly parses FC
  telemetry (confirmed with `-v`: detects the FC fine) but never calls
  `MI_RGN_Init` / draws anything.

- **OSD appears at boot, then vanishes for good after a few seconds**
  (Star6E SoCs, e.g. RunCam WiFiLink): `waybeam` does an internal
  cold-restart shortly after its own startup (fork+exec, to lock in a
  cold-boot FPS re-kick — see `src/star6e_pipeline.c` in
  [OpenIPC/waybeam](https://github.com/OpenIPC/waybeam)), which
  destroys and recreates its venc channel. `msposd` binds its OSD
  region to whatever venc channel exists at its own startup, once, via
  `MI_RGN_Init`, and never reattaches. This script always starts
  `msposd` right after `S95waybeam` returns, which is *before*
  waybeam's internal respawn fires — so the OSD region is always
  orphaned when the channel gets torn down, even with `-d` fixed.

## Fix

- Add `-d` to the `msposd` invocation.
- Add a background watcher (`watch_waybeam_respawn`) that detects the
  waybeam respawn via `/tmp/waybeam.log` content (re-scanning the whole
  file each check, since the respawn can complete before the watcher
  starts — not a fixed delay) and, once the respawned channel is
  actually up (a second `"ISP channel ready"` line, not a guessed
  sleep), restarts `msposd` so it reattaches. Bounded loops are a
  safety net, not the wait mechanism.
- No-op on SoCs/builds where `waybeam` doesn't do this respawn, or
  `majestic` is in use instead — the log marker never appears and the
  watcher just exits after its timeout.

## Test plan

- [x] Verified root cause with `msposd -v` in foreground: no
      `MI_RGN_Init`/region activity without `-d`; with `-d`, region
      created successfully and live telemetry flows.
- [x] Verified the channel-teardown race directly via `logread`:
      `client [X] disconnected, module:venc` → `MI_VENC_IMPL_DestroyChn`
      → `client [Y] connected, module:venc`, correlated with
      `waybeam`'s own `"Reinit requested: cold restart via fork+exec"` /
      `"Respawning:"` log lines.
- [x] Verified the fix end-to-end across three clean on-device reboots
      with no manual intervention: watcher log line
      (`"waybeam respawned post-boot, restarting msposd to reattach
      OSD"`) appears, `grep -c "ISP channel ready" /tmp/waybeam.log`
      reads 2 at that point (real signal, not a timing guess), `msposd`
      comes back up with `-d`, stable PID over 30+s of polling
      afterward on each reboot.
- [ ] Not tested on non-Star6E SoCs / majestic-based images (no such
      hardware on hand) — the watcher is designed to be a no-op there,
      but that's unverified.

### DIFF
--- a/general/package/wifibroadcast-ng/files/wifibroadcast
+++ b/general/package/wifibroadcast-ng/files/wifibroadcast
@@ -129,11 +129,63 @@ start_telemetry() {
 			size=$(curl -s localhost/api/v1/config.json | jsonfilter -qe "@.video0.size")
 		fi
 		msposd -b 115200 -c 8 -r "$osd_fps" -m /dev/"$serial" \
-			-o 127.0.0.1:"$port_tx" -z "$size" > /dev/null &
+			-o 127.0.0.1:"$port_tx" -z "$size" -d > /dev/null &
+
+		[ "$router" = "msposd" ] && watch_waybeam_respawn "$osd_fps" "$serial" "$port_tx" "$size"
 
 		sleep 5
 		echo "&L70&F28&G8CPU:&C TEMP:&T\n&B" >/tmp/MSPOSD.msg
 	fi
+}
+
+# msposd needs -d/--osd to actually draw the OSD overlay (MI_RGN_Init) —
+# without it, it runs and parses FC telemetry fine but never renders
+# anything. See github.com/OpenIPC/msposd for the flag.
+#
+# waybeam (the venc/encoder daemon on Star6E SoCs) does an internal
+# cold-restart shortly after its own startup (fork+exec, to lock in a
+# cold-boot FPS re-kick — see src/star6e_pipeline.c in OpenIPC/waybeam),
+# which destroys and recreates its venc channel. msposd binds its OSD
+# region to whatever venc channel exists at its own startup, once, via
+# MI_RGN_Init, and never reattaches — so if msposd starts before this
+# happens (which it always does: this script starts msposd right after
+# S95waybeam returns, which is before waybeam's internal respawn fires),
+# the OSD region is silently destroyed along with the old channel and
+# never comes back. Re-scan /tmp/waybeam.log (whole file each check,
+# since the respawn can complete before we start watching) for the
+# respawn, then wait for its "ISP channel ready" to appear a second time
+# (the real signal the new channel is up, not a guessed delay — it also
+# appears once during the normal pre-respawn boot), then restart msposd
+# so it reattaches. The bounded loops below are a safety net, not the
+# wait mechanism. No-op on SoCs/builds where waybeam doesn't do this
+# respawn (or majestic is in use instead) — /tmp/waybeam.log never gets
+# the marker and this just exits after the timeout.
+watch_waybeam_respawn() {
+	osd_fps="$1"; serial="$2"; port_tx="$3"; size="$4"
+	(
+		elapsed=0
+		while [ "$elapsed" -lt 30 ]; do
+			grep -q "Respawning:" /tmp/waybeam.log 2>/dev/null && break
+			sleep 0.5
+			elapsed=$((elapsed + 1))
+		done
+		grep -q "Respawning:" /tmp/waybeam.log 2>/dev/null || exit 0
+
+		elapsed=0
+		while [ "$elapsed" -lt 20 ]; do
+			[ "$(grep -c "ISP channel ready" /tmp/waybeam.log 2>/dev/null)" -ge 2 ] && break
+			sleep 0.3
+			elapsed=$((elapsed + 1))
+		done
+
+		echo_log "waybeam respawned post-boot, restarting msposd to reattach OSD"
+		killall -q msposd
+		while pidof msposd >/dev/null 2>&1; do
+			sleep 0.2
+		done
+		msposd -b 115200 -c 8 -r "$osd_fps" -m /dev/"$serial" \
+			-o 127.0.0.1:"$port_tx" -z "$size" -d > /dev/null &
+	) &
 }
 
 video_settings() {


general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
 			size=$(curl -s localhost/api/v1/config.json | jsonfilter -qe "@.video0.size")
 		fi
 		msposd -b 115200 -c 8 -r "$osd_fps" -m /dev/"$serial" \
-			-o 127.0.0.1:"$port_tx" -z "$size" > /dev/null &
+			-o 127.0.0.1:"$port_tx" -z "$size" -d > /dev/null &
```
openipc-ai, line 132: `-d` is correct and I want to keep it — but it is not the whole story, and the PR description has the cause slightly wrong.

`DrawOSD` starts `false` in msposd and is set from two places in the option loop: `case 'd'`, and `case 'z'` — but `case 'z'` sets it **only inside `if (limit)`**, where `limit = strchr(buffer, 'x')`. So `-z 1920x1080` already implies `-d`. The only way the existing line draws nothing is `$size` arriving empty or without an `x`, which happens when majestic's HTTP server is not yet listening at S98 time, or `.video0.size` is unset in `majestic.yaml`.

That matters because `-d` alone masks it rather than fixing it: with an empty `-z`, `set_resolution()` is never called and msposd falls back to its built-in default, clamped to 1280x720. The overlay then comes up at the wrong geometry instead of not at all — which is harder to notice than the current failure.

Please guard the value as well, something like:

```sh
size=$(curl -s localhost/api/v1/config.json | jsonfilter -qe "@.video0.size")
case "$size" in
	*x*) ;;
	*) size=1920x1080; echo_log "video0.size unavailable, OSD falling back to $size" ;;
esac
```

Keep `-d` as the belt-and-braces so the overlay no longer depends on that curl succeeding.

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
+
+		[ "$router" = "msposd" ] && watch_waybeam_respawn "$osd_fps" "$serial" "$port_tx" "$size"
```
openipc-ai, line 134: This is the blocking line. `waybeam` is built by no defconfig in this repository or in OpenIPC/builder — its `Config.in` has `depends on !BR2_PACKAGE_MAJESTIC`, and every builder defconfig that selects `BR2_PACKAGE_WIFIBROADCAST_NG` also sets `BR2_PACKAGE_MAJESTIC=y`.

So on every image that ships this script, `/tmp/waybeam.log` does not exist and the watcher is unreachable code that costs ~15 s of background polling after each boot. Details and the full defconfig list are in the review body.

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
 	fi
 }
 
+# msposd needs -d/--osd to actually draw the OSD overlay (MI_RGN_Init) —
+# without it, it runs and parses FC telemetry fine but never renders
+# anything. See github.com/OpenIPC/msposd for the flag.
+#
+# waybeam (the venc/encoder daemon on Star6E SoCs) does an internal
+# cold-restart shortly after its own startup (fork+exec, to lock in a
+# cold-boot FPS re-kick — see src/star6e_pipeline.c in OpenIPC/waybeam),
```
openipc-ai, line 147: This mechanism is not what waybeam does, and the comment will outlive the misunderstanding if it lands.

`star6e_pipeline_cold_boot_fps_rekick()` in `src/star6e_pipeline.c` is a plain `MI_SNR_SetFps` re-issued from the run loop about 1.5 s after start. It does not restart the process, and it does not touch the venc channel.

The fork+exec cold restart lives in `src/venc_respawn.c` and is reached only from `venc_api_request_reinit()`, whose callers are SIGHUP, `/api/v1/restart`, `/api/v1/defaults`, and a RESTART-class `/api/v1/set`. There is no automatic self-restart shortly after boot for this script to race with.

So whatever produced the teardown you captured in `logread` had some other trigger, and it is worth identifying before building a workaround around it — the timing assumption this watcher encodes ("msposd always starts before the respawn fires") may not hold for the real trigger.

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
+# since the respawn can complete before we start watching) for the
+# respawn, then wait for its "ISP channel ready" to appear a second time
+# (the real signal the new channel is up, not a guessed delay — it also
+# appears once during the normal pre-respawn boot), then restart msposd
+# so it reattaches. The bounded loops below are a safety net, not the
+# wait mechanism. No-op on SoCs/builds where waybeam doesn't do this
+# respawn (or majestic is in use instead) — /tmp/waybeam.log never gets
+# the marker and this just exits after the timeout.
+watch_waybeam_respawn() {
+	osd_fps="$1"; serial="$2"; port_tx="$3"; size="$4"
```
openipc-ai, line 164: These assignments are not `local`, so they write the caller's variables. It happens to be harmless here only because `start_telemetry()` uses the same four names with the same values — a rename on either side turns it into a silent bug.

`local` is available in the busybox ash built here (`general/overlay/usr/sbin/sysupgrade` uses it throughout), so:

```sh
watch_waybeam_respawn() {
	local osd_fps="$1" serial="$2" port_tx="$3" size="$4"
```

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
+	(
+		elapsed=0
+		while [ "$elapsed" -lt 30 ]; do
+			grep -q "Respawning:" /tmp/waybeam.log 2>/dev/null && break
+			sleep 0.5
+			elapsed=$((elapsed + 1))
+		done
```
openipc-ai, line 171: `elapsed` counts iterations, not seconds. With `sleep 0.5` the bound is 15 s, not 30; the second loop's `-lt 20` with `sleep 0.3` is 6 s, not 20. The surrounding prose ("just exits after the timeout") reads as though those numbers were seconds.

Either name it `tries`, or keep seconds and derive the bound — whichever, make the comment state the real wall-clock budget, since that is the number a reviewer needs to weigh against how long the encoder actually takes to come back.

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
+
+		elapsed=0
+		while [ "$elapsed" -lt 20 ]; do
+			[ "$(grep -c "ISP channel ready" /tmp/waybeam.log 2>/dev/null)" -ge 2 ] && break
+			sleep 0.3
+			elapsed=$((elapsed + 1))
+		done
+
+		echo_log "waybeam respawned post-boot, restarting msposd to reattach OSD"
+		killall -q msposd
```
openipc-ai, line 182: If the readiness loop times out, execution falls straight through to the restart anyway — there is no success flag and no post-loop guard. So when the replacement channel is slow or never arrives, msposd is killed and restarted against a channel that is not up, re-orphaning the region with no further retry. That is strictly worse than leaving the running instance alone: you lose whatever OSD was there.

Gate the restart on the condition actually being met:

```sh
ready=0
tries=0
while [ "$tries" -lt 20 ]; do
	[ "$(grep -c "ISP channel ready" /tmp/waybeam.log 2>/dev/null)" -ge 2 ] && { ready=1; break; }
	sleep 0.3
	tries=$((tries + 1))
done
[ "$ready" = 1 ] || { echo_log "waybeam respawned but the channel never came up, leaving msposd alone"; exit 0; }
```

(Also raised by the automated review — agreed.)

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
+		while pidof msposd >/dev/null 2>&1; do
+			sleep 0.2
+		done
```
openipc-ai, line 185: This wait is unbounded and never escalates. `killall -q` sends SIGTERM; if msposd is wedged, or if a second watcher starts a fresh msposd while this one is waiting, `pidof` never clears and this subshell spins on a 0.2 s sleep forever — on a flight camera, with no way to notice.

`general/package/waybeam/files/S95waybeam` in this same tree already solves exactly this, and its `stop()` is the pattern to copy: a bounded `wait_exit()`, then `killall -9`, then report the failure rather than pretend the stop happened. Please bound this the same way and give up loudly if msposd survives SIGKILL.

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
+		msposd -b 115200 -c 8 -r "$osd_fps" -m /dev/"$serial" \
+			-o 127.0.0.1:"$port_tx" -z "$size" -d > /dev/null &
```
openipc-ai, line 187: This is a byte-for-byte copy of the invocation at lines 131-132. Two copies of a command line with seven flags will drift — the `-d` this PR adds is itself an illustration of how easy it is for one of them to be wrong for a long time.

Factor it out:

```sh
start_msposd() {
	msposd -b 115200 -c 8 -r "$osd_fps" -m /dev/"$serial" \
		-o 127.0.0.1:"$port_tx" -z "$size" -d > /dev/null &
}
```

and call it from both places.

general/package/wifibroadcast-ng/files/wifibroadcast
```diff
@@ -129,13 +129,65 @@ start_telemetry() {
...
+	) &
```
openipc-ai, line 188: The subshell is backgrounded and untracked, and `stop()` kills only `wfb_rx wfb_tx wfb_tun msposd mavfwd`. So a watcher armed by a previous `start` outlives a stop, and there is no handle to cancel it.

Two reachable consequences:

* `wifibroadcast stop` inside the watcher's ~21 s window kills msposd, then the watcher brings it back up afterwards — the service is nominally stopped with msposd running.
* `wifibroadcast start` run twice leaves two watchers. They race each other's `killall`, and the older one's unbounded `pidof` wait (line 183) can latch onto the msposd the newer one just started and never return.

If the watcher survives review at all it needs to be tracked — write its PID to `/tmp` and have `stop()` kill it, or have it re-check on each iteration that the msposd it was armed for is still the current one.

(Also raised by the automated review — agreed.)

openipc-ai (changes requested):
Thanks for the detailed write-up — the `-d` half is a real bug with a real fix, and the `msposd -v` evidence for it is exactly the right kind. The watcher half has a blocking problem, though: it keys on a package that no board in either repository builds.

## Blocking — `waybeam` is selected by no defconfig, here or in builder

`watch_waybeam_respawn` depends entirely on `/tmp/waybeam.log` existing, which requires `waybeam` to be installed. It never is:

* `general/package/waybeam/Config.in` carries `depends on !BR2_PACKAGE_MAJESTIC` — the two drive the same sensor and encoder, and `sigmastar-osdrv-infinity6e` installs the MI libraries only when Majestic is off.
* Every defconfig in OpenIPC/builder that sets `BR2_PACKAGE_WIFIBROADCAST_NG=y` — all nine of them, `ssc338q_fpv`, `ssc338q_fpv_runcam-wifilink`, `ssc338q_fpv_openipc-urllc-aio`, `ssc338q_fpv_openipc-mario-aio`, `ssc338q_fpv_openipc-thinker-aio`, `ssc338q_fpv_caddx-fly`, `ssc338q_fpv_emax-wyvern-link`, `ssc30kq_fpv`, `ssc377qe_fpv_ccdcam-im50q01-tipoman` — also sets `BR2_PACKAGE_MAJESTIC=y`. So waybeam is excluded by construction on precisely the boards that ship this script.
* `grep -rl BR2_PACKAGE_WAYBEAM=y` across both trees returns nothing.

On every image built from these repositories the streamer is majestic, the marker never appears, and the watcher spends ~15 s after each boot grepping a file that does not exist before giving up. That is `best_practices.md` §6.1 — code no defconfig can reach.

If you were misled by the `NOT_BUILT` comment in `.github/scripts/ci-matrix.py` ("It is selected by the FPV variants in OpenIPC/builder"), that sentence is stale and I will fix it separately.

**This also means the test-plan evidence cannot have come from a stock nightly of `ssc338q_fpv_openipc-urllc-aio-nor`** — `grep -c "ISP channel ready" /tmp/waybeam.log` reading 2 requires a waybeam image, and that board builds majestic. Could you say how the image under test was built? If you are carrying a local waybeam-enabled variant, the watcher belongs wherever that variant lives, not in the shared script.

## The good news

The `-d` bug is real and it is on the configuration everyone actually runs. Split that out with the `$size` guard described inline and it is a small, shippable PR. See the inline comments for why `-z` alone was supposed to cover this and what actually goes wrong.

If the reattach problem is real beyond waybeam — and it plausibly is, since msposd binds its region once and majestic restarts on config changes too — the durable fix is in msposd: reattach on channel loss, rather than have a shell script scrape a log and kill a process. A watcher in `general/package/` reaches every FPV camera and can only ever cover the one restart path it was written for.

## Process notes

* Commit subject is `fix: msposd OSD lost on Star6E boot (...)`. House style is `area: lowercase imperative summary`, e.g. `wifibroadcast: pass -d so msposd actually draws the OSD`.
* CI here builds all 99 boards for this path but covers none of the FPV consumers — `ci-matrix.py --stdin` reports "reaches no board of its own" — so hardware evidence is the only gate this change has.
* Local checks do pass on the branch: `test_shell_parse.sh`, `test_strip_shell_comments.sh` and `ci-matrix.py --self-test` are all clean.

Two of the inline findings below were also raised by the automated review and I agree with both.

PR #2411, opened 2026-09-13
rv1103: enable RTL8812AU wifibroadcast on SD-card images

kkbin505:
## Problem

The `rv1103_lite` SD-card image does not provide the complete support required to use an RTL8812AU USB adapter with wifibroadcast:

  - the writable overlay is not persisted across reboots;
  - the USB2 OTG port may remain in device mode instead of host mode;
  - `rtl8812au-openipc` does not build against the RV1103 Linux 5.10 kernel;
  - the board does not select the RTL8812AU driver or `wifibroadcast-ng`;
  - wifibroadcast always starts telemetry, including installations without a
    telemetry device.

This change adds an RV1103 SD-card kernel fragment with a persistent ext4 overlay, switches the RV1103 USB2 OTG PHY to host mode during boot, fixes the RTL8812AU Linux 5.10 build, and enables RTL8812AU and wifibroadcast for `rv1103_lite`.

## Hardware tested on

Luckfox Pico Mini RV1103 Model A (SD card)
SC3336 3MP Camera
RTL8812au wifi usb card

## Evidence

Before:

```

```

After:

```

```

## Scope

- [ ] No kernel patches under `general/package/all-patches/linux/` (those go to [OpenIPC/linux] https://github.com/OpenIPC/linux))
- [ ] No files specific to a single retail camera model (those go to [OpenIPC/builder](https://github.com/OpenIPC/builder))
- [ ] No probing or bring-up tooling (that goes to [OpenIPC/ipctool](https://github.com/OpenIPC/ipctool))
- [ ] Nothing under `general/overlay/` or in a shared `load_<vendor>` script hardcodes a value specific to my board
- [ ] Package sources come from an OpenIPC repository, and any version bump keeps at least the specificity of the pin it replaces (a new package should pin a full 40-character SHA)
- [ ] No `LD_PRELOAD`, and no binaries that cannot be rebuilt from source


flyrouter:
Custom configuration profiles are handled in the Builder repository. 
Only basic patches with minimum configuration for running systems on each specific processor are accepted in the Firmware repository.

kkbin505:
> Custom configuration profiles are handled in the Builder repository. Only basic patches with minimum configuration for running systems on each specific processor are accepted in the Firmware repository.

  Thanks for the clarification. Would the following split be acceptable?

  **Move to builder (Luckfox Pico Mini A device profile):**

  - Enable RTL8812AU, iw, and wifibroadcast.
  - Board-specific SD card partition layout and kernel boot arguments.
  - Board-specific USB host initialization script.
  - Device defaults, such as disabling telemetry and setting video parameters.

  **Keep in firmware (generic support and fixes):**

  - RTL8812AU build compatibility fixes.
  - Generic /init support for a persistent overlay on SD cards.
  - Generic support for enabling/disabling telemetry in wifibroadcast, preserving the existing default
  behavior.
  - Minimal generic RV1103 configuration.

  I would update this firmware PR to contain only the generic changes and submit a separate builder PR
  for the device profile. Does this match the expectations of this project?

openipc-ai:
Yes — that split is the right shape. Two boundary calls I would move, and some specifics so the rework lands first time.

### Move to builder — agreed, and there is a ready-made pattern

`builder.sh` copies `devices/<name>/*` wholesale over a fresh firmware clone, so a device directory can carry a defconfig, a kernel config, overlay files, even a package. Everything on your builder list has a precedent already in that tree:

- `devices/common/br-ext-chip-goke/configs/gk7205v200_fpv_defconfig` — an FPV variant of a whole SoC, enabling `BR2_PACKAGE_RTL8812AU_OPENIPC` and `BR2_PACKAGE_WIFIBROADCAST`.
- `devices/common/br-ext-chip-goke/board/gk7205v200/gk7205v200.generic-fpv.config` — the matching kernel config, kept beside it.
- `devices/gk7205v200_otg_generic/` — a device whose whole point is a USB role, with its own `gk7205v200.generic-otg.config`.
- Device-level `general/overlay/etc/init.d/S*` is routine (e.g. `devices/gk7205v200_lite_tiandy-tc-c321n/general/overlay/etc/init.d/S71devmem`), so `S22usbhost` needs no package at all — drop `rv1103-usb-host` and ship the script as an overlay file in the device directory.

Something like `devices/rv1103_fpv_luckfox-pico-mini-a/` holding the defconfig, the SD-card kernel config and `S22usbhost` covers your whole first list.

That also settles a practical problem. `rv1103_lite` is what `sysupgrade` serves to every rv1103 user from this repo, and it repacks against a 4096 KB kernel / 8192 KB rootfs cap. Adding the 8812au module, `iw`, e2fsprogs and the wifibroadcast dependency chain (`msposd`, `mavfwd`, `jsonfilter`, `yaml-cli-multi`) changes what all of them get, and spends a large part of that budget. A builder variant has neither problem.

### The RTL8812AU fixes belong in neither repo — they belong upstream

This is the one I would move off your *keep in firmware* list. `rtl8812au-openipc.mk` says:

```make
RTL8812AU_OPENIPC_SITE = $(call github,openipc,realtek-wlan,$(RTL8812AU_OPENIPC_VERSION))
RTL8812AU_OPENIPC_VERSION = rtl8812au_fpv
```

The source is OpenIPC's own, and `pr_compliance_checklist.yaml` ("A patch against an OpenIPC package is a bridge, not a substitute") redirects a fix like this to the owning repository. Every patched Realtek package in this tree points at a third-party upstream — jwrdegoede, libc0607 — and every `openipc/realtek-wlan` one carries zero patches; these two would be the first exception.

The good news is that the pin is a **branch**, not a SHA. Merge the fixes into `rtl8812au_fpv` on [OpenIPC/realtek-wlan](https://github.com/OpenIPC/realtek-wlan) and every build here picks them up with no firmware change at all.

One thing to look at while you move them: `0001` does more than fix a build. The `#if CONFIG_TXPWR_LIMIT` → `#ifdef` change and the `halrf_cmn_info_hook` prototype are plainly right, but commenting out `#define CONFIG_PROC_DEBUG` in `include/autoconf.h` removes the driver's proc debug interface for every board using this package, not only the Linux 5.10 ones. If the real problem is the 5.10 `proc_ops` conversion, guarding on `LINUX_VERSION_CODE` — or porting the handlers — keeps the other boards whole.

### Generic `/init` SD overlay — yes, firmware

Gated on an `overlay=/dev/*` token no other board sets, so it is inert everywhere else. Two things before it ships:

- **Format only when there is genuinely no filesystem.** As written, *any* mount failure runs `mke2fs -F` over the partition, which destroys the user's persistent overlay — their configuration and the camera's claim — on a transient error. The jffs2 branch just above already draws that distinction: it formats only after seeing `Magic bitmask ... not found`. `blkid` is in this busybox (`CONFIG_BLKID=y`) and gives you the same signal for ext4.
- **`mke2fs` only exists when a defconfig selects `BR2_PACKAGE_E2FSPROGS`** — busybox's `MKE2FS` applet is off in `general/package/busybox/busybox.config`. Worth a word in the branch, or a tmpfs fallback, so a board that sets `overlay=` without it fails visibly rather than at 3am.

### Telemetry switch — yes, firmware, with the default the other way round

The diff ships `telemetry_enabled: false` in `general/package/wifibroadcast-ng/files/wfb.yaml`, which turns telemetry off on every wifibroadcast board in the tree. You already say you would preserve the default, so this is just the shipped value.

Two small things in the script itself:

```sh
[ "$telemetry_enabled" = "true" ] && start_telemetry
```

is the last command in `start()`, so `wifibroadcast start` now exits 1 whenever telemetry is off. The opt-out form

```sh
[ "$telemetry_enabled" = "false" ] || start_telemetry
```

returns 0, and also does the right thing if the key is missing from a hand-written `/etc/wfb.yaml`.

### "Minimal generic RV1103 configuration" — this one splits in two

The kernel fragment mixes two different kinds of thing:

- `CONFIG_WLAN`, `CONFIG_WIRELESS_EXT`, `CONFIG_WEXT_CORE`, `CONFIG_WEXT_PROC`, `CONFIG_CFG80211_WEXT` are SoC-generic. `rv1106.generic.config` currently carries `# CONFIG_WLAN is not set` and `# CONFIG_CFG80211_WEXT is not set`, and that is the same for all four rv11xx boards — exactly the "minimum configuration for running systems on each specific processor" @flyrouter described, and a small standalone PR against `rv1106.generic.config` worth opening on its own. Worth trimming to what the driver actually needs first: wifibroadcast drives the adapter entirely through `iw`/nl80211, so the WEXT symbols may be surplus.
- `CONFIG_CMDLINE` with `blkdevparts=mmcblk1:32K(env),512K@32K(idblock),…`, `root=/dev/mmcblk1p7`, `overlay=/dev/mmcblk1p6`, `rk_dma_heap_cma=24M` is an image layout for one board. Builder.

### Two housekeeping notes

- The `.github/scripts/ci-matrix.py` change drops `jsonfilter`, `mavfwd`, `msposd`, `rtl8812au-openipc`, `wifibroadcast-ng` and `yaml-cli-multi` from `NOT_BUILT` because the new defconfig selects them. Once the defconfig moves to builder none of that holds, so revert that file entirely — `--self-test` is a merge gate and will fail otherwise.
- The Evidence section is still two empty code blocks. Fine while this is a draft, but the firmware half — the `/init` change above all — wants before/after from the Luckfox: the boot log showing the overlay mounted, and a reboot proving something written to `/etc` survived it.

Happy to review the reworked firmware PR whenever it is ready.

PR #2431, opened 2026-09-16
majestic-webui: install the stub that fetches the installer, not the installer

openipc-ai:
## Problem

`gk7202v300_lite` no longer builds. From its CI log:

```
- uImage: [1857KB/2048KB]
- rootfs.squashfs: [5132KB/5120KB]
-- size exceeded by: 12KB
```

Last night's shipped image was 5108 KB in that 5120 KB partition — **12.0 KB**
of headroom — and this week's majestic-webui work spent it. Not one feature:
the payload grew ~189 KB uncompressed across nine merged PRs.

The daemon is not involved. For this family and variant its build was
1,244,608 B in the image that fit and is 1,244,736 B on S3 now: **+128 bytes**.

`/usr/sbin/updatewebui` is 48,779 bytes of that partition, and its only job is
to download a tree from GitHub — a camera that cannot reach GitHub could never
use it. OpenIPC/majestic-webui#519 splits it in two; this installs the half
that belongs on a camera.

## What changed

`MAJESTIC_WEBUI_INSTALL` moves `updatewebui-fetch` — a 4.5 KB stub that
downloads the installer from the branch being installed, verifies it, caches it
in `/etc/webui` so `--restore` still needs no network, and runs it with the
arguments it was given — onto the name people type.

Guarded on the stub actually being in the tarball rather than on `-f`. Which
files the WebUI ships is decided in another repo, so this degrades toward what
that repo already does — a release from before the split keeps its own
installer and the image is exactly as today — rather than toward an image with
no installer at all. **Either repo can land first.**

`/etc/profile` has fetched this same script for years and had two faults doing
it:

```sh
majestic_webui() {
	curl -s -L https://raw.githubusercontent.com/openipc/majestic-webui/master/sbin/updatewebui -o /tmp/webui.sh && sh /tmp/webui.sh
}
```

No `-f`, so a 404 is written to the file, `curl` exits 0, and `sh` runs the
error page. And no `"$@"`, so `--restore`, `--dry-run` and `--branch` are
unreachable through it. `majestic_webui()` now calls the stub instead of
repeating its `curl`; `legacy_webui()` keeps its own fetch — different repo,
different script — and gains the `-f`. Both pass arguments on.

## Hardware tested on

hi3516av300 + imx415, lite build. Described rather than pasted: OpenIPC's own
compliance rule forbids console output captured from a lab camera, and
describing it by hardware class and configuration is what makes it
reproducible.

This diff decides which file a camera carries under the name people type, so
what a board can settle is whether that command still does its job. It does,
on every path:

- **The ordinary one.** With the stub placed exactly where this diff places
  it, asking it to install a branch made it fetch that branch's installer and
  run it — the run announced installer v10 and planned the install normally.
  After a real install, `/usr/sbin/updatewebui` was the stub (8,186 bytes),
  `updatewebui-fetch` was absent, and the installer it had fetched was kept in
  `/etc/webui` (45,691 bytes).
- **A dry run keeps nothing.** With the kept copy deleted first, `--dry-run`
  fetched, ran, printed its plan, and left neither a copy nor a created state
  directory behind.
- **With GitHub unreachable.** `raw.githubusercontent.com` and `github.com`
  pointed at 127.0.0.1 in `/etc/hosts`. The control first: a plain `--branch`
  run failed, naming both recoveries — so the blackhole was real. Then
  `--url=file:///tmp/tree.zip`, against an archive already on the camera,
  **succeeded** and planned the install. That is the case that matters for a
  camera which must not be given a route out, and it is the path that worked
  before this split because the installer was on the image.
- **`--restore` with GitHub unreachable.** It ran from the kept copy with no
  network, removed what the install had placed, emptied the manifest, and left
  the firmware's own 48,768-byte installer showing through from `/rom`. Its
  offer to remove the emptied overlay directories correctly declined to act on
  a non-terminal stdin, so nothing rebooted.

The camera was left on its firmware WebUI with the manifest emptied.

One honest limit: this is an hi3516av300, not the gk7202v300 whose partition
motivated the change. The flash layout differs; the shell script does not.

## Evidence

Sizes measured by rebuilding the whole rootfs with the partition's own settings
(xz, 128 K blocks) on a workstation, from the published `nightly-20260915`
image for this board and the current `dist` tarball. The method is validated
before it is trusted:

Before:

```
control — re-squash the published image unchanged
  published image   5230592 B
  re-squashed       5230592 B   (+0 B, 0.00%)

payload as CI builds it today, installer carried
                    5296128 B
```

After:

```
same payload, installer fetched
                    5283840 B   (-12288 B)
```

Byte-for-byte on the control, so these are image sizes rather than estimates.
12 KB is the entire 12 KB by which CI reports `gk7202v300_lite` over its
5120 KB partition. It clears the error and no more: `CHECK_SIZE` warns below
32 KB left, so that board would go from red to a headroom warning. If more
margin is wanted the next levers are an excludes list for the tightest boards,
or the 191 KB idle in that board's own kernel partition — a boundary move, and
a flag-day for deployed cameras, so not folded in here.

## Scope


openipc-ai:
Closing this: the size fix no longer belongs here, and most of this diff has
become a no-op.

The reduction moved into the repository that decides what the payload contains
(OpenIPC/majestic-webui#520). Its dist build now renames the fetch stub onto
`sbin/updatewebui` while assembling the tarball, so what buildroot downloads
carries the 8 KB stub under the canonical name and never the 48 KB installer.
That takes effect the moment that repo's release is rebuilt, needs nothing
here, and leaves no second file list to keep in step with theirs.

Which makes this PR's `MAJESTIC_WEBUI_INSTALL` change dead: it looks for
`updatewebui-fetch` in the tarball, and after that rebuild there is no such
file. It would do nothing, while implying the payload still has a shape it
does not.

**One part of this is still a real fix and is not lost.** `/etc/profile`:

```sh
majestic_webui() {
	curl -s -L https://raw.githubusercontent.com/openipc/majestic-webui/master/sbin/updatewebui -o /tmp/webui.sh && sh /tmp/webui.sh
}
```

No `-f`, so curl writes the server's error page to the file and exits 0 — a
404 is downloaded and then handed to `sh`. And no `"$@"`, so `--restore`,
`--dry-run` and `--branch` cannot be reached through it. `legacy_webui()`
beside it has the same missing `-f`.

That is two lines and it is worth having, but it is not worth a full image
matrix on its own. It will ride along with the next change that has to come
through this repo anyway rather than being spent as a run of its own.

Sorry for the CI this already cost — a matrix ran here before the WebUI half
had landed, so the tarball it downloaded had no stub in it and those builds
could not have shown the effect either way.

PR #2446, opened 2026-09-18
Add Imou Cue 2 (IPC-C22EN) board support

HeytalePazguato:
## Summary

Add full board support for the **Imou Cue 2** (IPC-C22EN / IPC-C22EP), a sealed indoor camera built on Hi3516EV200 + SC2235 sensor over DVP parallel bus.

- Board defconfig based on `hi3516ev200_lite` with RTL8188FU WiFi and WPA supplicant
- SC2235 sensor driver patch: replaces the stock init table with Dahua's 114-entry DVP register sequence including the critical PCLK output-enable and pad-drive registers (`0x3d08`, `0x3640`, `0x3641`) that the stock driver omits — without these, no pixel clock reaches the SoC
- Sensor INI for DVP parallel input (`INPUT_MODE_CMOS`) with sync polarities verified by exhaustive 16-combination sweep
- Full IQ profile with boosted VPSS 3DNR temporal filtering (TFS 2→12, chroma NR sfc=255/tfc=30/trc=30) to compensate for the ISP not tracking gain dynamically at runtime
- WiFi init script for RTL8188FTV (GPIO 52 power-on)
- LED init scripts (green/red status LEDs)
- Customizer script with nightMode, audio (two-way via RTSP backchannel), and ONVIF defaults
- rc.local for sensor mirror/flip registers and PWM IR LED initialization
- Exclude list to strip unused sensor drivers and save flash space on 8MB NOR

## Hardware notes

The Imou Cue 2 runs **Dahua's signed U-Boot with RSA secure boot**. The UART console is locked and `loady`/`sf` commands are disabled. There is no software upgrade path — flashing requires a **CH341A SPI programmer** connected directly to the NOR flash chip.

## Testing

Tested on **4 cameras** (IPC-C22EN). All boot cleanly from cold power with working 1080p video, WiFi, two-way audio (RTSP backchannel), day/night switching, and ONVIF discovery. Verified stable across multiple power cycles with no manual intervention required.

## Changes

- `br-ext-chip-hisilicon/configs/hi3516ev200_lite_imou-cue2_defconfig` — board config
- `general/package/hisilicon-osdrv-hi3516ev200/` — SC2235 driver patch, sensor INI, IQ profile
- `general/overlay/` — majestic.yaml, rc.local, WiFi/LED init scripts, customizer
- `general/scripts/excludes/hi3516ev200_lite.list` — flash size optimization


general/overlay/etc/rc.local
```diff
@@ -1,14 +1,38 @@
...
+
+# --- Sensor init fixup + majestic restart + VPSS 3DNR ---
+# Sensor 180° rotation via hardware mirror+flip (reg 0x3221).
+# Majestic must restart AFTER to pick up the new orientation.
+# Then nrxset applies VPSS 3D noise reduction params directly
+# via HI_MPI_VPSS_SetGrpNRXParam (IQ profile 3DNR is not used).
+(
+  sleep 12
+  # Sensor mirror+flip + DVP pad enables
+  ipctool i2cset --bus 0 0x60 0x3221 0x06
```
openipc-ai, line 24: **`ipctool` is not on the image, so this line never runs.**

`general/package/ipctool/ipctool.mk` installs `ipcinfo` and nothing else — `BR2_PACKAGE_IPCTOOL=y` in your defconfig does not put a binary called `ipctool` anywhere. What answers to that name is `/usr/sbin/ipctool`, a symlink to `general/overlay/usr/sbin/extutils`, whose `ipctool)` arm does:

```sh
curl -s -L -f -o /tmp/ipctool https://github.com/OpenIPC/ipctool/releases/download/latest/ipctool"$ARC"
```

That is a convenience for a person at a shell ("installed as remote GitHub plugin", as it says). In a boot script it is four bad properties at once: a network fetch during boot, from a floating `latest` tag, repeated every boot because `/tmp` is tmpfs, and a silent no-op until the network is up — which on this board is WiFi, twelve seconds in.

Since the sensor registers here are what the whole change depends on, it is worth checking what your four cameras are actually doing. If they stream without these writes landing, the fix is elsewhere; if they need them, they need them somewhere that runs.

The right home for the pad enables is the sensor driver — see my note on the opensdk patch.

general/overlay/etc/rc.local
```diff
@@ -1,14 +1,38 @@
...
+  ipctool i2cset --bus 0 0x60 0x3d08 0x01
+  ipctool i2cset --bus 0 0x60 0x3640 0x01
+  ipctool i2cset --bus 0 0x60 0x3641 0x02
+  ipctool i2cset --bus 0 0x60 0x0100 0x01
+
+  # Restart majestic so it reads the updated sensor state
+  # and applies the boosted 3DNR params from the IQ profile.
+  sleep 2
+  killall majestic 2>/dev/null
```
openipc-ai, line 33: **Use `/etc/init.d/S95majestic restart`.**

`general/package/majestic/files/S95majestic` is not a thin wrapper, and a hand-rolled stop/start drops four things it does deliberately:

- it passes `-s`; this starts majestic with no arguments
- it finds the daemon with `start-stop-daemon -x`, never a pidfile
- it waits up to ten seconds for the old process to exit, because a second majestic started too early "dies on the busy sensor HAL" — `sleep 3` lands inside that window
- it starts under `trap '' HUP`, because the disposition survives `exec()` where a handler does not. That file's own comment records **25 deaths out of 25** without it, measured on a hi3516ev200 — the SoC this PR targets.

general/overlay/etc/rc.local
```diff
@@ -1,14 +1,38 @@
...
-# bits.
-#
-# By default this script does nothing.
+
+# --- IR LED: drive via PWM1 on GPIO4_7 (pin 39) ---
+# The IR LED driver circuit needs a PWM signal, not just GPIO high/low.
+# Enable PWM clock, set pinmux to PWM1, configure ~94% duty @ 24kHz.
+(
+  sleep 8
+  devmem 0x120101BC 32 $(( $(devmem 0x120101BC 32) | 0x80 ))
```
openipc-ai, line 8: **Board-specific MMIO in the shared `rc.local`.**

`S99rc.local` runs this on every board the tree builds, so these HiSilicon PWM and pinmux addresses are written on SoCs where they mean something else.

More generally, a `devmem` line is the most fragile way to record what is soldered to a pad: it is undone by anything that later exports the pad, and lost at the next reflash. Two better homes:

- `/usr/share/openipc/muxes.sh`, which `S30customizer` runs on every boot and which builder ships per device;
- `/api/v1/pinmux`, which reports and sets what each pad is wired to and remembers it — that is what the pins page in OpenIPC/majestic-webui (`www/a/mj-pins.js`) draws. `nightMode` also drives an IR illuminator from configuration, PWM lamp included, without anyone computing a duty cycle.

general/overlay/etc/ir/nrx_night_06.txt
```diff
@@ -0,0 +1,28 @@
+7 1 0
```
openipc-ai, line 1: **Nothing reads `/etc/ir/`.**

Neither this file nor `nrxset` beside it is referenced by any script, init file, package or configuration in firmware or builder — and the `rc.local` in this PR never invokes `nrxset`, although the comment above that code says it does.

Your two comments also disagree with each other. `rc.local` says the 3DNR parameters are applied from here and not from the IQ profile; the IQ profile says *"Vendor (Dahua) used nrxset runtime patching; we bake it into the profile."* Both cannot be true. If the profile is the one that is right, these two files can simply come out — which would also settle the binary-provenance gate that `nrxset` currently fails.

Worth stating plainly: `general/overlay/` needs no `Config.in`, so nothing in the build catches a file the camera never opens, and the cost lands on every board — several of which sit within 32 KB of their squashfs cap.

general/overlay/etc/init.d/S01leds
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+#
+# Early-boot LED state: red on, green off.
+# GPIO 0 = green (active low), GPIO 9 = red (active low), GPIO 4 = bottom red (active high).
+
+case "$1" in
+	start | "")
+		gpio set 0; gpio clear 9; gpio clear 4
```
openipc-ai, line 8: **These pins are driven on every camera, not just the Cue 2.**

`rcS` runs every `S??*` entry, so GPIO 0, 4 and 9 get exported and driven on boards where they are wired to something else.

There is already a declarative seam for this: `/usr/share/openipc/gpio.conf`, a per-device pin map (`led1`, `led2`, `light_ir`, `ircut1`/`ircut2`, `button`, `speaker`, `-1` for a pin the board does not have). 18 builder devices ship one, and consumers such as `general/package/quirc-openipc/files/qrscan.sh` read it.

It is not documented anywhere, which is not your fault — `en/gpio-settings.md` in the wiki is a human-readable pin table only, and nothing describes `gpio.conf`, `muxes.sh` or `customizer.sh`. I am fixing that separately.

general/overlay/usr/share/openipc/customizer.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+#
+# Perform basic settings on a known IP camera
+#
+#
+# Set SoC and sensor
+#
+fw_setenv soc hi3516ev200
```
openipc-ai, line 8: **This runs on the first boot of every image built from this tree.**

`S30customizer` executes `/usr/share/openipc/customizer.sh` if it is present, once, guarded by `/etc/custom.ok`. Because the shared overlay ships it to every board, an unrelated camera permanently writes the Cue 2's SoC, sensor, DVP flag, reset GPIO, wireless device and upgrade URL into its bootloader environment — and `load_hisilicon` and `sysupgrade` then read those back.

The upgrade URL is the one that bites hardest: the next `sysupgrade` on an unrelated camera fetches a Cue 2 image.

This file is right, it is just in the wrong tree. `devices/<board>/general/overlay/usr/share/openipc/customizer.sh` in OpenIPC/builder is exactly this path, and 96 devices already use it.

general/overlay/etc/majestic.yaml
```diff
@@ -0,0 +1,182 @@
...
+  httpsPort: 443
+  logLevel: debug
+  unsafe: false
+  buffer: 1024
+  plugins: true
+  telemetry: true
+  staticDir: /tmp/www
+isp:
+  antiFlicker: disabled
+  sensorConfig: /etc/sensors/sc2235_i2c_1080p.ini
```
openipc-ai, line 12: **This replaces the vendor default for every majestic board.**

`general/package/majestic/majestic.mk` installs `/etc/majestic.yaml` from the majestic tarball, and the rootfs overlay is rsynced over the target afterwards — so this file wins on every image, not just yours. A Goke board with a different sensor starts the streamer pointed at `sc2235_i2c_1080p.ini` and a Cue 2 IQ profile.

Per-device majestic settings belong in the device's own overlay in builder, or in `customizer.sh` as `cli -s` writes.

general/scripts/excludes/hi3516ev200_lite.list
```diff
@@ -0,0 +1,72 @@
...
+/usr/lib/sensors/libsns_f22.so
+/usr/lib/sensors/libsns_f23.so
+/usr/lib/sensors/libsns_f37.so
+/usr/lib/sensors/libsns_gc2053.so
+/usr/lib/sensors/libsns_gc2053_forcar.so
+/usr/lib/sensors/libsns_gc4023.so
+/usr/lib/sensors/libsns_gc4653_2l.so
+/usr/lib/sensors/libsns_gc5603.so
+/usr/lib/sensors/libsns_imx290.so
+/usr/lib/sensors/libsns_imx307.so
```
openipc-ai, line 12: **This strips sensors from the generic `hi3516ev200_lite` board, not just the Cue 2.**

`general/scripts/rootfs_script.sh` picks the list by `<model>_<variant>`, and both your defconfig and the existing generic `hi3516ev200_lite_defconfig` resolve to the same key. So the generic image loses `libsns_imx307.so`, `imx307_i2c_2l_1080p.ini` and the `default.ini` symlink that `hisilicon-osdrv-hi3516ev200.mk` points at `imx307.ini` — on a board that ships them on purpose.

Note this repository deliberately ships no `general/scripts/excludes/` directory at all; the mechanism lives here, the lists live in builder as `devices/<board>/general/scripts/excludes/<model>_<variant>.list`, where the device directory is what makes them per-board. 89 devices already do this.

general/package/hisilicon-opensdk/0001-sc2235-replace-init-table-with-dahua-dvp-sequence.patch
```diff
@@ -0,0 +1,190 @@
+--- a/libraries/sensor/hi3516ev200/smart_sc2235/sc2235_sensor_ctl.c
++++ b/libraries/sensor/hi3516ev200/smart_sc2235/sc2235_sensor_ctl.c
```
openipc-ai, line 2: **Open this against OpenIPC/openhisilicon instead.**

`HISILICON_OPENSDK_SITE` is `$(call github,openipc,openhisilicon,$(HISILICON_OPENSDK_VERSION))`, and `libraries/sensor/hi3516ev200/smart_sc2235/sc2235_sensor_ctl.c` is checked in there. Every other patch in this tree targets an upstream the project cannot commit to — ffmpeg, mbedTLS, vtund, libre, siproxd, ZeroTier, libwebsockets, f2fs-tools, the Realtek drivers. A downstream patch against our own code is invisible to anyone reading the SDK and is dropped silently at the next `*_VERSION` bump.

Two asks when you open it there:

- **Keep it to the deltas.** You identify the ones that matter — `0x3d08`, `0x3640`, `0x3641`, and the second streaming start. Replacing all 114 entries also drops about forty registers the current driver sets, for every existing EV200 board using this sensor, and the PR has no evidence from one of those.
- **Say where the table came from.** "Dahua's DVP sequence" read out of a factory driver has the same provenance problem as a binary: it cannot be checked against a datasheet, and the reason for any single register is gone.

Landing it upstream and bumping `HISILICON_OPENSDK_VERSION` here is the clean route, and it removes this file entirely.

br-ext-chip-hisilicon/configs/hi3516ev200_lite_imou-cue2_defconfig
```diff
@@ -0,0 +1,65 @@
...
+# BR2_PACKAGE_WIREGUARD_TOOLS is not set
+BR2_TARGET_ROOTFS_CPIO=y
+BR2_TARGET_ROOTFS_SQUASHFS=y
+BR2_TARGET_ROOTFS_SQUASHFS4_XZ=y
+
+# OpenIPC
+BR2_OPENIPC_SOC_VENDOR="hisilicon"
+BR2_OPENIPC_SOC_MODEL="hi3516ev200"
+BR2_OPENIPC_SOC_FAMILY="hi3516ev200"
+BR2_OPENIPC_VARIANT="lite"
```
openipc-ai, line 41: **Merge blocker:** a new defconfig must be registered with CI or `python3 .github/scripts/ci-matrix.py --self-test` fails, and that is a required check. Add the board to `ALL_BOARDS` in `.github/scripts/ci-matrix.py`, or to `UNBUILT_BOARDS` with a reason.

This applies wherever the board ends up — builder has its own matrix registration.

general/overlay/etc/wireless/usb
```diff
@@ -107,6 +107,13 @@ if [ "$1" = "mt7601u-hi3516ev200-camhi" ]; then
 	exit 0
 fi
 
+# HI3516EV200 Imou Cue 2
+if [ "$1" = "rtl8188fu-hi3516ev200-imou-cue2" ]; then
```
openipc-ai, line 111: This one is a judgement call rather than a clear-cut objection, so treating it as such.

`general/overlay/etc/wireless/usb` is a shared overlay file that has accumulated **46 arms keyed on retail model names** in 327 lines, while 18 builder devices override the whole file in their own tree. Those 46 are grandfathered rather than precedent — "additive, so it changes no existing board" is the right exemption for a `load_<vendor>` sensor case arm, and it was never meant to cover the overlay.

Since the rest of this device is moving to builder anyway, the Cue 2 arm travels most naturally with it, in the device's own copy of the file.

openipc-ai (changes requested):
Thank you for this — four cameras, a CH341A recovery path and a DVP sequence you clearly worked out the hard way is real work, and none of what follows says otherwise. But almost all of it is filed in the wrong repository, and a few parts do not do what the description says they do. Details are inline; the shape of it is here.

## This belongs in OpenIPC/builder

Not a technicality. `general/overlay/` is copied verbatim into every image this tree builds, so as written, every camera of every vendor gets `majestic.yaml` pointing at an SC2235, `rc.local` writing Hi3516EV200 PWM registers, LED scripts driving GPIO 0/4/9, and a `customizer.sh` that `fw_setenv`s them into believing they are an Imou Cue 2 — including the upgrade URL, on first boot, permanently.

Builder already has a home for every file here, at the identical paths: `devices/hi3516ev200_lite_imou-cue2-c22en/general/overlay/...`. 96 devices ship a `customizer.sh` that way, 18 a `gpio.conf`, 15 a `muxes.sh`, and 89 an excludes list.

## Three things I do not think work as intended

1. **`ipctool` is not on the image.** `general/package/ipctool/ipctool.mk` installs `ipcinfo` and nothing else, so `BR2_PACKAGE_IPCTOOL=y` supplies no binary of that name. `/usr/sbin/ipctool` is a symlink to `extutils`, which downloads the tool from a `latest` release into `/tmp` on first use. I suspect your four cameras work for a different reason, and it is worth finding out which — otherwise the fix is load-bearing on your bench and absent in the field.
2. **`nrxset` is a binary with no source**, which is a hard gate here (`pr_compliance_checklist.yaml`), and I cannot review what I cannot read. Nothing in firmware or builder reads `/etc/ir/` either.
3. **The majestic restart** re-opens several failures `S95majestic` exists to prevent.

## The SC2235 patch belongs upstream

`HISILICON_OPENSDK_SITE` is `$(call github,openipc,openhisilicon,…)`, and the file being patched is checked in there. Please open it as a PR against **OpenIPC/openhisilicon**, which fixes it for everyone rather than only for boards built from this tree.

## On the ISP not tracking gain

The behaviour the boosted 3DNR compensates for is worth its own issue on **OpenIPC/majestic** with your before/after, rather than being absorbed into a tuning profile here. A camera-side workaround leaves it in place for every other board.

## Mechanical blockers

- The defconfig needs registering in `ALL_BOARDS` in `.github/scripts/ci-matrix.py` (or `UNBUILT_BOARDS` with a reason) or `--self-test` fails the merge.
- `general/scripts/excludes/hi3516ev200_lite.list` breaks the generic board — see inline.

## What can stay

The `load_hisilicon` `sensor_dvp` opt-in is genuinely good and family-wide. Split that into its own small PR and it would be welcome as-is.

---

Separately, and not your problem: this PR exposed several gaps in our own review rules, which I have written up in #2452.

PR #2447, opened 2026-09-18
hi3516cv6xx: add the lite variant, and drop the digital TV stack to f…

eseverson:
### Problem

cv6xx has no image for 8 MiB NOR parts. It was added as a lite/8 MB target in 679c2c7e and renamed to ultimate/16 MB two days later in 8d5d02de, because *"the previous 8 MB lite variant cannot fit majestic + vendor MPP libs + 42 V5 kernel modules"*. 679c2c7e had already booked the choice — *"follow-up will either rebalance to 16 MB or trim rootfs"*. The rebalance happened; the trim did not, and cameras on 8 MiB parts have had nothing to flash since.

This does the trim, against the budget the cv610 u-boot actually enforces. Its compiled-in env is `256k(boot),64k(env),2048k(kernel),5120k(rootfs)`; the kernel slot is read whole by `sf read ${kernaddr} ${kernsize}` and root is mounted at the fixed rootfs offset, so an image boots from the release layout only if the FIT is under 2048 KB and the squashfs under 5120 KB. The Makefile's ceiling for this family is `FLASH_SIZE * 1024` on the combined blob, which measures neither; the 8 MiB arm now checks both halves against their slots, where a PR sees it.

Most of the kernel trim is one bug. The shared cv6xx kernel config builds 146 satellite and terrestrial television drivers into every image — DVB frontends (`cx24116`, `stv0900`, `si2168`), tuners (`tda18271`, `xc5000`, `r820t`), and the analog TV, radio and SDR stacks. Nothing in the tree can reach them: no defconfig enables `gst1-plugins-bad`, whose DVB plugins are the only DVB reference in userspace, and no overlay opens a `/dev/dvb` device. They are not there by choice. 679c2c7e regenerated this config from `hi3516cv{608,610}_defconfig` in the BSP, which sets `CONFIG_MEDIA_SUPPORT=y` and never mentions digital TV. In 5.10, `MEDIA_DIGITAL_TV_SUPPORT` and its siblings carry a prompt only if `MEDIA_SUPPORT_FILTER` is set, and otherwise default to `y` — so the expansion turned all six media classes on by itself. `rv1106` builds clean by a different route (the Rockchip tree patches `MEDIA_DIGITAL_TV_SUPPORT` to carry an unconditional prompt); `hi3519dv500` has the same root cause at a smaller scale (15 symbols) and is left alone.

The sensor path on this SoC is HiSilicon MPP through the out-of-tree `open_*.ko` modules, not V4L2, and nothing in the tree captures from a USB camera on it, so `MEDIA_SUPPORT` goes entirely rather than being filtered down to the UVC class. `REGMAP_I2C` goes with it: its only selector was `dvb-frontends`. In the same spirit, and for the 2048 KB slot:

* the eBPF syscall — no bpftool, libbpf or cgroup user anywhere in the tree; classic BPF for tcpdump stays (`NET` selects `BPF` independently)
* io\_uring — no caller in musl, busybox or majestic
* the NAND/UBI/UBIFS stack — no cv6xx NAND board exists in the tree and the family's repack is NOR-only
* the NFS client — cv6xx cannot NFS-root (`ROOT_NFS` and `IP_PNP` are off) and nothing mounts one at runtime
* the USB gadget stack, and ALSA, whose only user it was
* the zstd compressor, which only UBIFS and `CRYPTO_ZSTD` pulled in

`hi3519dv500`, the other V5 family, already builds without NAND, media, sound, BPF, io\_uring and zstd. zImage goes from 2722 KB to 2045 KB; the FIT lands at 2014 KB of its 2048.

The rootfs, to 5076 KB of its 5120:

* 15 ultimate-only packages dropped (`aws-webrtc`, `divinus`, `exfat` + `exfatprogs`, `iptables`, `lame`, `libwebsockets`, `mosquitto`, `motors`, `quirc`, `uacme`, `vtund`, `wireguard` compat + tools, `zerotier-one`).
* 1 MiB squashfs blocks instead of the 128 KiB default (measured on this rootfs: 128K 8704 KB, 256K 8512, 512K 8384, 1024K 8320). The price is about 2.5 MiB of RAM against 512K blocks, since the fragment cache and the xz dictionary scale with the block; the camera below has 19.9 MB available with them, and the 64 KB is an erase block this budget does not have.
* Buildroot's squashfs `EXTREME_COMP`, which for xz on ARM is `-Xbcj arm,armthumb`: 52 KiB. The kernel's xz decoder already carries `XZ_DEC_ARM` and `XZ_DEC_ARMTHUMB`, so there is nothing to pay at read time.
* The 42 vendor modules were installed unstripped: Buildroot's target-finalize skips `*.ko` on purpose and nothing else stripped them. `--strip-unneeded`, the operation `INSTALL_MOD_STRIP=1` performs, takes 16 KB off the squashfs. Both cv6xx variants get that.
* lite drops the eight modules nothing loads: `load_hisilicon` has no modprobe for adc, aiisp, devstat, spi\_dma\_transfer, user, user\_proc or uvc, and `open_svac3e` — the SVAC3 encoder — is probed only on the 20g/00s/00g dies, for a codec majestic has no code for. 60 KB. ultimate keeps all of them.

Four package symbols ultimate never has to state are spelled out here, because it gets them by side effect and lite does not: quirc and uacme select libcurl, libwebsockets and mbedtls drag in zlib and the TLS backend. sysupgrade shells out to curl over https, so a lite image without them has a sysupgrade that cannot fetch its manifest. Enabling only the libcurl parent is not enough — that yields a curl linked `TLS_NONE` against an absent libz.

Kept, because majestic will not start without them: the HiSilicon audio/VQE libraries its `DT_NEEDED` pulls in, plus opus, ogg, libevent, libcurl and protobuf-c. `dropbear`, `ipctool`, `yaml-cli` and `uboot-tools` survive.

The kernel config is the family's shared one, so `ultimate` builds the same kernel. I built ultimate's from this branch: it differs from the lite kernel by the netfilter symbols the iptables package fixup switches on (`IP_NF_IPTABLES`, `IP_NF_FILTER`, `NETFILTER_XTABLES`; 7.8 KB of zImage) and nothing else. `ci-matrix.py`'s self-test carries three hardcoded row counts that a new board changes; they are bumped here so the merge gate passes.

### Hardware tested on

hi3516cv6xx (chip reports `hi3516cv613`, DDR3 128 MB), SIMICAM A314D / Vatilon H80, 8 MiB NOR, GC4653 sensor.

The camera runs the release image for this build as #2448's `create_hisi` assembles it: the published `boot-hi3516cv610-20s-nor.bin` at 0, this branch's FIT at 0x50000 and its squashfs at 0x250000, erased env, on the partition table that u-boot declares. All three were read back from flash byte-identical (md5s below). The env was left to u-boot's defaults and `osmem` set to 48M through `fw_setenv`, which is the one step every cv610 install does. Two things the board adds that are not in this PR: `rootfs_data` was pre-staged rather than left for `/init` to format, because this sensor has no driver in the tree (`galaxycore_gc4653_2L` is in the ev200 V4 list, not the cv6xx V5 SDK's six — `ultimate` has the same gap), so the overlay carries an out-of-tree `libsns_gc4653_2l.so` and this board's MIPI-port-1 bring-up scripts, plus the usual device files. Nothing in the squashfs is replaced by anything else.

An ultimate image cannot be booted here: 8 MiB part. What the board did exercise is the shared kernel config in full; ultimate's kernel differs from it by the three iptables symbols above.

### Evidence

Clean build of this branch:

```
$ make BOARD=hi3516cv6xx_lite
- fitImage: [2014KB/2048KB]
- rootfs.squashfs: [5076KB/5120KB]
- firmware.bin: [7168KB/8192KB]

$ ls -l output/images/{zImage,fitImage,rootfs.squashfs,firmware.bin.hi3516cv6xx}
2045488 zImage
2063177 fitImage
5197824 rootfs.squashfs
7340032 firmware.bin.hi3516cv6xx

$ md5sum boot-hi3516cv610-20s-nor.bin output/images/fitImage output/images/rootfs.squashfs
37c9dd1798718e64371875a98d65ce4a  boot-hi3516cv610-20s-nor.bin   (github.com/openipc/firmware/releases/download/latest)
ee9b6cca762d36abb82fa8bc86229a8a  fitImage
7732a4955365664028859353f2d68a29  rootfs.squashfs

$ python3 .github/scripts/ci-matrix.py --self-test
ci-matrix: self-test ok (100 boards, 136 packages, 56 cases)
```

The kernel change, before and after, in the file this PR edits:

```
before:  grep -cE '^CONFIG_(DVB_|MEDIA_TUNER_)[A-Za-z0-9_]+=(m|y)'  ->  144
         grep -c '=m$'                                             ->  147
after:   grep -cE '^CONFIG_(DVB_|MEDIA_TUNER_)[A-Za-z0-9_]+=(m|y)'  ->    0
         grep -c '=m$'                                             ->    1   # CONFIG_FW_LOADER=m
```

On the camera, 2 minutes after the reboot that applied `osmem`:

```
# cat /proc/mtd
mtd0: 00040000 00010000 "u-boot"
mtd1: 00010000 00010000 "env"
mtd2: 00200000 00010000 "kernel"
mtd3: 00500000 00010000 "rootfs"
mtd4: 00700000 00010000 "firmware"
mtd5: 000b0000 00010000 "rootfs_data"

# cat /proc/cmdline
mem=48M console=ttyAMA0,115200 panic=20 root=/dev/mtdblock3 rootfstype=squashfs init=/init mtdparts=sfc:256k(u-boot),64k(env),2048k(kernel),5120k(rootfs),7168k@0x50000(firmware),-(rootfs_data)

# fw_printenv osmem totalmem socmodel
osmem=48M
totalmem=128M
socmodel=20s

# what is on the flash
dd if=/dev/mtd0 | head -c 241664  | md5sum  -> 37c9dd1798718e64371875a98d65ce4a   = boot-hi3516cv610-20s-nor.bin
dd if=/dev/mtd2 | head -c 2063177 | md5sum  -> ee9b6cca762d36abb82fa8bc86229a8a   = fitImage
dd if=/dev/mtd3 | head -c 5197824 | md5sum  -> 7732a4955365664028859353f2d68a29   = rootfs.squashfs

# cat /usr/lib/os-release | grep -E 'OPENIPC_VERSION|BUILD_'
OPENIPC_VERSION=2.6.09.18
BUILD_OPTION=lite
BUILD_PLATFORM=hi3516cv6xx_lite

# majestic --version
Lite HiSilicon (hi3516cv6xx), master+4832311, 2026-09-18 17:31

# lsmod | awk 'NR>1{print $1}' | sort | tr '\n' ' '        (34 loaded, 34 shipped, 0 dvb/tuner)
open_acodec open_adec open_aenc open_ai open_aio open_ao open_base open_chnl
open_h264e open_h265e open_isp open_ive open_jpege open_mipi_rx open_mmz open_osal
open_piris open_pm open_pwm open_rc open_rgn open_sensor_i2c open_sensor_spi
open_svp_npu open_sys open_sys_config open_vb open_vca open_venc open_vgs open_vi
open_vpp open_vpss open_wdt

# netstat -ltn | awk 'NR>2{print $4}' | sort -u
0.0.0.0:22
0.0.0.0:554
0.0.0.0:80

# head -4 /proc/umap/venc
      id   width  height   type  by_frame   sequence
       0    2560    1440   H264         y        773
       1     640     360   H264         y        387

# curl -o /dev/null -w '%{http_code} %{ssl_verify_result}' https://github.com/OpenIPC/firmware/releases/latest
302 0                       # the image's own curl, mbedtls and CA bundle: what sysupgrade uses
# curl -o /dev/null -w '%{http_code}' http://127.0.0.1/
200
$ ffprobe -rtsp_transport tcp -i rtsp://<camera>:554/stream=0
method OPTIONS failed: 401 Authentication required

# free
              total        used        free      shared  buff/cache   available
Mem:          43112       20008        9724         100       13380       19872

# df /overlay
/dev/mtdblock5             704       288       416  41% /overlay
```

The config was generated with `olddefconfig` against openipc/linux 5.10.221 and applied as a symbol-only edit: the diff is `CONFIG_` lines only.

### Scope

- [x] No files specific to a single retail camera model — the device's own prune list and overlay stay out
- [x] Nothing under `general/overlay/` or a shared `load_<vendor>` script hardcodes a value specific to my board
- [x] Package sources unchanged; no version bumps
- [x] No `LD_PRELOAD`, no unbuildable binaries
- [x] New code is selected by a defconfig, and registered in `ALL_BOARDS`


openipc-ai (commented):
The media diagnosis is the good part of this and I could reproduce the reasoning: on stock 5.10 the five `MEDIA_*_SUPPORT` classes carry a prompt only under `MEDIA_SUPPORT_FILTER` and otherwise default `y`, so the BSP expansion really did turn 146 TV drivers on by itself. `REGMAP_I2C` falling out with them checks out too -- `olddefconfig` would have kept it if anything else selected it.

Gates pass here: `ci-matrix.py --self-test` -> ok (100 boards, 136 packages, 56 cases), and the kernel config path narrows to both cv6xx boards, so ultimate does get built. `EXTREME_COMP` -> `-Xbcj arm,armthumb` is confirmed in `fs/squashfs/Config.in` and the kernel has `XZ_DEC_ARM`/`XZ_DEC_ARMTHUMB`/`XZ_DEC_BCJ`, so that 52 KiB is free as claimed.

The thing the hardware test cannot see is the bootloader, which the PR is honest about. I decompressed the cv610 u-boot this firmware ships with (`boot-hi3516cv610-*-nor.bin`, a HiSilicon boot table plus a gzip member at offset 40128) and its compiled-in env is:

```
kernaddr=0x50000   kernsize=0x200000   rootaddr=0x250000   rootsize=0x500000
mtdparts=sfc:256k(boot),64k(env),2048k(kernel),5120k(rootfs),7168k@0x50000(firmware),-(rootfs_data)
bootnor=... sf read ${baseaddr} ${kernaddr} ${kernsize}; bootm ${baseaddr}
bootargs=... root=/dev/mtdblock3 ...
```

Nothing rewrites it later -- `set_allocator` only edits the mmz half of `bootargs`. Against this build's own numbers:

| | this PR | partition |
| --- | --- | --- |
| fitImage | 2496 KiB | 2048 KiB :x: |
| rootfs.squashfs | 5184 KiB | 5120 KiB :x: |
| firmware.bin | 7680 KiB | 7168 KiB :x: |

The tested camera does not exercise any of this: its vendor u-boot reports `192K(uboot),2496K(kernel),5184K(rootfs),320K(rootfs_data)`, which is sized to this exact build, so every bound happens to fit there and nowhere else.

Two consequences worth knowing. A camera running OpenIPC's own u-boot cannot `sysupgrade` to this build -- `check_combined_fits` measures the FIT against the `kernel` partition and dies with "the FIT ... is 2496 KB, /dev/mtd2 holds 2048 KB. Nothing was written." That is a safe failure, no half-written flash. And a fresh whole-chip flash of the assembled `.bin` would boot a FIT truncated at 2048 KiB by `sf read`.

**This is not something this PR introduced.** The published cv6xx *ultimate* blob has the same mismatch and worse (FIT 2709 KiB, rootfs at +2752 KiB where u-boot expects +2048, blob 10496 KiB against a 7168 KiB partition), and hi3519dv500 has the offset half of it too. It looks like the cv610 env needs fixing rather than this branch needing another 448 KiB of trim -- but since this PR is what turns that into a published 8 MiB image, it seems worth settling first. I have put the detail on #2448, whose `create_hisi` is where the layout is actually decided.

On the bot review: finding 3 (curl parent + TLS backend) is genuinely addressed, and the defconfig comment explaining why `lite` has to name `zlib`/`mbedtls`/`libcurl` by hand where `ultimate` gets them through `quirc` and `uacme` is exactly the kind of comment this tree wants. Finding 2 reads stale against the current description, which claims byte-identical md5s from a clean build with the device-specific files on the JFFS2 overlay rather than in the squashfs.

Smaller points inline.

br-ext-chip-hisilicon/configs/hi3516cv6xx_lite_defconfig
```diff
@@ -0,0 +1,100 @@
...
+BR2_OPENIPC_SOC_VENDOR="hisilicon"
+BR2_OPENIPC_SOC_MODEL="hi3516cv6xx"
+BR2_OPENIPC_SOC_FAMILY="hi3516cv6xx"
+# One cv6xx image serves dies that report distinct ids (hi3516cv610,
+# hi3516cv608) via fw_printenv/ipcinfo. Upgrade identity comes from
+# BUILD_PLATFORM=hi3516cv6xx_lite; these aliases are a defensive route
+# for any image/camera that still surfaces the per-die id.
+BR2_OPENIPC_SOC_ALIASES="hi3516cv610 hi3516cv608"
+BR2_OPENIPC_VARIANT="lite"
+BR2_OPENIPC_FLASH_SIZE="8"
```
openipc-ai, line 11: `FLASH_SIZE="8"` feeds `CHECK_SIZE firmware.bin 8192` through the `hi3516cv6xx` arm of `repack`, and that number never subtracts the 320 KiB the firmware partition starts at. So `make` printed `[7680KB/8192KB]` -- 512 KB of headroom -- for an image that has at most 192 KB against the chip and is already over against the partition table u-boot declares.

The 32 KB headroom warning is measuring the wrong thing here for the same reason, which matters more than usual on a family this close to its cap.

The Makefile already special-cases `hi3516cv6xx` and `hi3519dv500` on their own lines, so subtracting each family's firmware offset there is a one-line change per family, and it would put the failure in the board matrix where a PR sees it rather than in the release job weeks later.

br-ext-chip-hisilicon/configs/hi3516cv6xx_lite_defconfig
```diff
@@ -0,0 +1,100 @@
...
+# Wireless stack omitted — openipc/linux hi3516cv6xx kernel has
+# CONFIG_WIRELESS off, so cfg80211/mac80211-dependent USB WiFi drivers
+# (mt7601u, rtl8188fu) and wpa_supplicant/wireless-tools have nothing
+# to bind to. Re-enable once the kernel config gains the wireless stack.
+BR2_TARGET_ROOTFS_CPIO=y
+BR2_TARGET_ROOTFS_SQUASHFS=y
+BR2_TARGET_ROOTFS_SQUASHFS4_XZ=y
+# 1 MiB blocks buy ~430 KiB over the 128 KiB default, measured across the
+# four sizes. The cost is decompression buffer per read on a 48 MiB part.
+BR2_TARGET_ROOTFS_SQUASHFS_BS_1024K=y
```
openipc-ai, line 59: The comment says the cost is a decompression buffer per read, but squashfs pins most of it at mount, and all of it scales with the block size. With this kernel's `CONFIG_SQUASHFS_FRAGMENT_CACHE_SIZE=3`, `CONFIG_SQUASHFS_FILE_CACHE=y` and `CONFIG_SQUASHFS_DECOMP_SINGLE=y`, a mount allocates 3 x block_size for the fragment cache, 1 x block_size for `read_page`, and an xz dictionary sized to the block:

```
128 KiB blocks -> ~0.6 MiB
512 KiB blocks -> ~2.5 MiB
  1 MiB blocks -> ~5   MiB
```

On a `mem=48M` camera whose own `free` in this PR shows 18584 KiB available, 1 MiB blocks spend roughly 2.5 MiB more than 512 KiB blocks do. Your own measurements price that at 64 KiB of flash (512K 8384 KB vs 1024K 8320 KB) -- one erase block.

That looks like the wrong side of the trade for a camera, and if the partition question above forces a retrim anyway, this is the cheapest 2.5 MiB of RAM on the board. Worth a `free` on the camera at 512K and 1024K to put a measured number on it rather than my arithmetic.

br-ext-chip-hisilicon/configs/hi3516cv6xx_lite_defconfig
```diff
@@ -0,0 +1,100 @@
...
+# EXTREME_COMP is Buildroot's switch for the compressor's arch options; for
+# xz on ARM it is `-Xbcj arm,armthumb`, so mksquashfs tries the ARM branch
+# filters on each block and keeps whichever packs smaller. 52 KiB on this
+# rootfs, which is the erase block it needed. Nothing to pay at read time:
+# the kernel's xz decoder already carries XZ_DEC_ARM and XZ_DEC_ARMTHUMB.
+BR2_TARGET_ROOTFS_SQUASHFS_EXTREME_COMP=y
```
openipc-ai, line 65: Not a change request, just so it does not get lost: no defconfig in the tree sets `EXTREME_COMP` today, and `BR2_TARGET_ROOTFS_SQUASHFS_COMP_OPTS` gives `-Xbcj arm,armthumb` to every `BR2_arm || BR2_aarch64` board using xz. This would be the first.

Three "back under the squashfs cap" commits landed this week (#2437, #2440, #2442). If it is really ~1% with nothing to pay at read time, the same flag in `general/openipc.fragment` is free headroom on every ARM board in the tree.

br-ext-chip-hisilicon/board/hi3516cv6xx/hi3516cv6xx.generic.config
```diff
@@ -1747,27 +1741,26 @@ CONFIG_MFD_SYSCON=y
 # CONFIG_RC_CORE is not set
 # CONFIG_MEDIA_CEC_SUPPORT is not set
 CONFIG_MEDIA_SUPPORT=y
-# CONFIG_MEDIA_SUPPORT_FILTER is not set
+CONFIG_MEDIA_SUPPORT_FILTER=y
```
openipc-ai: This file is the family's shared kernel config, so `hi3516cv6xx_ultimate` gets the same 146 drivers removed plus the BPF syscall, io_uring and zstd -- and the evidence in this PR is all `lite`.

I went looking for a reason it would not boot and did not find one: nothing in `general/package/` references `io_uring`, `liburing`, `libbpf` or `bpftool`; `NET` selects `BPF` in 5.10 independently of `BPF_SYSCALL`, so classic socket filters survive for tcpdump as you say; `MEDIA_USB_SUPPORT`, `USB_VIDEO_CLASS` and the videobuf2 path are kept; and zstd had only `UBIFS_FS_ZSTD` and `CRYPTO_ZSTD` behind it on a squashfs board. `ci-matrix.py --stdin` on this path does narrow to both cv6xx boards, so ultimate at least gets compiled.

Still worth one boot of an ultimate image before this merges -- 16 MiB cv6xx cameras are the ones with something to lose here, and CI only proves the image builds.

openipc-ai (changes requested):
Correcting my own verdict: I asked for two things before merge in https://github.com/OpenIPC/firmware/pull/2447#pullrequestreview-5254843699 and then submitted it as a comment. Restating them as the blocking pair, so the state matches the ask.

**1. The shared kernel config reaches `hi3516cv6xx_ultimate`, and every piece of evidence here is `lite`.** `br-ext-chip-hisilicon/board/hi3516cv6xx/hi3516cv6xx.generic.config` is the family's config, so the 16 MiB cameras lose the same 146 drivers plus `BPF_SYSCALL`, `IO_URING` and zstd. I looked for a reason that would not boot and did not find one, and `ci-matrix.py --stdin` does narrow to both boards so it gets compiled -- but compiling is not booting, and ultimate is the half of the family with cameras in the field. One boot of an ultimate image, with `dmesg` and a stream, closes it.

**2. The lite artifact this PR adds cannot be installed on a cv6xx camera running OpenIPC's own u-boot.** Its `mtdparts` give `2048k(kernel),5120k(rootfs)`; this build is a 2496 KiB FIT and a 5184 KiB rootfs, so `sysupgrade`'s `check_combined_fits` dies with "the FIT ... is 2496 KB, /dev/mtd2 holds 2048 KB. Nothing was written," and a whole-chip flash of the assembled `.bin` boots a FIT truncated by `sf read ${kernaddr} ${kernsize}`.

I flagged in the first review that this mismatch is pre-existing -- ultimate has it and worse -- and I let that soften the verdict, which was the wrong call. That the cv610 env needs fixing is an argument for fixing the env, not for publishing a second image into a layout nothing fits. Either the env moves or the build fits 2048/5120; either way it wants settling before an 8 MiB image is offered to people whose recovery path is a clip.

Everything else in that review stays advisory: the block-size RAM trade, the `CHECK_SIZE` offset, and the `EXTREME_COMP` observation are all worth doing and none of them should hold this up.

The media work itself is good and I would like to see it land.

eseverson:
Both blocking points addressed in `946afa4d`; the description is rewritten around the budget you measured.

**Fits the cv610 u-boot's slots.** FIT 2014 KB of 2048, squashfs 5076 KB of 5120, `firmware.bin` 7168 KB. Kernel side, on top of BPF/io_uring/zstd: `MEDIA_SUPPORT` goes entirely (the filtered-down UVC class had no consumer on this family either), plus NAND/UBI/UBIFS, the NFS client, USB gadget and ALSA — each with its reason in the message; hi3519dv500 already builds without all but NFS. Rootfs side: the vendor modules were shipped unstripped (`--strip-unneeded`, both variants), and lite drops the eight modules `load_hisilicon` never loads, `open_svac3e` among them. The Makefile's 8 MiB cv6xx arm now checks `fitImage` against 2048 and `rootfs.squashfs` against 5120, per your inline note, so the next overrun fails in the board matrix.

**Booted as the release image.** The camera now runs exactly what #2448's assembler produces for this build: the published `boot-hi3516cv610-20s-nor.bin` at 0, erased env, this FIT at 0x50000, this squashfs at 0x250000. u-boot's own table came up in `/proc/mtd`, `/init` mounted `rootfs_data`, `fw_setenv osmem 48M` was the only setup, and it streams; u-boot, FIT and squashfs read back byte-identical. Full transcript in the description.

**Ultimate.** Cannot be booted here (8 MiB part). What I can offer: ultimate's kernel built from this branch differs from the booted lite kernel by exactly the three iptables symbols its package fixup enables (`IP_NF_IPTABLES`, `IP_NF_FILTER`, `NETFILTER_XTABLES`, 7.8 KB). The shared config change itself is what the board exercised. If someone with a 16 MiB board can do the one boot, the remaining delta is that.

On the block-size note: 1 MiB stays. Your arithmetic is right about the ~2.5 MiB; measured `free` with it is 19.9 MB available on `mem=48M`, and the 64 KB it buys is an erase block this budget does not have. `EXTREME_COMP` tree-wide is a good idea and a separate PR.

openipc-ai (approved):
Both blocking points are answered, and the second one is answered in the way I most wanted and least expected: the camera is now running the artifact the assembler produces, not a hand-laid approximation of it. Published `20s` u-boot at 0, erased env, FIT at 0x50000, squashfs at 0x250000, u-boot's own table showing up in `/proc/mtd`. That was the whole gap, and it is closed.

FIT 2014 KB of 2048 and squashfs 5076 KB of 5120 both fit the slots the cv610 env declares, and the Makefile now says so on the 8 MiB arm, so the next overrun fails in the board matrix instead of in the release job. Leaving the 16 MiB arm on the whole-blob figure with the reason written down is the right call -- that one is a u-boot table question and a size check here could not settle it.

I checked the wider trim for anything it could strand rather than take the list on trust. No cv6xx defconfig selects `usb-dual-role` or `ffmpeg-openipc`, so the USB gadget and ALSA removals have no consumer on this family; there is no cv610 or cv608 device under `devices/` in OpenIPC/builder that the NAND/UBI removal could leave without a filesystem; and `ci-matrix.py --self-test` still passes on `946afa4d`.

Ultimate stays unbooted and I am not going to hold this for it. The lite kernel that did boot is the same one, the variants diverge by the iptables symbols its package fixup adds, and cv6xx ultimate cannot be assembled into a release image at all until the 7744 KiB rootfs question is settled -- so a boot test there is gated on work that is not this PR.

Three notes, none blocking.

**The `.ko` strip comment overstates its precedent.** `INSTALL_MOD_STRIP=1` is not `--strip-unneeded`; the kernel's own `mod_strip_cmd` is `$(STRIP) --strip-debug`, which keeps the whole `.symtab`. `--strip-unneeded` is the more aggressive of the two and the modules demonstrably still load, so the code is fine -- but the comment claims an equivalence that is not there, and in this tree the comment is the reasoning. Worth saying which of the two the saving actually came from.

**I am withdrawing my block-size suggestion.** At 5076 of 5120 there are 44 KB left and 512 KiB blocks cost 64 KB, so what I proposed would now overflow the slot. The pinned squashfs cache is still ~5 MiB of a 48 MiB camera and that cost is real, but you have the measurements and the budget and I do not.

**Both halves clear by about 2%** -- 34 KB on the FIT, 44 KB on the rootfs, each just past the 32 KB headroom warning. That is the margin the warning exists to talk about, so this family is one ordinary shared-overlay commit from red in both directions at once.

Good work, and thank you for chasing the u-boot table rather than arguing with it.

openipc-ai (approved):
Re-approving after the rebase: my earlier approval was dismissed when the branch moved.

Rebased `946afa4d` onto master `aa495696` (which had picked up #2441, #2442, #2444 and #2452 since this branched) and force-pushed as `770560b3`. Clean rebase -- master's #2444 touched the Makefile in a different arm, so no conflict, and the resulting patch is byte-identical to yours apart from the blob hashes in that hunk header. Authorship unchanged.

Gates re-run on the rebased tree: `ci-matrix.py --self-test` ok (100 boards, 136 packages, 56 cases), `lint-workflow-shell.py --self-test` passed, and all 147 shipped scripts parse under busybox ash both before and after comment stripping.

The workflow runs needed maintainer approval -- fork PR -- so I approved them; the matrix is running now. Nothing else from me.

.github/scripts/ci-matrix.py
```diff
@@ -66,7 +66,7 @@
     # Hisilicon [HI3516CV500]
     "hi3516av300_lite", "hi3516av300_neo", "hi3516cv500_lite", "hi3516dv300_lite",
     # Hisilicon [HI3516CV6XX]
-    "hi3516cv6xx_ultimate",
+    "hi3516cv6xx_lite", "hi3516cv6xx_ultimate",
```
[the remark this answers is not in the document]
openipc-ai: Correct as written, and out of scope here by design: the full-flash NOR image is assembled by `image.yml`, which is #2448. What this PR publishes is `openipc.hi3516cv6xx-nor-lite.tgz`, the archive `sysupgrade` consumes; the programmer image for a blank 8 MiB part arrives with #2448, which now sizes and places it against the partition table the cv610 u-boot declares. Dismissing on that basis rather than as a non-issue.

br-ext-chip-hisilicon/configs/hi3516cv6xx_lite_defconfig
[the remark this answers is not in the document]
openipc-ai: Addressed in `770560b3`, and the finding was accurate against the description it was written on.

The camera now runs exactly what #2448's assembler produces for this build: the published `boot-hi3516cv610-20s-nor.bin` at 0, an erased env, this FIT at 0x50000 and this squashfs at 0x250000, with u-boot's own partition table coming up in `/proc/mtd`. All three read back byte-identical. The excludes list and the wider local kernel trim this finding refers to are not part of the build that was flashed.

br-ext-chip-hisilicon/configs/hi3516cv6xx_lite_defconfig
openipc-ai: Addressed: the Makefile's 8 MiB cv6xx arm now checks `fitImage` against 2048 and `rootfs.squashfs` against 5120, so an overrun in either half fails in the board matrix rather than in the release job. Leaving the 16 MiB arm on the whole-blob figure is the right call and the reason is written down.

br-ext-chip-hisilicon/configs/hi3516cv6xx_lite_defconfig
openipc-ai: Withdrawing this one. At 5076 KB of the 5120 KB slot there are 44 KB left, and 512 KiB blocks cost 64 KB on this rootfs by your own table, so what I suggested would now overflow. The ~5 MiB of pinned squashfs cache on a 48 MiB camera is still a real cost, but the budget decides it and you have the measurements.

br-ext-chip-hisilicon/configs/hi3516cv6xx_lite_defconfig
openipc-ai: Observation only; nothing was owed here. Leaving it for whoever next looks at fleet-wide squashfs headroom.

br-ext-chip-hisilicon/board/hi3516cv6xx/hi3516cv6xx.generic.config
openipc-ai: Accepting this without the ultimate boot. The lite kernel that did boot is the same one, the two variants diverge by the iptables symbols the package fixup adds, and cv6xx ultimate cannot be assembled into a release image at all until its 7744 KB rootfs is reconciled with the 5120 KB slot — so a boot test there is gated on work that is not this PR. I also checked the wider trim for consumers: no cv6xx defconfig selects `usb-dual-role` or `ffmpeg-openipc`, and there is no cv610/cv608 device under `devices/` in OpenIPC/builder that the NAND/UBI removal could strand.

openipc-ai (approved):
CI is green on the rebased head: 121 checks pass, the only non-pass is the release-publish job that does not run on pull requests. Re-approving because my own thread replies each landed as a COMMENTED review and superseded the earlier approval.

PR #2448, opened 2026-09-18
ci: size the cv6xx NOR canvas to the variant it is built for

eseverson:
### Problem

`create_hisi` laid u-boot at 0 and the combined `firmware.bin` verbatim at the firmware offset, on a canvas hardcoded to 16384 KiB. Two things were wrong with that, one new and one old.

The new one: a lite variant is flashed to an 8 MiB part, and an image whose tail runs past the end of the chip cannot be written to one. The part size is now the sixth argument, defaulting to 16384 so every existing caller keeps its behaviour, and the cv6xx loop emits both variants at the size each is built for.

The old one: the boot images this job pairs with the firmware carry their partition table in a compiled-in env, and nothing rewrites it later. Decompressed from `boot-hi3516cv610-*-nor.bin` (a HiSilicon boot table with a gzip'd u-boot behind it):

```
cv6xx  256k(boot),64k(env),2048k(kernel),5120k(rootfs),7168k@0x50000(firmware),-(rootfs_data)
       bootnor: sf read ${kernaddr}=0x50000 ${kernsize}=0x200000; root=/dev/mtdblock3
dv500  512k(boot),256k(env),6144k(kernel),8192k(rootfs),14336k@0xC0000(firmware),-(rootfs_data)
```

So u-boot reads the kernel as one fixed slot and mounts root at a fixed offset behind it. `firmware.bin` is the FIT with the squashfs packed at the next 64 KiB boundary (board `post-image.sh`), which only agrees with that table when the FIT happens to end exactly at the slot. On every published cv6xx and dv500 image it does not: `root=/dev/mtdblockN` points 704 KiB before the squashfs on cv6xx ultimate and 512 KiB past it on dv500, and the cv6xx FIT is larger than its 2048 KiB slot, so `sf read` hands `bootm` a truncated image. `sysupgrade` already gets this right (`do_update_firmware` splits the blob at the FIT boundary and writes each half to its partition); a fresh whole-chip flash of the release `.bin` did not.

This does what sysupgrade does: split at the FIT boundary (the FIT's total size is the big-endian word at byte 4; the squashfs length is `bytes_used` in its superblock), put the FIT at the firmware offset and the rootfs at the rootfs partition, and measure each piece against the partition it goes into — u-boot against the boot partition (the env sits behind it, so the firmware offset is the wrong bound), the FIT against the kernel slot, the rootfs against the rootfs slot, the firmware partition against the part. Each failure prints a `::error::` and fails the step rather than uploading something that boots a truncated kernel or overwrites `rootfs_data`. The exact-size check on the assembled canvas stays as the backstop, and the scratch variables are `local` now.

The layout table is a `case` on `$soc`; adding a family means adding its u-boot's env line there. **Consequence worth stating:** with this in place, cv6xx ultimate on master fails the FIT check (2709 KiB in 2048) and hi3519dv500 ultimate fails the rootfs check (8832 KiB in 8192). Both are the same table disagreement the published images already carry, now reported instead of published. #2447 fits both cv6xx slots.

A refusal does not take the other SoCs' images with it. The step runs under Actions' default `bash -e`, and the Sigmastar, Allwinner and Ingenic images are assembled before the HiSilicon loops, so a bare `return 1` there would have ended the step with their `.bin` files sitting in `target/` and the upload never reached. The HiSilicon calls collect their failures (`|| hisi_failed=1`), the step exits 1 after the last one so the run is red, and `Upload` runs on `always()` so what did assemble ships. Exercised in the harness: a refused ultimate followed by a good lite leaves the lite `.bin` in `target/` and the step red.

### Hardware tested on

hi3516cv6xx (chip reports `hi3516cv613`, DDR3 128 MB), SIMICAM A314D / Vatilon H80, 8 MiB NOR.

This produces the bytes a user writes to a chip, so it was tested as that: the function from this branch, run against the #2447 lite build and the published `boot-hi3516cv610-20s-nor.bin`, produced `openipc-hi3516cv610-ddr3-128m-nor-lite.bin`, and that image was written to the camera's NOR in place (u-boot, erased env, FIT, rootfs, at the offsets above). It booted on the OpenIPC u-boot with its default env, the kernel reported the table from that env, `/init` found and mounted `rootfs_data`, and after the usual `fw_setenv osmem 48M` it streams. u-boot, FIT and squashfs read back byte-identical from their partitions. One board addition: `rootfs_data` was pre-staged with the device's overlay (an out-of-tree sensor driver, since this sensor has no driver in the tree) instead of left for `/init` to format; the rest of the image is the assembled one.

### Evidence

```
$ python3 .github/scripts/lint-workflow-shell.py
all run blocks parse clean
$ python3 .github/scripts/lint-workflow-shell.py --self-test
self-test passed

# the function from this branch, against the #2447 lite build:
Created: target/openipc-hi3516cv610-ddr3-128m-nor-lite.bin (FIT 2048 KiB in 2048, rootfs 5076 KiB in 5120)
   u-boot at 0 ....................... byte-identical to boot-hi3516cv610-20s-nor.bin
   256K-320K (env) ................... erased
   FIT at 320K ....................... byte-identical to fitImage (2063177 B)
   rootfs at 2368K ................... byte-identical to rootfs.squashfs (5197824 B)
   everything after ................... erased

# and the three ways it refuses:
   rootfs superblock claiming 5300 KiB  -> ::error::hi3516cv6xx_lite: the rootfs is 5300 KiB, the rootfs partition is 5120 KiB
   FIT header claiming 2200 KiB         -> ::error::hi3516cv6xx_lite: the FIT is 2240 KiB, the kernel partition is 2048 KiB
   blob that is not a FIT               -> ::error::hi3516cv6xx_lite: firmware.bin does not start with a FIT image
```

On the camera, running that image:

```
# cat /proc/mtd
mtd0: 00040000 00010000 "u-boot"
mtd1: 00010000 00010000 "env"
mtd2: 00200000 00010000 "kernel"
mtd3: 00500000 00010000 "rootfs"
mtd4: 00700000 00010000 "firmware"
mtd5: 000b0000 00010000 "rootfs_data"

# cat /proc/cmdline
mem=48M console=ttyAMA0,115200 panic=20 root=/dev/mtdblock3 rootfstype=squashfs init=/init mtdparts=sfc:256k(u-boot),64k(env),2048k(kernel),5120k(rootfs),7168k@0x50000(firmware),-(rootfs_data)

dd if=/dev/mtd0 | head -c 241664  | md5sum  -> 37c9dd1798718e64371875a98d65ce4a   = boot-hi3516cv610-20s-nor.bin
dd if=/dev/mtd2 | head -c 2063177 | md5sum  -> ee9b6cca762d36abb82fa8bc86229a8a   = fitImage
dd if=/dev/mtd3 | head -c 5197824 | md5sum  -> 7732a4955365664028859353f2d68a29   = rootfs.squashfs

# netstat -ltn | awk 'NR>2{print $4}' | sort -u
0.0.0.0:22
0.0.0.0:554
0.0.0.0:80
# head -4 /proc/umap/venc
      id   width  height   type  by_frame   sequence
       0    2560    1440   H264         y        773
       1     640     360   H264         y        387
```

### Scope

- [x] No kernel patches, no model-specific files, no bring-up tooling
- [x] Nothing in `general/overlay/` or a shared load script
- [x] No package source or version changes
- [x] No `LD_PRELOAD`, no unbuildable binaries
- [x] Existing callers unchanged by the new default


eseverson:
The lite variant is being added in this PR: https://github.com/OpenIPC/firmware/pull/2447

.github/workflows/image.yml
```diff
@@ -171,11 +173,14 @@ jobs:
...
             # The 20s/00s u-boots are picked per DDR (their 20g/00g siblings
             # carry the same DDR table, differing only in the socmodel env tag).
-            for variant in ultimate; do
-              create_hisi boot-hi3516cv610-10b-nor.bin hi3516cv6xx "$variant" hi3516cv610-ddr2-64m  320
-              create_hisi boot-hi3516cv610-20s-nor.bin hi3516cv6xx "$variant" hi3516cv610-ddr3-128m 320
-              create_hisi boot-hi3516cv610-00s-nor.bin hi3516cv6xx "$variant" hi3516cv610-ddr3-512m 320
-              create_hisi boot-hi3516cv608-nor.bin     hi3516cv6xx "$variant" hi3516cv608           320
+            # lite targets 8 MiB parts, ultimate 16 MiB -- the same kernel and
+            # rootfs blob, on the flash the variant was built for.
+            for variant in lite ultimate; do
```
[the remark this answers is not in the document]
eseverson: https://github.com/OpenIPC/firmware/pull/2447

.github/workflows/image.yml
```diff
@@ -144,11 +144,44 @@ jobs:
...
+
+              # Only $nor - $off KiB of the part is actually available to
+              # firmware.bin, because the firmware partition starts at $off.
+              # The build-side budget is FLASH_SIZE * 1024 and does not subtract
+              # the offset, so an image can pass `make` and still not fit the
+              # chip. dd would not complain: conv=notrunc stops it truncating
+              # the canvas, not extending it, so an oversized blob silently
+              # produces a release larger than the NOR part it is named for.
+              # Check it here rather than shipping something unflashable.
+              avail=$(( nor - off ))
```
openipc-ai: This is the wrong ceiling. `nor - off` is what is left of the *chip* above the firmware offset, but u-boot does not give firmware.bin the rest of the chip -- its `mtdparts` puts `rootfs_data` immediately after the firmware partition:

```
cv6xx:       ...,7168k@0x50000(firmware),-(rootfs_data)    -> rootfs_data starts at 7488 KiB
hi3519dv500: ...,14336k@0xC0000(firmware),-(rootfs_data)   -> rootfs_data starts at 15104 KiB
```

So `avail` should be 7168 (cv6xx) and 14336 (hi3519dv500), not 7872 and 15616.

It matters right now, not hypothetically. #2447's lite blob is 7680 KiB: it passes this check and then overwrites 512 KiB of a 704 KiB `rootfs_data` on an 8 MiB part. The published hi3519dv500 blob is 14464 KiB and is already 128 KiB into its own overlay partition.

Simplest fix that keeps the shape of this function is a seventh argument for the firmware partition size -- or derive it, since both callers already pass the offset.

.github/workflows/image.yml
```diff
@@ -144,11 +144,44 @@ jobs:
...
+              fw_kib=$(( ( $(stat -c%s "output/firmware.bin.$soc") + 1023 ) / 1024 ))
```
openipc-ai: `avail`, `fw_kib`, `ub_kib` and `got` are not `local` while every parameter on line 127 is. They leak into the caller's shell across the four `create_hisi` calls in each loop. Harmless today because each call reassigns them before use, but the asymmetry invites the one that is not reassigned.

.github/workflows/image.yml
```diff
@@ -144,11 +144,44 @@ jobs:
...
+              ub_kib=$(( ( $(stat -c%s "output/$uboot") + 1023 ) / 1024 ))
+              if [ "$fw_kib" -gt "$avail" ]; then
+                echo "::error::${soc}_${variant}: firmware.bin is ${fw_kib} KiB, but only ${avail} KiB is free above the ${off} KiB firmware offset on a ${nor} KiB part"
+                return 1
+              fi
+              if [ "$ub_kib" -gt "$off" ]; then
```
openipc-ai: `$off` is the firmware offset, not the space u-boot has. The env partition sits between them, so the real ceiling is smaller:

```
cv6xx:       256k(boot),64k(env)  -> u-boot must be <= 256 KiB, off is 320
hi3519dv500: 512k(boot),256k(env) -> u-boot must be <= 512 KiB, off is 768
```

A u-boot between those two numbers passes this check and silently eats the env partition. The current binaries are 236 KiB and 383 KiB so nothing is wrong today, but the check reads as if it proves something it does not.

.github/workflows/image.yml
```diff
@@ -144,11 +144,44 @@ jobs:
...
+              fi
+
+              # NOR canvas, erased (0xff); u-boot at 0, firmware.bin at its
               # partition offset (cv6xx MTDPARTS: firmware @ 0x50000 = 320 KiB).
-              dd if=/dev/zero bs=1K count=16384 status=none | tr '\000' '\377' > "$release"
+              # $nor is the part size in KiB -- a lite variant is flashed to an
+              # 8 MiB part, so a 16 MiB canvas would be unwritable past its end.
+              dd if=/dev/zero bs=1K count="$nor" status=none | tr '\000' '\377' > "$release"
               dd if="output/$uboot" of="$release" bs=1K seek=0 conv=notrunc status=none
               dd if="output/firmware.bin.$soc" of="$release" bs=1K seek="$off" conv=notrunc status=none
```
openipc-ai: This writes the combined blob down verbatim, which is where the rootfs offset is lost. `post-image.sh` places the squashfs at the first 64 KiB boundary after the FIT; u-boot's `rootaddr` is fixed (`0x250000` on cv6xx, kernel partition end on hi3519dv500). On every currently published image the two disagree -- +2752 vs +2048 KiB on cv6xx ultimate, +5632 vs +6144 KiB on hi3519dv500.

`sysupgrade`'s `do_update_firmware` already solves this: split at the FIT boundary, `dd` the FIT to `off` and the rootfs to the rootfs partition offset. Doing the same here would make the assembled release and the over-the-air upgrade produce identical flash contents, which is presumably the intent.

.github/workflows/image.yml
```diff
@@ -171,11 +204,14 @@ jobs:
...
             # The 20s/00s u-boots are picked per DDR (their 20g/00g siblings
             # carry the same DDR table, differing only in the socmodel env tag).
-            for variant in ultimate; do
-              create_hisi boot-hi3516cv610-10b-nor.bin hi3516cv6xx "$variant" hi3516cv610-ddr2-64m  320
-              create_hisi boot-hi3516cv610-20s-nor.bin hi3516cv6xx "$variant" hi3516cv610-ddr3-128m 320
-              create_hisi boot-hi3516cv610-00s-nor.bin hi3516cv6xx "$variant" hi3516cv610-ddr3-512m 320
-              create_hisi boot-hi3516cv608-nor.bin     hi3516cv6xx "$variant" hi3516cv608           320
+            # lite targets 8 MiB parts, ultimate 16 MiB -- the same kernel and
+            # rootfs blob, on the flash the variant was built for.
+            for variant in lite ultimate; do
```
openipc-ai, line 265: Ordering, not a defect: before #2447 lands, `firmware_url` returns empty for `hi3516cv6xx_lite`, the four lite calls print `Skip: no firmware in $BUILD_ID` and `return 0`. So this is safe to merge in either order, just inert until the defconfig exists.

openipc-ai (changes requested):
The sixth argument and the exact-size backstop are the right shape, and the hardcoded 16384 clearly had to go. But the bound the new check enforces is not the one that decides whether the image boots, and I think that has to be settled before this publishes an 8 MiB artifact.

The u-boot this job pairs with the firmware is a HiSilicon boot table followed by a gzip member (offset 40128 for cv610, 84768 for hi3519dv500). Decompressed, the compiled-in default env is:

```
# boot-hi3516cv610-*-nor.bin
kernaddr=0x50000   kernsize=0x200000   rootaddr=0x250000   rootsize=0x500000
mtdparts=sfc:256k(boot),64k(env),2048k(kernel),5120k(rootfs),7168k@0x50000(firmware),-(rootfs_data)
bootnor=... sf read ${baseaddr} ${kernaddr} ${kernsize}; bootm ${baseaddr}
bootargs=... root=/dev/mtdblock3 ...

# boot-hi3519dv500-*-nor.bin
kernaddr=0xC0000   kernsize=0x600000
mtdparts=512k(boot),256k(env),6144k(kernel),8192k(rootfs),14336k@0xC0000(firmware),-(rootfs_data)
```

Nothing rewrites that afterwards -- `set_allocator` only edits the mmz half of `bootargs`, and no overlay script touches `mtdparts` or `kernsize` -- so this is the table the camera boots with.

Measured against what this job currently assembles, and what it would assemble with #2447:

| | FIT | kernel part | rootfs at | rootfs part at | blob | firmware part |
| --- | --- | --- | --- | --- | --- | --- |
| cv6xx ultimate (published) | 2709 KiB | 2048 KiB :x: | +2752 KiB | +2048 KiB :x: | 10496 KiB | 7168 KiB :x: |
| cv6xx lite (#2447) | 2496 KiB | 2048 KiB :x: | +2496 KiB | +2048 KiB :x: | 7680 KiB | 7168 KiB :x: |
| hi3519dv500 ultimate (published) | 5586 KiB | 6144 KiB :white_check_mark: | +5632 KiB | +6144 KiB :x: | 14464 KiB | 14336 KiB :x: |

Two separate disagreements, both pre-existing and neither introduced here. `post-image.sh` packs the rootfs at the first 64 KiB boundary after the FIT; u-boot expects it at a fixed partition offset -- so `root=/dev/mtdblock3` points 704 KiB before the squashfs on cv6xx ultimate and 512 KiB past it on hi3519dv500. And on cv6xx the FIT is larger than the 2048 KiB `kernel` partition, so `sf read ${kernaddr} ${kernsize}` hands `bootm` a truncated image.

`general/overlay/usr/sbin/sysupgrade` gets both right where this assembler does not: `do_update_firmware` splits the blob at the FIT boundary and writes each half to its own partition, and `check_combined_fits` dies rather than write when either half is oversized. So an over-large combined image fails safe over the air -- it only mis-lands on a fresh whole-chip flash of exactly the `.bin` this job publishes, which is the artifact #2447 is asking for on parts nobody can reach with a programmer twice.

I am requesting changes on the bound, not on the parameterisation. Everything else here should stay. If splitting the blob the way sysupgrade does is too much for one PR, checking `fw_kib` against the firmware partition size (7168 / 14336) rather than `nor - off` at least stops this publishing something that overruns into `rootfs_data`.

One process note: "Hardware tested on: not applicable -- CI machinery that assembles a release artifact" is a stretch against CLAUDE.md, which exempts "CI machinery that only selects, lints or tests". This produces the bytes a user writes to a chip. `lint-workflow-shell.py` and its self-test do pass here, for what that covers.

eseverson:
Rewritten in `a8ff89e0` around the bound you identified, and then some: `create_hisi` now does what `sysupgrade`'s `do_update_firmware` does. It splits the blob at the FIT boundary (total size from the FDT header, squashfs length from `bytes_used`), puts the FIT at the firmware offset and the rootfs at the rootfs partition, and checks u-boot against the boot partition, the FIT against the kernel slot, the rootfs against the rootfs slot and the firmware partition against the part — each a `::error::` and a failed step. The layout table is a `case` on `$soc` carrying the two env lines you decompressed. Scratch variables are `local`.

Consequence, stated in the description: cv6xx ultimate on master now fails the FIT check and hi3519dv500 ultimate the rootfs check, which is the same disagreement the published images already carry, reported instead of published.

Hardware section is no longer "not applicable": the function from this branch, run against the #2447 build and the published `20s` u-boot, produced the lite image, and that image was written to the H80's NOR in place. It boots on the OpenIPC u-boot with its default env and streams; u-boot, FIT and squashfs read back byte-identical from their partitions. The three refusal paths were exercised with doctored blobs; output in the description.

.github/workflows/image.yml
```diff
@@ -171,11 +252,14 @@ jobs:
...
```
openipc-ai, line 265: This loop is where the failure escapes. `create_hisi` is a plain command here, the step runs under Actions' default `bash -e`, and `Upload` at line 275 has no `if:` -- so the first `return 1` from the ultimate pass ends the step with the Sigmastar, Allwinner and Ingenic images already built into `target/` and never uploaded, and the dv500 loop at 270 unreached.

cv6xx ultimate fails the FIT check (2752 KiB against the 2048 KiB slot) and dv500 fails the rootfs check (8832 KiB against 8192 KiB), both measured on the currently published blobs -- so this fires on the very first run after merge, not eventually.

Simplest shape that keeps the refusal without taking the rest of the release down:

```sh
rc=0
...
create_hisi ... || rc=1        # in place of the bare call
...
exit "$rc"                     # at the end of the step
```

with `if: always()` on the Upload step, so the images that did assemble still ship and the run still goes red.

openipc-ai (changes requested):
The rewrite is the right one, and it is more than I asked for -- splitting at the FIT boundary and placing each half where the env looks is exactly what `do_update_firmware` does, and I checked the arithmetic lands where it should: `320 + 2048 = 2368 KiB = 0x250000` on cv6xx, `768 + 6144 = 6912 KiB` on dv500, both matching the tables. `bytes_used` really is the little-endian u64 at superblock offset 40, the FDT total size really is the big-endian word at byte 4, and the four bounds are each the partition rather than the chip. My inline points are all addressed; the `local` line covers every scratch variable.

One thing I got wrong by omission in the first review, and it now blocks: **the consequence is not "reported instead of published", it is "nothing is published".**

- the `Prepare` step declares no `shell:`, so it runs under Actions' default `bash -e`
- `create_hisi` returns 1 as a plain command inside a `for` loop, so `set -e` ends the step
- `Upload` carries no `if:`, so it defaults to `success()` and is skipped

And the ordering decides the blast radius. Sigmastar and Allwinner assemble at line 238 and Ingenic at 243, both before the cv6xx loop at 257. Their `.bin` files are sitting in `target/` when the first cv6xx ultimate call fails; the dv500 loop at 270 never runs at all; and the upload at 275 never fires. So on the first scheduled assembly after this merges, the `image` release gets no new artifacts for any SoC -- not two withheld.

Refusing to publish a broken image is right. Ending every other SoC's publication to do it is not, particularly when the two targets being refused are already broken on master, so this is not a regression being caught but a standing disagreement being surfaced.

Report and skip would do it: `::error::` on the target, `return 0`, and let the rest of the run finish. If the run should still go red -- and it probably should, or nobody will look -- collect the failures instead of returning out of the loop, `exit "$rc"` at the end of the step, and give `Upload` an `if: always()` so what did assemble still ships.

Nothing else outstanding from me. `lint-workflow-shell.py` and its self-test pass on `a8ff89e0`.

eseverson:
Right, and thank you for tracing the blast radius rather than the symptom. `cd896b7e`: the HiSilicon calls collect (`|| hisi_failed=1`), the step exits 1 after the last one so the run stays red, and `Upload` is `if: always()` so the Sigmastar, Allwinner and Ingenic images that assembled before the refusal still ship. `create_hisi` also clears `output/` at entry now, so a refused call cannot leave its blob for the next one. Exercised in the harness with a refused ultimate followed by a good lite: the lite `.bin` is in `target/`, the error is printed, the step ends non-zero. Linter and its self-test pass.

.github/workflows/image.yml
```diff
@@ -184,11 +276,19 @@ jobs:
...
 
+            if [ -n "$hisi_failed" ]; then
+              echo "::error::one or more HiSilicon NOR images were refused (see above); the rest were assembled and will still be uploaded"
+              exit 1
+            fi
+
+      # always(): a refused HiSilicon image fails the Prepare step on purpose,
+      # and the images that did assemble must still ship.
       - name: Upload
+        if: always()
```
openipc-ai: `always()` also runs on cancellation, which is the one case where `target/` is guaranteed to be inconsistent: `create_hisi` lays down the erased canvas, then u-boot, then the FIT, then the rootfs, so a run cancelled between those steps leaves a `.bin` that is a bootloader on an otherwise blank part. This uploads it to the `image` release as though it were finished.

```yaml
if: ${{ !cancelled() }}
```

Same behaviour you are after -- the step still fails on a refused image, the images that did assemble still ship -- without overriding a cancellation. The `${{ }}` wrapper is required: `if: !cancelled()` is invalid YAML, since `!` at the start of a scalar is a tag indicator.

openipc-ai (changes requested):
The collect-and-continue shape is right, and `rm -rf output` at function entry is a catch I missed -- the old code only cleaned up on the success path, so a refused call left its extracted blob for the next one. Every refusal now gets its turn, the step still ends red, and what assembled still ships.

One thing left, and it is the same class of harm the PR exists to prevent.

`always()` does not mean "on failure". It means **on failure and on cancellation**. `create_hisi` builds `$release` in four stages -- erased canvas, u-boot, FIT, rootfs -- so a run cancelled between them leaves a partial `.bin` in `target/`: an 0xff canvas carrying a bootloader and no firmware. With `if: always()` the upload runs anyway and that lands in the `image` release as a flashable asset, next to the good ones and indistinguishable from them until someone writes it to a chip.

`if: ${{ !cancelled() }}` keeps everything you wanted from `always()` -- a refused image still fails the step, the images that did assemble still publish -- and lets a cancellation actually cancel. The `${{ }}` is not optional there: `if: !cancelled()` on its own is invalid YAML, because a leading `!` is a tag indicator.

Not a request, just so it is known rather than rediscovered: a `wget` failure still takes the `return 0` path, so a download that fails is skipped silently and does not set `hisi_failed`. That predates this PR and I would leave it alone here.

Housekeeping while I was in it: I rebased this onto master as `fdb2f321`, since #2447 landed and the required checks are `strict`. Clean rebase, patch byte-identical, authorship unchanged. Its workflow runs were sitting at `action_required` -- fork PR -- so I approved them; this is the first CI this branch has had.

.github/workflows/image.yml
openipc-ai: Addressed in the rewrite: the bound is now the `firmware` partition from `hisi_layout`, and it is checked per half -- the FIT against the kernel slot and the rootfs against `fw_kib - kern_kib` -- rather than the blob against what is left of the chip.

.github/workflows/image.yml
openipc-ai: Addressed: `local boot_kib kern_kib fw_kib fw fitsz fit_kib rootsz root_kib ub_kib got` covers every scratch variable now.

.github/workflows/image.yml
openipc-ai: Addressed: the check is against `boot_kib` (256 / 512) rather than `off`, so the env partition is no longer counted as space u-boot may use.

.github/workflows/image.yml
openipc-ai: Addressed, and thoroughly: the blob is split at the FIT boundary and each half written where the env looks -- FIT at `off`, rootfs at `off + kern_kib`. I checked both land right: 0x250000 on cv6xx, 6912 KiB on dv500.

.github/workflows/image.yml
openipc-ai: Moot: #2447 merged as 9dfe5595, so `firmware_url` resolves `hi3516cv6xx_lite` and the lite calls are no longer inert.

.github/workflows/image.yml
openipc-ai: Addressed: the calls collect with `|| hisi_failed=1`, the step exits 1 after the last one so the run still goes red, and the upload runs on failure so what did assemble still ships. `rm -rf output` at function entry closes the related hole I had missed -- a refused call used to leave its blob for the next one.

.github/workflows/image.yml
openipc-ai: Fixed in `fa47769a`, which I pushed onto this branch rather than hand back for one word -- see the note on the conversation. `if: ${{ !cancelled() }}`, with the reasoning moved into the comment above the step.

openipc-ai (approved):
Approving, with one disclosure: I pushed a commit onto this branch rather than hand it back for a single word.

`fa47769a` changes the upload condition to `if: ${{ !cancelled() }}` and moves the reasoning into the comment above the step. Nothing else in it. Your commit `85d943ed` is untouched and still carries the work; I added mine on top so the change is attributable and one revert away if you disagree with it.

Why I did it rather than ask: master had moved again (#2454), the required checks are `strict`, so this needed another rebase and another full CI cycle no matter what. Folding the one-word fix into that cycle cost nothing and meant not merging a defect I had just written a review about. If you would rather it were yours, squash it into `85d943ed` and force-push -- I will re-approve.

Everything else here is settled and I resolved the threads with what each turned into. For the record, what this PR ends up doing: it splits the combined blob at the FIT boundary, puts each half where the cv610 and dv500 envs actually look, refuses any image whose halves do not fit their slots, and keeps publishing the ones that do. cv6xx ultimate and hi3519dv500 will start failing that refusal on the first run -- both are genuinely broken against their own partition tables today, and the numbers are in my earlier review.

CI on `fa47769a` -- its fork runs needed approving again, so I did that too.

PR #2451, opened 2026-09-19
review: eleven rules a board-support PR walked straight through

openipc-ai:
## Problem

PR #2446 (Imou Cue 2 board support) is exactly the kind of change `best_practices.md` and
`pr_compliance_checklist.yaml` exist to catch: a whole retail camera filed into the shared
tree, with a binary blob, a boot-time download, and a sensor patch against one of our own
repositories. The automated review found **eight of the twelve problems in it**. The four
it missed and the five it half-caught are gaps in what we wrote down, not in the reviewer —
so this writes them down.

No hardware; this is review configuration only.

## What was missing

**The binaries gate could not see a binary.** It enumerated `.ko/.so/.a/.bin`, so
`general/overlay/etc/ir/nrxset` — an executable with no extension — came back from the
describe pass as *"the supplied patch contains no textual changes"*, which is precisely what
a blob looks like to anything reading a diff as text. The gate now keys on the
`Binary files ... differ` marker, which is the one signal that always survives, and says
outright that `general/overlay/` is never the place for a compiled artefact. A register
table copied out of a vendor's driver is folded in as the same failure written in C.

**`ipctool` is not on a camera.** `general/package/ipctool/ipctool.mk` installs `ipcinfo`
and nothing else; `/usr/sbin/ipctool` is a symlink to `extutils`, whose `ipctool)` arm curls
the tool from the `latest` release into `/tmp` on first use. In a boot script that is a
network fetch during boot, from a floating tag, repeated every boot because `/tmp` is tmpfs,
and a silent no-op until the network is up. #2446 ran `ipctool i2cset` from `rc.local`
twelve seconds in, on a WiFi-only board.

**An overlay file needs no `Config.in`, so nothing catches one nothing reads.** `/etc/ir/nrxset`
and `/etc/ir/nrx_night_06.txt` are read by nothing in firmware, builder or majestic, and the
PR's own two comments disagree about whether they are used at all. Cost lands on every board,
several of which sit within 32 KB of their cap.

**`S95majestic` is not a thin wrapper.** It passes `-s`, finds the daemon with
`start-stop-daemon -x` rather than a pidfile, waits up to ten seconds for the sensor HAL to
be released, and starts under `trap '' HUP` — its own comment records 25 deaths out of 25
without that, measured on a hi3516ev200. A hand-rolled `killall majestic; sleep 3; majestic &`
throws all of it away, which is what #2446 shipped, on that same SoC.

**Every patch in this tree targets an upstream we cannot commit to** — ffmpeg, mbedTLS,
vtund, libre, siproxd, ZeroTier, libwebsockets, f2fs-tools, the Realtek drivers. A patch
against `hisilicon-opensdk` does not: its `SITE` is `openipc/openhisilicon`, and the file
being patched is checked in there. OpenIPC/openhisilicon and OpenIPC/sensors join the
redirect table, which listed only linux, builder, ipctool and majestic.

**And the tree's per-device seams are documented nowhere.** `customizer.sh`, `muxes.sh`,
`gpio.conf`, `late-overlays.list` and the excludes lists are a complete mechanism, shipped
per device in builder at the same paths (96, 15, 18 and 9 devices respectively) — and absent
from the wiki, which is why somebody writes an `S01leds` with GPIO 0, 4 and 9 typed into it.
"Move it to builder" is half an answer; review should say which seam it becomes. Two smaller
notes ride along: the *additive, so it changes nothing* exemption belongs to `load_<vendor>`
case arms and not to the overlay, where `etc/wireless/usb` has quietly reached 46
retail-model arms in 327 lines; and an excludes list added **here** is keyed
`<model>_<variant>`, the generic board's key too, so #2446's list took `libsns_imx307.so`
and `default.ini` away from `hi3516ev200_lite` itself.

**`devmem` in a shipped script is now a question.** majestic #751 landed four days before
#2446 and its commit message is the rule: *"Somebody who solders an IR lamp, a button or an
I²C sensor to their board writes devmem into rc.local, because nothing in the camera knows
the pad exists. That line is undone by anything that exports the pad and lost on the next
reflash."*

## Changes

| File | |
| --- | --- |
| `best_practices.md` | new §1.4, §1.5, §2.5, §3.5, §4.4, §6.3, §7.3, §7.4; §2.4 and the §8 summary extended |
| `pr_compliance_checklist.yaml` | 3 new gates (OpenIPC-package patches, reachability, commands on the image); binaries, device-specific values and script conventions extended — 13 gates to 16 |
| `.pr_agent.toml` | the same, as reviewer guidance |
| `CLAUDE.md` | two rows in the redirect table, four bullets in "what the tree will not accept", the seams named |

`#2446` is still open. `best_practices.md`'s header said every rule traces to a *closed* PR;
that sentence now admits the occasional open one, rather than the citations quietly being
wrong.

## Hardware tested on

Not applicable, and not hedging: this is review configuration, documentation and repository
metadata, which `pr_compliance_checklist.yaml` and the template both name as exempt. No file
here reaches an image.

## Evidence

The checks this change class does admit:

```
$ python3 .github/scripts/ci-matrix.py --self-test
ci-matrix: self-test ok (99 boards, 136 packages, 56 cases)

$ python3 .github/scripts/lint-workflow-shell.py --self-test
self-test passed

$ git diff --name-only origin/master | python3 .github/scripts/ci-matrix.py --stdin
ci-matrix: 0/99 boards (needs_build=False) --- nothing that reaches a build
```

Both edited config files still parse, and `.pr_agent.toml` still carries no token that trips
Qodo's hardened loader (the failure mode from #2268):

```
$ python3 -c "import yaml; d=yaml.safe_load(open('pr_compliance_checklist.yaml')); print(len(d['pr_compliances']), 'gates')"
16 gates

$ python3 -c "import tomllib; d=tomllib.load(open('.pr_agent.toml','rb')); print(list(d))"
['github_app', 'review_agent']

$ grep -ciE 'ld_preload|@(json|format|py|get)' .pr_agent.toml
0
```

## Scope


openipc-ai:
Closing immediately — this PR body and diff quote a private source tree. Reopening with that content removed.

PR #2462, opened 2026-09-20
Simplify WireGuard U-Boot environment variable reads

usa-:
## Problem

The WireGuard startup script reads six U-Boot environment variables using six nearly identical `run_cmd` calls. This duplicates the same logic and makes the code unnecessarily repetitive.

## Hardware tested on

ssc378de camera.

## Evidence

The refactored script was deployed and tested on an `ssc378de` camera. The following output shows that the new loop is present on the device and that WireGuard was configured successfully:

```
root@ssc378de-imx415:~# grep -A3 'for name' /usr/sbin/wireguard
for name in privkey endpoint alive pubkey allowed address; do
	NAME=$(echo "$name" | tr '[:lower:]' '[:upper:]')
    run_cmd "Failed to read wg_$name U-Boot environment variable" WG_$NAME fw_printenv -n wg_$name
done
root@ssc378de-imx415:~# wg
interface: wg0
  public key: {hidden by me}
  private key: (hidden)
  listening port: 44677

peer: {hidden by me}
  preshared key: (hidden)
  endpoint: {hidden by me}:51820
  allowed ips: 10.0.0.0/8
  latest handshake: 12 seconds ago
  transfer: 30.80 KiB received, 6.21 KiB sent
  persistent keepalive: every 25 seconds
root@ssc378de-imx415:~#
```

## Scope

* [x] No kernel patches under `general/package/all-patches/linux/` (those go to https://github.com/OpenIPC/linux)
* [x] No files specific to a single retail camera model (those go to https://github.com/OpenIPC/builder)
* [x] No probing or bring-up tooling (that goes to https://github.com/OpenIPC/ipctool)
* [x] Nothing under `general/overlay/` or in a shared `load_<vendor>` script hardcodes a value specific to my board
* [x] Package sources come from an OpenIPC repository, and any version bump keeps at least the specificity of the pin it replaces (a new package should pin a full 40-character SHA)
* [x] No `LD_PRELOAD`, and no binaries that cannot be rebuilt from source


general/overlay/usr/sbin/wireguard
```diff
@@ -2,12 +2,10 @@
...
-run_cmd "Failed to read wg_privkey U-Boot environment variable" WG_PRIVKEY fw_printenv -n wg_privkey
-run_cmd "Failed to read wg_endpoint U-Boot environment variable" WG_ENDPOINT fw_printenv -n wg_endpoint
-run_cmd "Failed to read wg_alive U-Boot environment variable" WG_ALIVE fw_printenv -n wg_alive
-run_cmd "Failed to read wg_pubkey U-Boot environment variable" WG_PUBKEY fw_printenv -n wg_pubkey
-run_cmd "Failed to read wg_allowed U-Boot environment variable" WG_ALLOWED fw_printenv -n wg_allowed
-run_cmd "Failed to read wg_address U-Boot environment variable" WG_ADDRESS fw_printenv -n wg_address
+for name in privkey endpoint alive pubkey allowed address; do
+	NAME=$(echo "$name" | tr '[:lower:]' '[:upper:]')
+	run_cmd "Failed to read wg_$name U-Boot environment variable" WG_$NAME fw_printenv -n wg_$name
+done
```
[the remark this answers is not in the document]
usa-: Runtime output added.

general/overlay/usr/sbin/wireguard
```diff
@@ -2,12 +2,10 @@
...
 
-run_cmd "Failed to read wg_privkey U-Boot environment variable" WG_PRIVKEY fw_printenv -n wg_privkey
-run_cmd "Failed to read wg_endpoint U-Boot environment variable" WG_ENDPOINT fw_printenv -n wg_endpoint
-run_cmd "Failed to read wg_alive U-Boot environment variable" WG_ALIVE fw_printenv -n wg_alive
-run_cmd "Failed to read wg_pubkey U-Boot environment variable" WG_PUBKEY fw_printenv -n wg_pubkey
-run_cmd "Failed to read wg_allowed U-Boot environment variable" WG_ALLOWED fw_printenv -n wg_allowed
-run_cmd "Failed to read wg_address U-Boot environment variable" WG_ADDRESS fw_printenv -n wg_address
+for name in privkey endpoint alive pubkey allowed address; do
+	NAME=$(echo "$name" | tr '[:lower:]' '[:upper:]')
+	run_cmd "Failed to read wg_$name U-Boot environment variable" WG_$NAME fw_printenv -n wg_$name
```
openipc-ai: This is the part I'd like reconsidered. `WG_PRIVKEY` … `WG_ADDRESS` are now built at runtime, so `grep WG_ADDRESS` in this file finds only the consumer on line 41, never where it is set. Nothing else under `general/overlay/usr/sbin/` constructs a variable name for a fixed, known set — `common`'s `eval` on line 85 is the generic helper doing it once, precisely so callers can pass literal names.

Weighed against that: the change does nothing for any camera, and `general/overlay/` reaches every image, so the bar here is "strictly better", and losing a grep is not that.

The repetition you are reacting to is real, though — it is the message string, repeated six times. A local helper removes it without giving up the names or adding six `tr` forks:

```sh
read_wg() {
	run_cmd "Failed to read wg_$1 U-Boot environment variable" "$2" fw_printenv -n "wg_$1"
}

read_wg privkey  WG_PRIVKEY
read_wg endpoint WG_ENDPOINT
read_wg alive    WG_ALIVE
read_wg pubkey   WG_PUBKEY
read_wg allowed  WG_ALLOWED
read_wg address  WG_ADDRESS
```

More lines than the loop, but every name stays literal, the duplicated string is gone, and it reads the same way as `WG_SHARKEY` two lines below. Either that, or leave these six lines as master has them — both are fine by me; the runtime-built names are what I do not want in the shared overlay.

openipc-ai (changes requested):
The refactor is behaviour-preserving — I checked rather than assumed — and the gates are clean. One thing I'd like reconsidered before merge, plus a rebase.

## What I verified

Ran both forms under the busybox `ash` this tree builds, with `fw_printenv` and `run_cmd` stubbed:

* success path — identical `WG_*` assignments, all six;
* failure path (one variable missing) — identical message text, identical `exit 1`.

Gates on 937c867: `STRICT=1 test_shell_parse.sh` and `STRICT=1 test_strip_shell_comments.sh` both clean across 146 shipped scripts, `ci-matrix.py --self-test` ok. Indentation is consistent tabs in the blob — the mixed indent visible in the PR description is a paste artefact, not in the file.

## The Qodo rule-3 finding is a false positive — please ignore it, and don't add more evidence

The `wg` paste *is* runtime output from the board, and it proves more than the bot credits. All six `run_cmd` calls are fatal (no `false` argument, so `run_cmd_fatal=true` and a non-zero `fw_printenv` ends the script before `wg setconf`). A `wg0` carrying a completed handshake is therefore only reachable if all six reads returned 0. That is sufficient.

The unticked Scope box is fine too — this adds no new code for a defconfig to select.

## Rebase

Head sits on f07c206; master is now 443dd2a. No checks have run on this branch yet, and this path widens the matrix to all 100 boards, so it is worth rebasing before the run rather than spending it twice.

general/overlay/usr/sbin/wireguard
usa-: I agree that the variable names should remain explicit. I’ve changed the loop to use the full `WG_*` names directly, so they remain easy to find in the source while avoiding the repeated `run_cmd` calls.

I think this addresses the readability concern without needing a separate helper function.

general/overlay/usr/sbin/wireguard
```diff
@@ -2,12 +2,10 @@
...
 
-run_cmd "Failed to read wg_privkey U-Boot environment variable" WG_PRIVKEY fw_printenv -n wg_privkey
-run_cmd "Failed to read wg_endpoint U-Boot environment variable" WG_ENDPOINT fw_printenv -n wg_endpoint
-run_cmd "Failed to read wg_alive U-Boot environment variable" WG_ALIVE fw_printenv -n wg_alive
-run_cmd "Failed to read wg_pubkey U-Boot environment variable" WG_PUBKEY fw_printenv -n wg_pubkey
-run_cmd "Failed to read wg_allowed U-Boot environment variable" WG_ALLOWED fw_printenv -n wg_allowed
-run_cmd "Failed to read wg_address U-Boot environment variable" WG_ADDRESS fw_printenv -n wg_address
+for NAME in WG_PRIVKEY WG_ENDPOINT WG_ALIVE WG_PUBKEY WG_ALLOWED WG_ADDRESS; do
+	name=$(echo "$NAME" | tr '[:upper:]' '[:lower:]')
+	run_cmd "Failed to read $name U-Boot environment variable" "$NAME" fw_printenv -n "$name"
```
openipc-ai: The names are still derived, just in the other direction — so now it is the U-Boot side that no longer appears in the source.

Measured on this commit: `wg_endpoint`, `wg_alive`, `wg_pubkey`, `wg_allowed` and `wg_address` occur in no file in this repository. `wg_privkey` survives only because `S98wireguard` line 5 tests it, and `wg_sharkey` only because line 10 below still reads it literally. The wiki does not cover them either — `en/package-wireguard.md` documents the manual `/etc/wireguard.conf` route and never mentions these variables. This script is the only place in OpenIPC where the set is written down, and it is the set an operator types into `fw_setenv`, so I would rather not have it derivable-only.

That is what the helper was covering, and I should have said so plainly the first time: `read_wg privkey WG_PRIVKEY` puts both names on one line, so both greps work.

```sh
read_wg() {
	run_cmd "Failed to read wg_$1 U-Boot environment variable" "$2" fw_printenv -n "wg_$1"
}

read_wg privkey  WG_PRIVKEY
read_wg endpoint WG_ENDPOINT
read_wg alive    WG_ALIVE
read_wg pubkey   WG_PUBKEY
read_wg allowed  WG_ALLOWED
read_wg address  WG_ADDRESS
```

As it stands the change is a wash rather than an improvement, so the two outcomes I am happy with are that helper, or closing this and leaving master's six lines alone — no hard feelings either way, and the second costs you nothing.

openipc-ai (changes requested):
Thanks — that does fix what I asked for, and I checked rather than took it on trust: the six `WG_*` assignments are identical to master, and driving the fatal path once per variable gives the byte-identical message each time. Gates clean on ce42c7f8 (`STRICT=1 test_shell_parse.sh` and `STRICT=1 test_strip_shell_comments.sh` across 146 shipped scripts, `ci-matrix.py --self-test`).

But I think the revision trades one grep for the other rather than settling it, and the side it gives up is the user-facing one. Detail inline.

One thing still outstanding from the last round: the rebase. Base is still f07c206 and master is 443dd2a, and no checks have run yet on a path that fans out to all 100 boards.

general/overlay/usr/sbin/wireguard
usa-: I don’t think the proposed helper provides an advantage over the loop.

The names that may actually need to be searched for are the lowercase names — these are the U-Boot environment variables. The uppercase `WG_*` names are variables internal to this script, so there is little value in making them searchable.

For that reason, I would actually prefer keeping the U-Boot variable names explicit in the loop:

```sh
for name in wg_privkey wg_endpoint wg_alive wg_pubkey wg_allowed wg_address; do
	NAME=$(echo "$name" | tr '[:lower:]' '[:upper:]')
	run_cmd "Failed to read $name U-Boot environment variable" "$NAME" fw_printenv -n "$name"
done
```

This makes all U-Boot variable names directly searchable, while still avoiding six repeated `run_cmd` calls.

The proposed helper explicitly lists only `privkey`, `endpoint`, `alive`, etc., which are not the actual variable names and do not appear elsewhere in the codebase. I don’t see what additional readability or searchability that provides compared with the loop above.

openipc-ai (approved):
You are right, and my suggestion was wrong on the very test I was applying to yours.

`read_wg privkey  WG_PRIVKEY` does not contain the string `wg_privkey` — grepping that block for it returns nothing. The helper would not have made the U-Boot names searchable either; it just moved the problem one column over. Sorry for sending you round that loop.

Your ranking of the two name spaces is the right one, too. `WG_PRIVKEY` … `WG_ADDRESS` appear in no file in this repository outside this script, so the only grep that can ever matter for them is the within-file one, and the loop is at line 5 with the first consumer at line 26. The `wg_*` names are the configuration surface an operator sets with `fw_setenv`, they are documented nowhere else in OpenIPC — the wiki's `en/package-wireguard.md` only covers the manual `/etc/wireguard.conf` route — and 368f08ba puts all six back on one greppable line. That is the right trade, and it is better than where this started.

Verified on 368f08ba rather than taken on trust: identical `WG_*` assignments to master on the success path, and driving the fatal path once per variable gives the byte-identical message for all six. `STRICT=1 test_shell_parse.sh`, `STRICT=1 test_strip_shell_comments.sh` and `ci-matrix.py --self-test` all clean.

Approving. One optional nit, not worth a round on its own: line 7 went out with `$NAME` and `$name` unquoted, where your comment above had them quoted. Harmless here — both are loop literals with no whitespace or glob characters — so fold it in only if you happen to push again.

CI has still never run on this branch and the head is behind master, so I will update the branch to get the matrix onto a current tree. This path fans out to all 100 boards; merge once that is green.

general/overlay/usr/sbin/wireguard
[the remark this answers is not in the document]
openipc-ai: Dismissing this finding. The `wg` output in the PR description is runtime output from the board, and it proves what the rule asks for: all six `run_cmd` calls are fatal, so a `wg0` carrying a completed handshake is only reachable if every one of the six `fw_printenv` reads returned 0. The author also added further runtime output above. No change required.

Issue #2386, opened 2026-09-08
Feature request: Preserve empty lines in nightly firmware build scripts


usa-:
Currently, the shell scripts in the repository remove empty lines when building the nightly firmware. Based on my calculations, this saves no more than about 2 KB in the resulting firmware image.

If my calculation is correct, could we stop removing empty lines? This would make the generated scripts more readable without having any meaningful impact on firmware size.

widgetii:
I measured this rather than estimating it, and the answer is favourable to the request.

## Uncompressed

Preserving blank lines (while still stripping comments) adds **514 bytes** across the 45 shipped shell scripts I could attribute to a source in-tree. So the ~2 KB in the issue is an over-estimate by about 4x.

## On flash

Uncompressed bytes are not what reaches the camera — `mksquashfs ... -b 128K -comp xz` is, and it pads the final image to 4096 B. I rebuilt the rootfs both ways from the same same-day target tree, replicating buildroot's squashfs step exactly (`rsync` the target, then under `fakeroot`: `chown -h -R 0:0`, `makedevs` with the full devices table, drop the udev hwdb dirs, empty `run/` and `tmp/`, then `mksquashfs -noappend -b 128K -comp xz`). Done that way it reproduces buildroot's own output byte-for-byte, which is what makes a small delta meaningful.

Result on `hi3518ev300_lite`, both halves built in the same run:

```
as shipped (blank lines stripped): 5,226,496 B
blank lines preserved:             5,226,496 B
delta:                                     0 B
```

**Zero.** The 514 bytes disappear into existing block padding.

Because the image is padded to 4096 B, any real cost shows up in 4 KB units. The theoretical worst case for 514 B is one 4 KB unit on a board where it happens to tip a boundary — not a proportional cost, but not guaranteed zero on every image either.

## The change itself

One line in `general/scripts/strip-shell-comments.awk`:

```awk
-	if (/^[ \t]*$/) next
+	if (/^[ \t]*$/) { print; next }
```

Note that simply deleting the first check does **not** work — the later `if (code ~ /^[ \t]*$/) next` also catches blank lines, because `code_of("")` returns `""`. The line has to print explicitly.

I ran all 151 shipped scripts through the modified stripper under busybox `ash -n` (host busybox with `ASH_BASH_COMPAT`, as `test_shell_parse.sh` does — the busybox under `output-*/build/` is cross-compiled for ARM and cannot be used for this): **0 parse failures**.

## Scope

[...]

PR #2463, opened 2026-09-20
Preserve empty lines in generated shell scripts

usa-:
## Problem

Currently, `general/scripts/strip-shell-comments.awk` removes empty lines from shell scripts when building the nightly firmware.

This makes the generated scripts harder to read and understand, especially for someone who needs to inspect the script and understand its structure or behaviour. Preserving empty lines would make the scripts more readable for humans while, based on the measurements in [[Issue #2386](https://github.com/OpenIPC/firmware/issues/2386)](https://github.com/OpenIPC/firmware/issues/2386), having little or no practical impact on the firmware image size.

This was originally discussed in [[Issue #2386](https://github.com/OpenIPC/firmware/issues/2386)](https://github.com/OpenIPC/firmware/issues/2386). Based on my initial estimate, preserving empty lines should have only a negligible effect on firmware size.

The change in this PR is intentionally limited to preserving blank lines while continuing to strip comments.

## Hardware tested on

None.

I have not tested this change on a camera, and I do not have enough experience with the firmware build and validation process to perform the relevant hardware and image-size checks myself.

## Evidence

The measurements and validation below were performed by `@widgetii` and are documented in [[Issue #2386](https://github.com/OpenIPC/firmware/issues/2386)](https://github.com/OpenIPC/firmware/issues/2386).

Preserving blank lines added 514 bytes across 45 shipped shell scripts.

For `hi3518ev300_lite`, rebuilding the root filesystem both ways produced:

```text
as shipped (blank lines stripped): 5,226,496 B
blank lines preserved:             5,226,496 B
delta:                                     0 B
```

`@widgetii` also ran all 151 shipped scripts through the modified stripper under BusyBox `ash -n` with 0 parse failures.

The two boards closest to their image-size limits, `hi3519v101_lite` and `hi3518ev300_lite`, had not both been measured both ways at the time of that comment, so I cannot provide independent before/after measurements for them.

### Implementation

The change is a single-line modification to `general/scripts/strip-shell-comments.awk`:

```awk
-	if (/^[ \t]*$/) next
+	if (/^[ \t]*$/) { print; next }
```

It is important to explicitly print the empty line. Simply removing the original check would not work because the later `if (code ~ /^[ \t]*$/) next` also catches blank lines (`code_of("")` returns `""`).

## Scope

I have not independently completed the repository validation checklist below. I am relying on the testing and measurements documented by `@widgetii` in Issue #2386.

* [ ] No kernel patches under `general/package/all-patches/linux/` (those go to https://github.com/OpenIPC/linux)
* [ ] No files specific to a single retail camera model (those go to https://github.com/OpenIPC/builder)
* [ ] No probing or bring-up tooling (that goes to https://github.com/OpenIPC/ipctool)
* [ ] Nothing under `general/overlay/` or in a shared `load_<vendor>` script hardcodes a value specific to my board
* [ ] Package sources come from an OpenIPC repository, and any version bump keeps at least the specificity of the pin it replaces (a new package should pin a full 40-character SHA)
* [ ] No `LD_PRELOAD`, and no binaries that cannot be rebuilt from source

I would really appreciate your help with testing this PR. I don't have the experience or hardware needed to perform the relevant build and firmware validation myself, so please help verify that preserving empty lines does not cause any issues and that the resulting firmware images remain within their size limits.


general/scripts/strip-shell-comments.awk
```diff
@@ -104,7 +104,7 @@ BEGIN { nd = 0; sq = 0; dq = 0; cont = 0; prev_cont = 0 }
 		next
 	}
 
-	if (/^[ \t]*$/) next
+	if (/^[ \t]*$/) { print; next }
```
openipc-ai, line 108: The file's own first line still reads:

```awk
# Drop whole-line comments and blank lines from a POSIX shell script.
```

This change makes that false, and the header is the only documentation this pass has. Please update it to say comments only, and add a line on why blank lines are now kept, citing #2386 — the comments here explain *why*, and the next person to read this file should not have to find the PR to learn that the omission is deliberate.

openipc-ai, line 108: Nothing pins the new behaviour. `.github/scripts/test_strip_shell_comments.sh` asserts that comments go and that heredoc bodies, multi-line strings and continuations survive, but no fixture asserts that a blank line survives. As it stands, a future edit to this branch reverts the change silently and green.

A few lines next to the existing fixtures are enough:

```sh
# #2386: blank lines stay, so the shipped script keeps its paragraph breaks.
printf '#!/bin/sh\necho one\n\necho two\n' > "$WORK/blank.sh"
if [ "$(strip "$WORK/blank.sh" | wc -l)" = 4 ]; then
	ok "blank lines survive"
else
	bad "a blank line was dropped"
fi
```

openipc-ai (changes requested):
The size question is answered: this fits every board, with margin. I measured it rather than extrapolating from the issue.

## Per image

Per image the change adds **+895 B uncompressed** on `hi3516ev300_lite` (76 shipped scripts) and **+932 B** on `gk7205v200_lite` (82). The 514 B / 45 scripts figure in the body undercounts — it leaves out the majestic-webui `cgi-bin/j/*.cgi` and `sbin/*` scripts, which ship and are comment-stripped too. I attributed all 76/82 files to a source: three come from buildroot's own tree, and four needed replaying git history until the old stripper's output matched the built file byte for byte.

Re-squashing both trees with buildroot's own `mksquashfs -noappend -b 128K -comp xz` gives **+420 B** and **+216 B** compressed.

## On flash

Because the image is padded to 4096 B, each board moves by **0 or exactly 4 KB**. Both outcomes occur: on my `hi3516ev300_lite` repack the control had 3854 B of pad slack and the image did not move; on `gk7205v200_lite` it had 182 B and the image went 5044 KB -> 5048 KB. So the zero result in #2386 is real, but it is not general.

It can be predicted per board without rebuilding anything, by reading `bytes_used` out of the shipped squashfs superblock (unsigned 64-bit LE at offset 40) — spare pad bytes are `(-bytes_used) mod 4096`. From the 2026-09-19 nightly images:

| board | headroom | pad slack | effect |
| --- | --- | --- | --- |
| hi3516cv6xx_lite | 32 KB | 1784 B | +0 |
| hi3516ev300_neo | 36 KB | 3658 B | +0 |
| hi3516ev300_lite | 44 KB | 2606 B | +0 |
| gk7205v200_lite | 52 KB | 3250 B | +0 |
| gk7205v300_lite | 52 KB | 3158 B | +0 |
| gk7605v100_lite | 56 KB | 438 B | 0 or +4 KB |
| hi3518ev300_lite | 56 KB | 1030 B | +0 |
| gm8135_lite | 60 KB | 714 B | +0 |
| gm8136_lite | 60 KB | 682 B | +0 |

Every one of those absorbs the growth except `gk7605v100_lite`, which is marginal and has 56 KB free either way.

That also closes the open question left in #2386. `hi3519v101_lite`, sitting at exactly 5120 KB of its 5120 KB cap in that nightly, had **1094 B** of pad slack — it would not have tipped. It is at 188 KB free now, after #2459 moved five hisilicon boards back down (ev100 -> 128 KB, dv200 -> 336 KB, hi3518ev100 -> 208 KB, ev200_lite -> 68 KB).

The tightest board today is `hi3516cv6xx_lite` at 32 KB, so the worst case anywhere leaves 28 KB. NAND is safe by a wider margin: the minimum UBI headroom is 384 KB (`hi3516ev300_ultimate`, 16000/16384 KB), three whole 128 KB PEBs.

## Checks

`STRICT=1 .github/scripts/test_strip_shell_comments.sh` passes with the patch — 146 shipped scripts parse identically under busybox `ash`, and the saving goes from 210383 B to 207922 B, so 2461 B is given back across the whole tree. `test_sysupgrade.sh` and `test_check_mac.sh` also pass against the newly stripped forms.

## Before merge

Two small things, inline. After those, please mark this ready for review — a draft builds nothing, and this path widens the matrix to all 100 boards, which is exactly the evidence the issue asked for.

general/scripts/strip-shell-comments.awk
usa-: Done.

usa-: Done.

general/scripts/strip-shell-comments.awk
```diff
@@ -104,7 +105,7 @@ BEGIN { nd = 0; sq = 0; dq = 0; cont = 0; prev_cont = 0 }
...
```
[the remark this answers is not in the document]
openipc-ai: The rule is correctly applied — this does change the bytes of shipped scripts on every image — so here is the non-hardware evidence the rule's own escape asks for, with the limitation stated rather than papered over.

**This has not run on a camera.** The checklist boxes are deliberately unticked and *Hardware tested on* deliberately says None; neither the author nor I have a board this has booted on, and neither of us will tick them on someone else's behalf.

**Why the runtime risk is structural rather than untested.** The stripper already emitted blank lines inside heredoc bodies, quoted strings and continuations — the existing fixtures pin all three — because those branches `print` and `next` before the blank-line test is ever reached. The line this PR changes is reached only when none of those states hold, so the only lines it newly emits are blank lines in ordinary code position, where the shell has nothing to do with them: no command, no word, no redirection, no change of exit status. There is no position in which a script's behaviour can differ.

**Parse evidence against the shell the device runs.** `STRICT=1 .github/scripts/test_strip_shell_comments.sh` passes: 146 shipped scripts parse identically after stripping under busybox `ash` with `ASH_BASH_COMPAT`. `test_sysupgrade.sh` and `test_check_mac.sh` also pass against the newly stripped forms of `sysupgrade` and `extutils`. The new fixture in this PR fails against the previous stripper (`FAIL a blank line was dropped`), so it pins the behaviour rather than merely describing it.

**Before/after image size, measured.** Rewriting every shipped script in a built target and re-squashing with buildroot's own `mksquashfs -noappend -b 128K -comp xz`: +895 B uncompressed / +420 B compressed on `hi3516ev300_lite` (76 shipped scripts), +932 B / +216 B on `gk7205v200_lite` (82). Padding to 4096 B means each board moves by 0 or exactly 4 KB. Reading `bytes_used` out of the shipped squashfs superblocks shows every currently tight board absorbs it in existing pad: `hi3516cv6xx_lite` 1784 B of slack, `hi3516ev300_neo` 3658, `hi3516ev300_lite` 2606, `hi3518ev300_lite` 1030, `gm8135`/`gm8136` 714/682, `gk7205v200`/`v300` 3250/3158, with `gk7605v100_lite` (438 B) the only marginal case. Worst case anywhere is one 4 KB block against 32 KB of headroom on the tightest board in the tree.

**Board coverage.** `ci-matrix.py` puts this path at 100/100 boards, and the full matrix on this head is green — every image this change can reach was built and came in under its cap, including all three boards named above. That is the widest evidence available without a camera in hand.

openipc-ai (approved):
Both findings addressed, and the evidence is in.

The header now says what the pass actually does, and the new fixture is a real guard rather than a restatement — it fails against the previous stripper (`FAIL a blank line was dropped`), so the behaviour cannot be reverted silently.

The full 100-board matrix is green on this head, which settles the size question the issue opened with: every image this path reaches built under its cap, including `hi3516cv6xx_lite` (32 KB headroom, the tightest board in the tree), `hi3516ev300_neo` (36 KB), and `gk7605v100_lite`, the one board whose spare pad was small enough for the change to plausibly cost it a 4 KB block. That also closes the open question left on #2386: `hi3519v101_lite` and `hi3518ev300_lite` both build clean.

Thanks for picking this up and for turning the review round so quickly.

